PR #85, opened 2015-02-08
Polish translation

Kucharskov:
With untranslated changelog (useless)
And added "polish_translation" with "summary"

I suggest move changelog and "langs summary" to separated file cus that's useless to translate and copying between langs - when translator need change/fix name we need to edit all langs. Strange.


vishal0071:
OK, will do.

PR #86, opened 2015-02-09
Separated changelogs and translators

Kucharskov:
- Changelog.xml - separated changelogs in other langs. If translator
didn't translate new changelog Android should use english strings
- Trasnaltors.xml - separated names of users which translate app. Only in english folder - the same names in all langs in app. (easy to fix mistake in name)
- values-pl - minor bugfix and added last "Table(s)" string

I can only do that. But i dunno how to implement using changelog.xml and
translators.xml in app.

Regards!
Kucharskov ;)


vishal0071:
Thanks for your contribution (:

PR #114, opened 2015-04-09
Force chooser on open-with, allow default handler for archives

lbdroid:
Long-click selection --> menu --> open with, should force chooser for file handler rather than opening the default handler, which will be automatically called on short-click.

### DIFF
--- a/src/com/amaze/filemanager/activities/TextReader.java
+++ b/src/com/amaze/filemanager/activities/TextReader.java
@@ -306,7 +306,7 @@ public class TextReader extends ActionBarActivity implements TextWatcher {
             utils.showProps(mFile, c, theme1);
             return true;
         } else if (item.getItemId() == R.id.openwith) {
-            utils.openunknown(mFile, c);
+            utils.openunknown(mFile, c, false);
         }
         return super.onOptionsItemSelected(item);
     }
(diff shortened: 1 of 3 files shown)


lbdroid:
The second commit in this pull request is to solve the issue of being able to maintain transition animations when launching internal archive file handlers, while still respecting system wide default file handlers for those files. It does so by querying the PackageManager to determine if (a) we ARE the default handler (handle internally), (b) there ARE NO handlers (handle internally), (c) there are multiple handlers (launch using openunknown, which will present an option to set default).

arpitkh96:
I will review your code by today

arpitkh96:
this didnt worked for me.it always opens new activity whatever is the case

lbdroid:
I'm going to try to reproduce. What file type are you testing with? Do you have alternative handlers for that file type installed? What are they?

arpitkh96:
Zip file... I tried both with and without other handlers

arpitkh96:
also tried settings "always" in app chooser

lbdroid:
Ah, I know what it is.
I just need a minute to fix it.

lbdroid:
Sorry about that. I sent you another pull request with the fix.

PR #115, opened 2015-04-10
Fix string comparison to actually compare

lbdroid:
I could have sworn that Java was one of the languages that allowed you to compare strings using the "==" operator.

### DIFF
--- a/src/com/amaze/filemanager/utils/Futils.java
+++ b/src/com/amaze/filemanager/utils/Futils.java
@@ -160,7 +160,7 @@ public class Futils {
         String s="";
         ResolveInfo rii = c.getPackageManager().resolveActivity(intent, PackageManager.MATCH_DEFAULT_ONLY);
         if (rii !=  null && rii.activityInfo != null) s = rii.activityInfo.packageName;
-        if (s=="com.amaze.filemanager" || rii==null) return true;
+        if (s.equals("com.amaze.filemanager") || rii==null) return true;
         else return false;
     }
 


arpitkh96:
lol. this one always kills me

lbdroid:
And believe it or not, I actually tested it, but apparently didn't realize that I'd changed its behavior.

arpitkh96:
btw i need some suggestions

lbdroid:
Certainly

arpitkh96:
did you liked the latest fab realignment and decreased row icons size to 40dp?

arpitkh96:
for me 40dp isnt eyes friendly but guidelines say it to be like this

lbdroid:
The FAB does seem a bit strange. I honestly didn't even notice it until you mentioned it just now. In fact, I have a hard time making it appear -- could be my build environment.

lbdroid:
Tell me if this is normal behavior; launch the application and the FAB is not visible. Drag the screen left and it appears at the left lower action bar. Drag the screen to the right, and it moves to the right lower action bar.

lbdroid:
Its much smaller than the FAB in google messenger.

lbdroid:
Hmm, that's a different type of button.

lbdroid:
I'm getting a really weird effect on my nexus 6 that I can't reproduce on my 7 or 9. The scrollbar is crazy fat and jumpy.

lbdroid:
The 6 is running 5.1, the other two are running 5.0.

As far as the FAB goes, if the jumping side to side thing is not a fault of my build environment, then it should be kept at the right. I'd almost be inclined to use a full size FAB at the lower right of the screen (like google messenger) in a high contrast color to stand out from the file list.

lbdroid:
Hmm, I don't think that the guidelines are actually saying that you need to keep it down to 40dp. The first part of the guidelines appears to *accidentally* imply that a FAB at the border of the action bar should be a mini-FAB by using that as the example for a mini. If you look down lower in the guidelines, specifically starting at the heading "Placement", it shows what is clearly a standard 56dp FAB at that location.

lbdroid:
Ah, I see that the changing sides is intentional. I would just keep it on the right, it looks too bunched up with the file list icons right there.

arpitkh96:
Yeah some bugs are there in fab with latest changes.. Are fixable

arpitkh96:
I don't know what scrollbar are you taking about.

arpitkh96:
Changing sides was implemented to indicate the tab

arpitkh96:
But I also really didn't liked the way it works now

lbdroid:
When the file list is longer than the physical screen and you drag it up and down, the position indicator on the right of the scrollview. I'll send you a picture of what it is doing.

arpitkh96:
The 40dp is probably used only when fab is at left top... Just to match fab size to maintain kind of consistency

arpitkh96:
That layout of being fab just above the icons is being used a lot in examples shown in guidelines

lbdroid:
screenshot_2015-04-10-14-55-29

(note that I'm messing around with the fab in the layout at the same time)

lbdroid:
And to compare, this is from a nexus 7:

screenshot_2015-04-10-14-59-58

arpitkh96:
It was expected to be 40 dp only (same as that it icons). Works fine on my moto g. And probably vishal tested this on nexus 4 too.

arpitkh96:
Actually I don't got what you wanna show from these

lbdroid:
On the first picture. Look at the big blocky gray line down the right edge. Its not supposed to look like that. Its supposed to look like the subtle gray line in the second image.

arpitkh96:
Oh.... That's fast scroll bar. Just hold and drag to fast scroll. 5.1 changed its look. It's not present in you nexus 7 as list requires specific least number of items according to screen size to activate fast scroll

lbdroid:
Ah, ok that explains it then. Actually, the Nexus 7 hasn't got 5.1 yet.

arpitkh96:
BTW it's 1am here. Time to sleep. Ttyl

lbdroid:
Good night!

PR #116, opened 2015-04-10
Some cleanup for the FAB

lbdroid:


vishal0071:
Why does it looks messy at left? In the guidelines it is given in both the sides. Plus, we're using this thing ti our advantage. If we're on left pane, fab comes to right, denoting that we can slide to right pane, and vice versa. 
You have any other good idea to denote the current pane?

lbdroid:
There are a lot of different elements that are close together and overlapping when it is on the left side. In particular, the icon for the first file or directory in the current directory actually intersects with the FAB.

One of the problems is that there is really nothing about the position of the FAB that actually implies anything to do with tabs.

I think the simplest way to indicate left or right tab would be to add a subtle arrow in the space to the left of the path heading.

screenshot_2015-04-10-20-41-12

lbdroid:
Now if you actually look at the guidelines;
http://www.google.com/design/spec/components/buttons.html#buttons-floating-action-button

Under the heading "Associated content". It does show it on the left of the panel (the right picture with title "My files"), HOWEVER, the heading "Folders" in that example moves the content list down in order to keep it from bunching up when the list is scrolled to its default (top) position.

Having the button move comes close to violating the guideline "Don't make floating action buttons bounce". Also the guideline "A floating action button shouldn't float in a random location".

vishal0071:
"Don't make fab bounce" denotes a different thing. See the image related to it, it pretends as if the fab is drawn on the surface as if it is bounced.
Also, fab is not being drawn on a random place here.

Floating action buttons come in two sizes: the default, which should be used in most cases, and the mini, which should only be used to create visual continuity with other elements on the screen.

And see the image related to it. The fab on left is mini and icons are of same size to fab.

However I do agree that it overlaps the first item on left and right both positions. I'll see what can be done about that.
As for the arrow to denote the pane, it can't be done. See the app bar section in structure guidelines.

lbdroid:
I know what it means, which is why I said it comes CLOSE to violating guidelines.

Nothing in computer science is ever random, despite how much effort is put into creating sources of randomness. They don't mean true random in the guidelines as far as positioning. They mean coming up in a different place where it is not expected. Moving the FAB like that does satisfy their definition of random. It confuses the user. Like I said, nothing about its position denotes a tab.

Now frankly if you want to get into it, a spinner is not a tab. That is a very strange interface element to use for that purpose.

Now given that there are only two pages, why do you need an indicator for it at all? I don't see the purpose.

lbdroid:
Ah, yes. You ARE violating guidelines with the spinner. Not the presence of the spinner, which is acceptable in that location, but rather its *function*, which should be to filter the items in the current fragment, not to change fragments.

lbdroid:
Can you explain the purpose of even having two tabs?

I.e.
1) Why have tabs at all?
2) Why 2 and not a dynamic number of tabs that the user can open as needed? I.e., long-click --> open in new tab.

lbdroid:
Here is a way to redesign to clean it all up;

1) Move the "sort-by" functionality into the spinner. No longer show tabs in the spinner.
2) The PATH indicator line should NOT be capitalizing names, because filenames in linux are case sensitive.
3) Add new function: open-in-new-[instance|tab]
a) When open-in-new-tab, add an actual tab bar.
b) New instance adds to application switcher (square button menu).

arpitkh96:
tabs are there for kind of multitasking for eg copy from one paste to other.
they are only limited to 2 as we were not able to implement a completely bugfree dynamic viewpager.Specially while handling rotation changes.
idea of capital paths was also copied from some of already present material designed apps

lbdroid:
It really should not be capitalized. It isn't absolutely *terrible* when browsing through sdcard fat32 filesystems, but once you are into the root, there is too much ambiguity added by that.

Can you tell me what the problem was with viewpager? I believe I can make it work...

arpitkh96:
fragment lifecycles.rotation causes fcs

lbdroid:
Hmm, well that really doesn't give me much to go on....

arpitkh96:
well.....adding tabs was almost fixed.but removing tabs sometimes causes blank fragment and rotation causes fcs(probably that was due to null savedInstance). i dont remember much now

lbdroid:
I'll take a look at it and see what I can come up with.

arpitkh96:
https://github.com/arpitkh96/AmazeFileManager/commits/master?page=8
before the commit "limiting 2 tabs" on this page will give you that state

vishal0071:
BTW I do agree fab position doesn't denote any standard related to pane position. But users will generate thinking as to what fab position denotes as the time passes and as they use file manager.

As for the 2 tabs, we had implementation on earlier versions of app for dynamic number of tabs (as fragments without viewpager). But view pager caused some un fixable bugs and we both agreed that we don't really need more than 2 viewpager tabs.

lbdroid:
The user won't get used to that. The user will just get frustrated that sometimes they can't predict the location of the button and sometimes have to reach for it.

vishal0071:
OK. You might be correct. So we've decided to put indicator dots similar to nova launcher with auto hide. And put the fab to left of app bar. What say?

lbdroid:
I would say put the fab on the right.

Two reasons;
1) the width of the items in the list typically don't stretch all the way over to the right side of the screen, which means no interference.
2) it is easier to reach. Most people are right handed, and hold the phone in their right hand. Once you start getting to big phones, especially the likes of Nexus 6, or users with smaller hands, or worse, both, it can get downright difficult to reach the left of the screen without switching over to 2 hands.

Example; (notice how there is nothing around the FAB)
screenshot_2015-04-11-13-19-41

vishal0071:
You're right, but it looks more cool over to the left. Design guidelines also show a file manager example  wtih fab over to the left. It dosn't really interfere anything else except icon. 

screenshot_2015-04-11-22-59-14

lbdroid:
"but it looks more cool over to the left."
lol.

Let me ask you this question then; what is a higher priority... utility? Or cool?

arpitkh96:
what if category headers are added ,still you will prefer right?

lbdroid:
The category headers make it look better, but I still prefer the right because of how my thumb is attached to my hand.

vishal0071:
You wanna join us with the project so that we can discuss the stuff over on hangouts or whatsapp?

lbdroid:
Sure, I suppose I could. Need anything from me to do that?

arpitkh96:
your email or preferably your whatsapp phone number?

lbdroid:
aserbinski@gmail.com
I don't use whatsapp.
You can delete this message once you take it down.

arpitkh96:
Done

PR #169, opened 2015-07-23
Portuguese update

smarquespt:


arpitkh96:
You might wanna also add those newly committed strings.then i will merge

PR #193, opened 2015-10-24
Updated Greek translation

VasilisKos:
Though I found the "videos, apks, documents, images" strings of the sliding menu, there isn't the "Audio" string. Anyway its translation is "Ήχος". Please add it if you can.


vishal0071:
Will fix it, thanks for pointing (:

PR #194, opened 2015-10-24
fix typo in README

prempalsingh:
couldn't stand that typo.


vishal0071:
What about, 'You may loose your files'?

prempalsingh:
There's a difference between loose and lose. The word you are looking for
here will be 'lose'.

vishal0071:
Ahh, OK got it (:

PR #202, opened 2015-12-21
Add Brazilian Portuguese translation

amalvarenga:
I'm not sure if values-pt-BR is the correct name to Brazilian Portuguese translation folder. Please, fix it if necessary.


vishal0071:
Rename this directory to *values-pt-rBR* and the Portuguese one to *values-pr-rPT* 
and make another commit which will automatically be added to this pull request.

PR #211, opened 2016-01-26
Redesign indicators of tabs

proninyaroslav:
Hello. I did a redesign of indicators, by making them circle. I think it's a good idea, now they fit perfectly into design of the application.


vishal0071:
Great!

PR #215, opened 2016-02-09
a tiny addition to customize the toast on opening

moovida:
This very small change would allow external apps to send a message to the user on opening of amaze. 

This could be seen as a file selection window title, done the very easy way.


vishal0071:
You forgot to assign value to title from the if case.

moovida:
Outch, making an error in so few lines of code is really bad, sorry. Going to fix it at once :-/

moovida:
Ok, fixed. Also extracted key to Constants to add some javadoc example, hope that is ok.

vishal0071:
Great!

PR #235, opened 2016-03-20
New search dialog based on SearchView

proninyaroslav:
Hello. I think that processing the search dialog to the standard form UI is a good idea.

screenshot_20160320-175704 screenshot_20160320-181002


vishal0071:
Nice idea! But I can't merge as there are conflicts in class MainActivity and MainActivityHelper. You might wanna sync latest commit and make changes according to it.
Also, the left margin is quite a lot between the hamburger icon and the magnifier icon. You can see the guidelines for the exact margin dp value.

proninyaroslav:
I merged 15bd463 and 7bcfc31 commits (Main, MainActivity and MainActivityHelper with your changes), but still Pull Request indicates to the conflict in branches.
Also, I fixed the problem with big indented from SearchView.
Before 
screenshot_20160320-175704 
After 
screenshot_20160321-130030

proninyaroslav:
I consider reasonable to rollback your commit (15bd463) and merge my (7bcfc31) with your changes.

PR #261, opened 2016-05-30
Update Google Play and F-Droid badges

relan:
Use the latest versions of the badges. Please see [here](https://github.com/relan/AmazeFileManager/blob/badges/README.md) how they look on the page.


vishal0071:
Thanks for this bro (:

PR #265, opened 2016-06-13
Added FTP Server functionality

yashwanthreddyg:
Hi, I added Ftp server using a service and added a nav bar item for getting to the ftp fragment. The server runs on a different thread. Please review it.

Thanks
Yashwanth


vishal0071:
@yashwanthreddyg there is some problem fetching the new dependencies from maven. Would you mind resolving this, I can't seem to fix it myself even after adding jar files of those.

yashwanthreddyg:
@vishal0071 I'll look into it.

yashwanthreddyg:
@vishal0071 can you tell me the names of the dependencies you're facing problem with.

vishal0071:
I was getting dirty build yesterday. Alhough maven dependencies still can't be resolved, don't know why. But I added jar files of slf4j and everything's fine now.

ka223:
@yashwanthreddyg:
That is very nice, but it seems, that I have the same problem with this amaze ftp server as over there: :/
https://github.com/ppareit/swiftp/issues/58

Any ideas?

Can I connect with Amaze to another ftp server?
This would be very useful!

yashwanthreddyg:
Hi @ka223 ,
The FTP Server doesn't yet use 'wakelock' which probably might help the app stay awake when the screen turns off. Will try to implement this change in the coming days.

ka223:
@yashwanthreddyg 
Thanks, but I this will unfortunately not solve the problem, at the filezilla log, the connection is lost with a timeout after every "Transfer complete" message.

vishal0071:
@ka223 I don't know how, but when I was trying ftp with a specific router at my workplace, it didn't work not matter how many times I tried. But as soon as I switched to another one there, it started working fluently. Can't say precisely as to what the changes in both router's configs were, but I strongly recommend you to try with atleast 2-3 different routers, and see if you can get it right.

PR #272, opened 2016-06-19
Notification for FTP; Service persistence after exit

yashwanthreddyg:
1. Added Notification for stopping the FTPService running in the background
2. FTPService will now run in the background even if the app is closed from the recents screen.
3. Added an image in the fragment for ftp (Added the default ic_ftp_* image, needs to be replaced with a bigger and better looking one).
4. Minor code refactoring.

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -126,8 +126,8 @@
             android:name="com.amaze.filemanager.services.ftpservice.FTPReceiver"
             android:exported="true">
             <intent-filter>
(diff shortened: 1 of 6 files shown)


vishal0071:
Tested it just now, well done!

yashwanthreddyg:
@vishal0071 
@pradyukrish is working on FTPClient. The service already supports authentication, we're just not using it. I'll update the code with authentication. I don't think these libraries provide the client functionality.

vishal0071:
Very well!

PR #278, opened 2016-06-26
Values fa

mohammadi66:
some persian words added to the values-fa (Iranian language)


vishal0071:
I can't see any Farsi translations except for two strings

PR #309, opened 2016-08-18
Added support for Android N app settings in App info

pylerSM:
Please format all source files... Code looks horrible without formatting


vishal0071:
You're right! We're already pushing well documented and clean code from some time now. But older code is still horrible.

PR #341, opened 2016-10-17
Cleaning all uses/instanciations of Futils

rpiotaix:
…as it already define one.


rpiotaix:
Prerequisite for #340

PR #346, opened 2016-10-23
Comments about methods of ColorUsage and ColorPreference.

rpiotaix:
Also changes the default behavior of the getPrimary to return the first primary color when the given index is not in the list of expected choices (1 or 2).


vishal0071:
Why do you remove timestamp BTW?

PR #371, opened 2016-11-17
FTP port can now be set from preferences

rpiotaix:


vishal0071:
Don't add a preference in settings screen about it. Add an option in contextual menu of the FTP fragment.

PR #372, opened 2016-11-18
Update TextReader.java

zhEdward:
Fix some deprecated api and  ViewAnimationUtils Compatibility(like in MainActivity)


rpiotaix:
It can also be interesting to rename 'class a' to something more understandable while you're at it :)

zhEdward:
hi @rpiotaix  i check  `original master` ,**class a**(scrollview ) has already there when pull this request(maybe be I import some package, cause git think those lines was modify) 
 sorry my poor english

PR #415, opened 2017-01-21
AmazeFileManager: Fix app name color at app manager 

martincz:
if current theme is light style, app name should be black color

Change-Id: If98fc6762d16193f9b59ae387be678e12a470926


vishal0071:
Try removing these text view line colors and replace all TextView widgets with [ThemedTextView](https://github.com/arpitkh96/AmazeFileManager/blob/master/src/main/java/com/amaze/filemanager/ui/views/ThemedTextView.java) in the fragment's layout. It's just more reusable.

martincz:
Ok, it's a better way. Thanks

PR #434, opened 2017-02-01
fix ext sdcard issues with SAF while copy, move and save in text reader

Unpublished:
TextReader:
hide save icon after successful save
set original text only after successful save

### DIFF
--- a/src/main/java/com/amaze/filemanager/utils/GenericCopyUtil.java
+++ b/src/main/java/com/amaze/filemanager/utils/GenericCopyUtil.java
@@ -6,6 +6,7 @@ import android.support.v4.provider.DocumentFile;
 import android.util.Log;
 
 import com.amaze.filemanager.filesystem.BaseFile;
+import com.amaze.filemanager.filesystem.FileUtil;
 import com.amaze.filemanager.filesystem.HFile;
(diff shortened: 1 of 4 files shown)


src/main/java/com/amaze/filemanager/utils/GenericCopyUtil.java
```diff
@@ -72,7 +73,18 @@ private void startCopy() throws IOException {
             } else {
 
                 // source file is neither smb nor otg; getting a channel from direct file instead of stream
-                inChannel = new RandomAccessFile(new File(mSourceFile.getPath()), "r").getChannel();
+                File file = new File(mSourceFile.getPath());
```
Unpublished: like this?

Unpublished:
It wasn't necessary for me to move a file from ext sdcard to sdcard and you're welcome :)

vishal0071:
If file is not readable directly from path (on sd cards), open input stream to it using ContentResolver#getInputStream. Similar to opening output stream. If a generic function already doesn't exist in FileUtil (similar to #isWritable) you could create one, or use File#canRead?

Unpublished:
Done, check it out :)

PR #458, opened 2017-02-26
Clean scrollview

rpiotaix:
Was not used (nothing was displayed and was not taking any space)


rpiotaix:
OMG, my bad.

Seems that it is used when clicking on the pathbar to decompose the path of the current folder.

PR #518, opened 2017-04-13
Changed names and tried to make the code more understandable at first glance

EmmanuelMess:
Check the commits for more details, this changes should not change how the app works, just make the code more readable (and in some cases efficient).
Note: there are some unused classes, that I did not delete, but should be checked (PropertiesSheet, CustumNestedScrollView, CustumViewPager and EditIntPreference). Also there is code commented that should be deleted or explained, and catches that don't contain code.


vishal0071:
Everything seems fine, you really did put quite a lot of time into this. I appreciate your help.
I've marked some places where 'dirs' is been replaced with 'a' which broke some critical code along with rendering many comments meaningless. Change them back and we'll be good to go.

EmmanuelMess:
Done.

PR #546, opened 2017-05-02
Fix crash when user rotates screen with color picker in settings screen

gitcoderoid:
https://github.com/arpitkh96/AmazeFileManager/issues/547


gitcoderoid:
This pull request provides one additional change. If you are on Settings screen and if you open ColorFragment then you rotate screen you are still on ColorFragment. Master branch has bug because if you rotate screen on ColorFragment you are on Settings again.

PR #549, opened 2017-05-03
Things

MaanooAk:
Really liked this project, so I improved some things here and there:
- Update delete dialog
- Extend text editor mime types (and a little bit faster)
- Show option for headers (really dont liked them)
- Made show modified date work (fixed padding, I hope it was disabled beacuse of it)
- Text reader monospace font option
- Extended Text reader intent filter (~~"application/*", mainly for js~~ javascript, json and xml)
- Extend mime types support
- Integral names fix

### DIFF
--- a/src/fdroid/java/com/amaze/filemanager/activities/MainActivity.java
+++ b/src/fdroid/java/com/amaze/filemanager/activities/MainActivity.java
@@ -2242,6 +2242,7 @@ public class MainActivity extends BaseActivity implements OnRequestPermissionsRe
                 break;
             case CUSTOM:
                 newPath = mainActivityHelper.getIntegralNames(news);
+                break;
             case DROPBOX:
             case BOX:
             case ONEDRIVE:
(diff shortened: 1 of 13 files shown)


Alexander--:
Doesn't that mean, that Amaze text editor will try to open application/ogg and application/octet-stream files? This isn't a wise course of action…

vishal0071:
@MaanooAk I don't have any idea about js, so you wanna state why put an intent filter for application/* to support js ? 
On a side note, you can always try to open any file with text editor using Open With in popup menu.

MaanooAk:
So, there are these mime types:
- application/javascript
- application/json
- application/xml

which are defined as `application` but they are actually text (and I use them a lot)

Possible solutions:
- Remove the `application/*` and forget about all of this
or
- Remove the `application/*` and add after the `text/*` the mimes listed above (I suggest this one)

vishal0071:
Alright, can you make changes reflecting your second solution and add another commit here?

PR #564, opened 2017-05-08
Changed how ThemedImageView works and solved black shortcut items on dark theme

EmmanuelMess:
Changed how ThemedImageView works, not very fundamental, but now it sets a color filter that's white (#FFFFFFFF) instead of using another image, changed layouts accordingly, but left the old png's.
Also, now shortcut preferences have white icon when on dark theme.

### DIFF
--- a/src/main/java/com/amaze/filemanager/ui/views/ThemedImageButton.java
+++ b/src/main/java/com/amaze/filemanager/ui/views/ThemedImageButton.java
@@ -1,0 +1,37 @@
+package com.amaze.filemanager.ui.views;
+
+import android.content.Context;
+import android.util.AttributeSet;
+import android.widget.ImageButton;
+
+/**
+ * @author Emmanuel
+ *         on 8/5/2017, at 13:39.
+ */
+
+public class ThemedImageButton extends ThemedImageView {
+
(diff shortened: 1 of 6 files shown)


vishal0071:
This will lead to slow layout rendering. Loading a resource is always better than manually processing it.

EmmanuelMess:
The thing is if android first converts vectors into bitmaps or first applies the color. I know your images are pngs, but the idea was to lay the grounds to start using vectors.

vishal0071:
We're already using vectors for places where it is really required, for eg. themed icons in drawer. We can't theme every icon resource for every color. But the least we can do is provide a separate dark and light icons for places it's required.

EmmanuelMess:
> But the least we can do is provide a separate dark and light icons for places it's required.

Why? Changing a vector's color is very inexpensive.

vishal0071:
@EmmanuelMess then for light theme make them gray. You can pick the shade from existing icons.

EmmanuelMess:
@vishal0071 For light theme they are black if clickable and grey if disabled. What are you proposing?

vishal0071:
Black is not an option. Pick a shade of grey from normal icons. For disabled ones, a relatively lighter shade will do.

PR #577, opened 2017-05-15
Feature : Implemented Image Viewer 

### LINKED ISSUE #502: Image Viewer
is it possible to implement an internal image viewer?

RamolaWeb:
Fix #502 to have a default image viewer for the application


RamolaWeb:
In this pull request I have created a Default Image Viewer in the application. All the images are now displayed through this and used Picasso for rendering large size image to prevent the Out of Memory Exception.

vishal0071:
Too buggy, isn't necessary either. I'm sorry but, we already have lots of things to finish up so can't help with this implementation. But if you get this bug free and fully working, I'll merge this into Amaze.
What I think would be basic requirement for production:
An auto-hideable toolbar with options menu, in which options to share, delete, set picture as (fires up intent to be used as wallpaper or profile picture or whatever), properties (orientation, camera architecture, height, width, exposure, iso etc, size, filename, etc).
And yeah, there's no pinch zoom and aspect ration is not preserved, image it being stretched to fill the screen. 
Good luck (:

RamolaWeb:
@vishal0071 Can You Please List Some of the bugs in the Pull Request found which i can fix. Also I would made the Basic Feature provided by You for the image Viewer in the Pull Request.

vishal0071:
1. Image stretches to match height in portrait (didn't check landscape)
2. No intent filters to handle intent from other apps
3. Only File (local) filesystem support. Add support for content provider, uris

PR #588, opened 2017-05-19
Feature/style delete dialog

theck13:
### Fix
Update the delete dialog style by including the following:
- Create custom view with Material-like "Directories" and "Files" text.
- Update deprecated callback methods from the MaterialDialog library [here](https://github.com/afollestad/material-dialogs/blob/master/core/src/main/java/com/afollestad/materialdialogs/MaterialDialog.java#L2194-L2246).
- Follow the Material Design guidelines for alert dialogs [here](https://material.io/guidelines/components/dialogs.html#dialogs-alerts).
- Organize code variable names and logic conditions for better readability.

### Test
1. Long-press directories and/or files.
2. Tap ***Delete*** action.
3. Notice delete dialog style.


src/main/java/com/amaze/filemanager/utils/Futils.java
```diff
@@ -680,90 +680,128 @@ public void onSelection(MaterialDialog materialDialog, View view, int i, CharSeq
...
-        dialogBuilder.title(mainFragment.getResources().getString(R.string.confirm));
-
         int counterDirectories = 0;
         int counterFiles = 0;
-        long longSizeTotal = 0;
+        long sizeTotal = 0;
         final ArrayList<BaseFile> itemsToDelete = new ArrayList<>();
-        StringBuilder listDirectories = new StringBuilder();
-        StringBuilder listFiles = new StringBuilder();
+        StringBuilder builderDirectories = new StringBuilder();
```
EmmanuelMess: No need to put the object name in the variable name.

theck13: There's no _need_ for it, but prepending `builder` might help understand how that variable is used.  I can remove that if it's a big deal.  Regardless, debating variable names can lead to [bikeshedding](https://en.wiktionary.org/wiki/bikeshedding).

EmmanuelMess: No problem, do as you wish.

vishal0071:
Nice addition, but I would need you to do some changes before I can merge this.
1. Transfer you method ````Futils#deleteFiles```` to ````GeneralDialogCreation```` and replace it's method ````#deleteFilesDialog```` because that's what is being called.
2. The total field doesn't add the size of folders. Look into ````HFile#folderSize``` which is extended by ````BaseFile````
3. Change the title back to 'Delete'. The description is good enough, we don't need that big a title.
Thanks for your contribution (:

theck13:
> Change the title back to 'Delete'. The description is good enough, we don't need that big a title.

Using "Delete items permanently?" for the title follows the [Material guidelines for dialog alerts](https://material.io/guidelines/components/dialogs.html#dialogs-alerts) more closely.  Since the action with this dialog deletes the directories and/or files permanently, it should fall into the "high-risk situations" described in the guidelines.  Users may not "be able to understand the choices based on the title and button text alone" if only "Delete" was the title.  The current changes give a "clear question or statement with an explanation in the content area" to eliminate any ambiguity as the guidelines suggest.

EmmanuelMess:
Delete as opposed to move to trash, it's obvious (it's permanent) in my opinion.

src/main/java/com/amaze/filemanager/ui/dialogs/GeneralDialogCreation.java
```diff
@@ -162,7 +162,7 @@ public static void deleteFilesDialog(ArrayList<LayoutElement> layoutElements, fi
                 directories.append(++counterDirectories)
                         .append(". ")
                         .append(layoutElement.getTitle());
-                // TODO: Get directory size.
+                sizeTotal += layoutElement.generateBaseFile().folderSize(mainFragment.getContext());
```
EmmanuelMess: folderSize() on main thread isn't a good idea.

vishal0071:
Indeed @theck13 but the only line that's in the description says the thing. If you're so fixated on keeping the title, then you might wanna remove the description.

Issue #660, opened 2017-07-09
While running app in grid view when we change orientation from portrait to landscape,the app automatically changes to list view.


NamanRawat1:
The app requires to saved state while working on grid view

EmmanuelMess:
Back to the mines it is(?). I'll solve this.

NamanRawat1:
Great looking forward to a solution to this problem.

vishal0071:
Partially fixed, please try new update (v3.1.2 Beta 13) and report

PR #662, opened 2017-07-12
Fixed 'show files as grid' state not being saved #660

EmmanuelMess:


EmmanuelMess:
Made a mistake, when rotating the state is passed to both tabs, not only the affected one (if a tab is a list and the other a grid, both end up being a grid).

EmmanuelMess:
Do pull now. Mind that it has some irregular errors when switching tabs between a grid and list and then 
reloading.

Issue #679, opened 2017-07-24
FTP Server isn't working on Android TV with Eth (fix coming)


Luca1991:
The FTP Server is not working on Android TV while connected to the net using wired (Eth) connection.
I'm fixing it and I'll push a PR soon. I'm opening this issue just to keep track.

Luca1991:
PR sent.

PR #680, opened 2017-07-24
FTP Server fixed while using Eth

Luca1991:
This is the fix for issue #679. Now the FTP Server is working even when connected using an Eth (wired) connection.
This also fix a possible crash due to a NPE in function getFTPAddressString().

Have a nice day.
Luca


src/main/java/com/amaze/filemanager/fragments/FTPServerFragment.java
```diff
@@ -496,7 +498,11 @@ public void onCheckedChanged(CompoundButton buttonView, boolean isChecked) {
      * @return address at which server is running
      */
     private String getFTPAddressString() {
-        return "ftp://" + FTPService.getLocalInetAddress(getContext()).getHostAddress() + ":" + FTPService.getPort();
+        String hostAddress = "";
+        InetAddress ia = FTPService.getLocalInetAddress(getContext());
+        if (ia != null)
+            hostAddress = ia.getHostAddress();
+        return "ftp://" + hostAddress  + ":" + FTPService.getPort();
```
EmmanuelMess: If `ia` is null `getFTPAddressString()` will return "ftp://:3333" (3333 being the port).
Why did you change this?

Luca1991: No, because if ia is null then getHostAddress will be called on a null reference, producing an NPE (tested on my Android TV).

EmmanuelMess: You are saying that if `ia` is null `getFTPAddressString()` will not return "ftp://:3333"?

Luca1991:
Sure. It crash BEFORE that the return value is even generated. If ia is NULL, then calling getHostAddress on ia will produce a NPE (and not a NULL string!). Obviously calling a function on a null object reference cause a NPE.

EmmanuelMess:
I am not saying calling a function on a null object will not cause a NPE; I am asking what will happen if `ia` is null here:
```
String hostAddress = "";
InetAddress ia = FTPService.getLocalInetAddress(getContext());
if (ia != null)
    hostAddress = ia.getHostAddress();
return "ftp://" + hostAddress  + ":" + FTPService.getPort();
```
because the hostAddress will end up being an empty string therefore the method will return `"ftp://" + "" + ":" + FTPService.getPort();`.

Luca1991:
Ah sorry I haven't understood your question correctly then. Yes then, if ia is NULL then hostAddress will be an empty string. I think this is correct, because if ia is NULL then there is a problem getting the InetAddress and since we don't have that value we can't return a valid address. I think that this is better than a crash honestly. Surely we can think a better solution :)

Luca1991:
@vishal0071 thanks for merging my PR, but sill if @EmmanuelMess have a better idea for fixing the NPE, I'm 100% open to any patch to my code (y)

EmmanuelMess:
I can't test because I don't have a TV but I think that `getFTPAddressString()` should either return a valid FTP address or throw an exception if no valid FTP address can actually be returned; in the case where `getFTPAddressString()` would throw an exception it shouldn't be used (why would you use `getFTPAddressString()` where you can't get a valid FTPAddress).

Luca1991:
@EmmanuelMess you are right: `getFTPAddressString()` should throw an exception in that case. Do you want to patch it yourself? Or I can patch it later ;)
Anyway thank you very much for your patience and sorry for my misunderstanding earlier 👍

EmmanuelMess:
I'll leave it to you then, but remember to use brackets when the if spans more than a line (as per [Google's Code Style for Contributors](https://source.android.com/source/code-style#use-standard-brace-style)).

PR #683, opened 2017-07-27
Restructure to typical format

Jawnnypoo:
This is a *massive* PR in terms of file changed, but the changes are very simple:

- Moved app code into an `app` folder, breaking the build.gradle up into the parts relevant to the `app` and relevant to the overall project
- Commit the gradle wrapper so that the project can be built immediately and the gradle version is dictated for the project. This updates the gradle verison to 4.0 which prepares the app properly for the upcoming changes with Android Studio 3.0


vishal0071:
Thanks! This was a greatly needed change.

Jawnnypoo:
No prob! I was thinking the project could also use some CI, if I add a `travis.yml` config file, would you guys be interested in enabling it for the project (so that PR's as big as this would can be checked)

vishal0071:
Yes, it'll be good for checking PRs. But I'm not sure if that would be worth, we don't have any unit tests in this project. And afaik, travis CI also runs tests after every commit.

Jawnnypoo:
You can specify if you want it to run test or just build. I can configure it to just build the project 👍

vishal0071:
Great! (:

Issue #515, opened 2017-04-12
[Feature request] FTP wifi ap & USB tethering support


adamsson6:
Hi, i recently discovered this app on f-droid and it's very cool, now it's my default file manager.

I saw that there was a ftp sever, but I cannot use it often because I'm moving frequently and I don't have access to my home router.

Please can you add support for the ftp so we can use ftp server in USB tethering or android Wifi AP mode ?

Thanks !

HEXcube:
Some more detailed info on this issue: the FTP Server shows **No Connection** when fone's Portable Hotspot is enabled. See:

![FTP Server on Portable Hotspot](https://user-images.githubusercontent.com/3755396/28245555-7e63b1f6-6a27-11e7-8113-dca781e9a588.png)

----

It's supposed to behave similar to the way it does when connected to regular WiFi:

![FTP Server on regular WiFi](https://user-images.githubusercontent.com/3755396/28245577-f5cafe48-6a27-11e7-9b7f-3cde0f998bf9.png)

For a working example, I tested Primitive FTPd ([Link to GitHub page](https://github.com/wolpi/prim-ftpd)) which works on both WiFi Client and Hotspot modes.

vishal0071:
Feature coming with new update, courtesy of @TranceLove (:

HEXcube:
Yay! :smiley: Kudos to @TranceLove ! I see the pull requests https://github.com/arpitkh96/AmazeFileManager/pull/685 and https://github.com/arpitkh96/AmazeFileManager/pull/686 are already merged. Looking forward to new release on F-Droid

HEXcube:
Got the new update from F-droid. FTP server now works with WiFi Hotspot. :smiley:

PR #684, opened 2017-07-28
Allow FTP service to run when wifi AP is enabled

TranceLove:
This is an attempt to solve #515 which allow FTP service to run when device is running with the device's wifi AP is enabled.


TranceLove:
Close this first, pending additional work

PR #685, opened 2017-07-28
Allow FTP service to run when wifi AP is enabled

TranceLove:
This is an attempt to solve #515 which allow FTP service to run when device is running with the device's wifi AP is enabled.

It also fixes the problem that URL of FTP server will stay the same when the device's connectivity changes, e.g from connected to wifi switched to become a wifi hotspot.


vishal0071:
Thank you (:

app/src/main/java/com/amaze/filemanager/services/ftpservice/FTPService.java
```diff
@@ -354,6 +360,12 @@ public static byte byteOfInt(int value, int which) {
         return (byte) (value >> shift);
     }
 
+    public static int getPort(Context context)
```
EmmanuelMess: You should pass a reference to SharedPreferences, instead of creating it, so that you don't accidentally have a method creating an instance when you already have one.

app/src/main/java/com/amaze/filemanager/services/ftpservice/FTPService.java
```diff
@@ -184,7 +182,7 @@ public void run() {
             }
         }
 
-        fac.setPort(port);
+        fac.setPort(preferences.getInt(PORT_PREFERENCE_KEY, DEFAULT_PORT));
```
EmmanuelMess: Shouldn't `getPort()` be used here?

TranceLove: With `getPort()` should accept SharedPreferences instead of Context, it should.

PR #693, opened 2017-07-29
Fix #689 and optimized

### LINKED ISSUE #689: Crash when performing any operation after 'Select all' is pressed in action mode
Try performing any operation like delete/compress after pressing on 'Select all' from action mode, it leads to a crash.

EmmanuelMess:
Fixed select all making appbar actions crash (#689). Also optimized how animation and checked booleans work.


EmmanuelMess:
Just realized something, don't pull yet.

EmmanuelMess:
On tablets when you make a selection and then rotate twice the selection will be cleaned before the drawer closes. 
You can pull now.

PR #700, opened 2017-08-01
#691 Enabled refresh for empty folders

Foso:
Empty folders can now be refreshed through swiping down


vishal0071:
Please set accent color to the swipe refresh layout as there is a black color accent by default and it doesn't match up with the app accent color.

PR #702, opened 2017-08-02
Changed SwipeRefreshLayout color sheme to accentColor

Foso:


vishal0071:
Thank you very much for the contribution.
You would need to tweak the headers a bit. Right now there is just an empty slot in place of heads, but a preferred UX would be to make them similar to the ones in list (a full row state files and folders below it). Something similar to Google Photos.

app/src/main/java/com/amaze/filemanager/adapters/RecyclerAdapter.java
```diff
@@ -299,7 +300,7 @@ public void createHeaders(boolean invalidate)  {
         boolean[] headers = new boolean[]{false, false};
 
         for (int i = 0; i < itemsDigested.size(); i++) {
-            if(mainFrag.IS_LIST) {// TODO: 31/5/2017 add fragments to gird view
+            // TODO: 31/5/2017 add fragments to gird view
```
EmmanuelMess: Next time get rid of the todo 😄

EmmanuelMess:
Headers in gridview aren't what they should be...

vishal0071:
Oh, I just noticed. Worked out a quick fix myself, so you don't need to bother @Foso Thank you for this (:

Issue #623, opened 2017-06-03
When viewing zip file path is shown as "/" (root)


EmmanuelMess:
Open a zip file, check appbar, ..."/*.zip" is now "/". It probably should be ".../Desktop.zip/" or similar.
<img src="https://cloud.githubusercontent.com/assets/10991116/26757437/84c47bdc-4891-11e7-8009-d70f59a11934.png" height=350/>

vishal0071:
Not a bug, it's by design.

EmmanuelMess:
@vishal0071 It would be nicer if you left "Amaze" in the appbar and put *.zip/ in the route.

vishal0071:
Sure, sounds good. But putting full path might not be a good idea, just a ````xyz.zip/``` preceding it would be fine. In case one want to see the full path, we can add an action mode menu item saying 'Details' and show a properties dialog, like we have in Amaze Text Viewer.

PR #723, opened 2017-08-21
Fix zip AppBar

EmmanuelMess:
* Fixed Toolbar title and BottomBar path (#623).
* Fixed root vector on BottomBar to better represent zip relative paths.
* Fixed crash when BottomBar was clicked after opening zip file.
* Cleaned up the code a bit.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/fragments/ZipViewer.java
+++ b/app/src/main/java/com/amaze/filemanager/fragments/ZipViewer.java
@@ -26,14 +26,12 @@ import android.content.ServiceConnection;
 import android.content.SharedPreferences;
 import android.content.res.Resources;
 import android.content.res.TypedArray;
-import android.database.Cursor;
 import android.graphics.drawable.ColorDrawable;
 import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
 import android.os.IBinder;
 import android.preference.PreferenceManager;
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/amaze/filemanager/fragments/ZipViewer.java
```diff
@@ -78,9 +76,19 @@
 import java.util.ArrayList;
 import java.util.Calendar;
 
-public class ZipViewer extends Fragment {
+public class ZipViewer extends Fragment implements BottomBarButtonPath {
```
szholdiyarov: shouldn't it be "ZipViewFragment"?

EmmanuelMess: Yes, but I would have to change a lot of fragments to have their correct name and that would make a lot more changes and the pull more prone to conflict. That should be corrected in another pull request (with just naming corrections).

vishal0071:
A little bug that I found,. pressing on bottom bar does show file hierarchy but pressing on any one of them doesn't do anything, also pressing on root icon (zip) force closes Amaze. Can't you look into it?

EmmanuelMess:
@vishal0071 done

EmmanuelMess:
Can anyone check with a rar file?

vishal0071:
Looks good.

PR #726, opened 2017-08-22
Clean unused resources and improve translation of Chinese

Bambooin:


app/src/main/res/values/translators.xml
```diff
@@ -33,7 +33,6 @@
     <string name="turkish_translation_summary" translatable="false">Hakan Güven</string>
     <string name="ukrainian_translation_summary" translatable="false">Pablo Luchiano</string>
     <string name="portuguese_translation_summary" translatable="false">Sérgio Marques</string>
-    <string name="brazilian_portuguese_translation_summary" translatable="false">André Marcelo Alvarenga</string>
```
EmmanuelMess: This should probably be used.

vishal0071:
Indeed it was

Issue #725, opened 2017-08-21
open file in separate process


seregaxvm:
Why Amaze opens files inside it's own task? Probably it's better to spawn separate process. This way you can switch between opened windows.

EmmanuelMess:
So, what you want is to be able to open several files at the same time?

seregaxvm:
Basically, yes. Right now I use module from xposed framework to simulate this behavior. It sets `FLAG_ACTIVITY_NEW_TASK` for new tasks.

EmmanuelMess:
Ok, I understand, probably should use `FLAG_ACTIVITY_NEW_DOCUMENT` though.

EmmanuelMess:
You should start explaining to @vishal0071 why anyone would use this feature, so that my pull request can exit limbo.

PR #734, opened 2017-08-25
Added setting to open files in new stack

EmmanuelMess:
* Added setting to open files in new stack (#725).
* Cleaned code a bit.


EmmanuelMess:
I need to test with DbViewer, but don't know which file extensions it opens.

EmmanuelMess:
Mind explaining?

vishal0071:
My bad, but we still need to discuss the feasibility of this feature.

EmmanuelMess:
Do speak your mind, and remember that this is turned off by default and is under "Advanced" in the settings category.

PR #748, opened 2017-09-01
Updated android tools and support libs to v26.0.1

EmmanuelMess:


vishal0071:
Let's not target sdk 26 yet.

EmmanuelMess:
Here's the [migration web from Android](https://developer.android.com/about/versions/oreo/android-8.0-migration.html). Check tables 1 and 2, if none of those are problematic then I think we should be good to update.

vishal0071:
Possible list of problems we might need to fix first
- Implementations of this method may no longer call Collections.sort(), or your app will throw an exception due to stack overflow.
- Apps targeting the Android 8.0 (API level 26) version of the SDK must implement one or more notification channels to post notifications to users.
- Remove broadcast receivers from your manifest file
- You can no longer assume that APKs reside in directories whose names end in -1 or -2. Apps should use sourceDir to get the directory, and not rely on the directory format directly.
- Managed profiles created on devices running Android 8.0 (API level 26) automatically have the DISALLOW_INSTALL_UNKNOWN_SOURCES user restriction enabled.
- findViewById() signature change
- Content change notifications
- Account access and discoverability
- Permissions

EmmanuelMess:
@vishal0071 ..."you should remove any broadcast receivers that are registered for _implicit_ broadcast intents." (the italics also were in the original).

vishal0071:
1. Most of our implicit broadcasts are not in manifest except the FTP server one.
2. I tested it now, it doesn't require any extra work from our side.
There are still many things left.

EmmanuelMess:
* I don't think we are calling `List.sort()` anywhere, they are all `Collections.sort()`.
* Are you sure the FTP sever has implicit broadcast receivers? They seem to explicit.
* What's "Content change notifications"?
* How has "permissions" changed since API 25?

EmmanuelMess:
Created independent branch to work on this. See [Project#1](https://github.com/TeamAmaze/AmazeFileManager/projects/1).

PR #767, opened 2017-09-08
Immutable class PasteHelper added

EmmanuelMess:
This class maintains the paths to be pasted after cut or copy, it aldo contains the operation (cur or copy).


vishal0071:
Awesome (:

PR #779, opened 2017-09-20
Fix dissapearing appbar title bug

EmmanuelMess:
Fix disappearing appbar title bug (#778) and simplified some code related to appbar title.


EmmanuelMess:
By getter you mean `getResources().getString(@StringRes)`->`toolbar.setTitle(@StringRes)`? If so, no I don't know the disadvantages, do tell me.

PR #789, opened 2017-09-30
Optimized files loading via callbacks

EmmanuelMess:
* I tried to make file loading faster via deleting the for loop that fills a list with `HybridParcelableFiles` (if you know anything at all about overheads in passing interfaces that can be reduced, please do tell me).
* Added a library that deals with concurrent radix trees, for more efficient checking if file is hidden, and implemented.
* Replaced `BitmapDrawable` with `Drawable` that is cheaper to get.


VishalNehra:
Great work! Couldn't have done better myself! (:  Sorry for taking too long to merge.

PR #817, opened 2017-11-05
ELEMENTS_LIST to singleton and truly thread safed

EmmanuelMess:
Needs some testing but this is thread safe and should help optimize LayoutElement's loading later.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/adapters/RecyclerAdapter.java
+++ b/app/src/main/java/com/amaze/filemanager/adapters/RecyclerAdapter.java
@@ -36,6 +36,7 @@ import com.amaze.filemanager.utils.provider.UtilitiesProviderInterface;
 import com.amaze.filemanager.utils.theme.AppTheme;
 
 import java.util.ArrayList;
+import java.util.List;
 
 /**
  * This class is the information that serves to load the files into a "list" (a RecyclerView).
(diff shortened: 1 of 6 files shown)


VishalNehra:
Why are you creating LayoutElementsManager a singleton class?  We have two instances of MainFragment, you forgot? Two separate instances of LayoutElements are there.

EmmanuelMess:
I knew I forgot something; I'm an idiot, lel.

Alexander--:
@EmmanuelMess somehow it feels like you are still confused about thread-safety of that code. Any reading from ArrayList is considered "iteration". Copying to another ArrayList involves iteration too. So, according to the newly added comment, you should have surrounded `new ArrayList<>(layoutElements)` with some kind of synchronization block...

IMO, the design of that code is simply hopeless.

EmmanuelMess:
@Alexander-- It isn't the best idea, but it isn't absolutely broken either. Each get() from the thread safe list is atomic, so you can't have "half an item". Items added at the last moment (to the thread safe list) would be added or not (to the copy) probabilistically, which isn't much of a problem. Do correct me if I'm wrong.

VishalNehra:
What he means is while we're iterating list, if there's another thread performing other operation, it would the consistency. So we need to explicitly use synchronized block. Now you could create a wrapper around that, but it isn't necessary according to me.
This code is just the basics of creating a thread safe class. It isn't meant for our usecase at all.

EmmanuelMess:
I understand, if I make another PR with this code I'll fix that.

Alexander--:
@EmmanuelMess By "hopeless" I meant the design, not just implementation. If it needs substantial mental gymnastics to prove it's correctness, it is already broken in terms of design (and will become truly broken anytime you make a mistake because of that).

Here is how to make it much simpler and *obviously* thread-safe: do not share the list between threads. It is already being copied/re-wrapped it all over the place, so there is not point in trying to share it anymore. Remove/rewrite any code, that still accesses ELEMENTS_LIST from non-main threads. Do not allow such access anymore. That will make everything nice and thread-safe.

EmmanuelMess:
@Alexander-- I am working on that, thanks for the complete answer though.

PR #822, opened 2017-11-10
Fixed FileObserver

EmmanuelMess:
* Moved CustomFileObserver to it's own class, used Handler to comunicate and moved FileObserver instantiation to main thread.
* Created new polling hack for Marshmallow as a workarround for a [bug](https://stackoverflow.com/questions/36237314/fileobserver-and-contentobserver-not-working-in-android-marshmallow).
* FileObserver only restarts if needed and stops in onPause(), restarting in onResume().
* LIST_ELEMENTS has given up on thread safety. And simplified code relating the LIST_ELEMENTS to the adapter, also deleted commented lines.
* Fixed a NullPointerException when there are hidden files in directory.

### Known issues:
* On Marshmallow, if you delete a file named X and then create a folder named X, between one poll and the next (5s), FileObserver will not call onEvent(). This issue can be solved, but would require a small overhead that probably isn't necessary, please comment if you think otherwise.
* When adding new files/folders to the list, they will appear at the end, not ordered.


VishalNehra:
Thank you for taking time to fixing this major bug!

There was a minor problem where many threads were racing for sendMesaages callback method, resulting in multiple copies of elements to be posted in list. This was happening because the boolean was being set to false in sendMessages, and it was possible for another thread to context switch just after setting that value and before the check in onEvent and receive the false'd value and access sendMessages even when another thread was still there.

PR #823, opened 2017-11-11
Clarity , optimization, doc fix

Pavneet-Sing:
Use null instead of empty string , no need to match both conditions in if statement

### DIFF
--- a/app/src/play/java/com/amaze/filemanager/activities/MainActivity.java
+++ b/app/src/play/java/com/amaze/filemanager/activities/MainActivity.java
@@ -488,7 +488,7 @@ public class MainActivity extends ThemedActivity implements
             mDrawerToggle.setHomeAsUpIndicator(R.drawable.ic_drawer_l);
         }
         //recents header color implementation
-        if (SDK_INT >= 21) {
+        if (SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
             ActivityManager.TaskDescription taskDescription = new ActivityManager.TaskDescription("Amaze",
                     ((BitmapDrawable) getResources().getDrawable(R.mipmap.ic_launcher)).getBitmap(),
                     getColorPreference().getColor(ColorUsage.getPrimary(MainActivity.currentTab)));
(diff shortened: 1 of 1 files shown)


app/src/play/java/com/amaze/filemanager/activities/MainActivity.java
```diff
@@ -549,11 +549,11 @@ public Void onPostExecute(Object result) {
                             if (file.isDirectory(MainActivity.this))
                                 goToMain(path);
                             else {
-                                goToMain("");
+                                goToMain(null);
```
EmmanuelMess: For anyone wanting to check if this will throw an exception, [it won't](https://github.com/TeamAmaze/AmazeFileManager/blob/5d9e9aa8e89c6a09451cc9231723522ce59da15c/app/src/play/java/com/amaze/filemanager/activities/MainActivity.java#L883).

app/src/play/java/com/amaze/filemanager/activities/MainActivity.java
```diff
@@ -631,7 +631,7 @@ private void initFabToSave(final ArrayList<Uri> uris) {
      * of possible deadlock situation and the asynchronous behaviour of LibSuperSU
      */
     private void initializeInteractiveShell() {
-        // only one looper can be associated to a thread. So we're making sure not to create new
+        // only one looper can be associated to a thread. So we are making sure not to create new
```
EmmanuelMess: This is a bit excessive...

Pavneet-Sing: yeah ,don't like shorthand terms in docs/comments, gives a little pause  :)

Pavneet-Sing:
@VishalNehra on one thing i want to emphasis on i.e. 

```
 new Thread(() -> {
            if (mGoogleApiClient != null)
                mGoogleApiClient.connect();
}).run();

``` 
so either it should be start to make a thread
```

 new Thread(() -> {
            if (mGoogleApiClient != null)
                mGoogleApiClient.connect();
}).start();

```
or just remove the thread code because it accomplishing nothing here , (although seems like sort for testing purpose)
let me know what you think

PR #824, opened 2017-11-13
optimization, removed uneefective code

Pavneet-Sing:
calling run on thread will not create a thread so it has no effect rather just filling up the heap

### DIFF
--- a/app/src/play/java/com/amaze/filemanager/activities/MainActivity.java
+++ b/app/src/play/java/com/amaze/filemanager/activities/MainActivity.java
@@ -1640,30 +1640,25 @@ public class MainActivity extends ThemedActivity implements
     @Override
     public void onConnectionSuspended(int i) {
         Log.d("G+", "Connection suspended");
-        new Thread(() -> {
-            if (mGoogleApiClient != null)
-                mGoogleApiClient.connect();
-        }).run();
+        if (mGoogleApiClient != null)
+            mGoogleApiClient.connect();
     }
 
     public void onConnectionFailed(final ConnectionResult result) {
         Log.d("G+", "Connection failed" + result.getErrorCode() + result.getErrorCode());
         if (!mIntentInProgress && result.hasResolution()) {
-            new Thread(() -> {
-                try {
-                    mIntentInProgress = true;
-                    startIntentSenderForResult(result.getResolution().getIntentSender(),
+            try {
+                mIntentInProgress = true;
+                startIntentSenderForResult(result.getResolution().getIntentSender(),
                             RC_SIGN_IN, null, 0, 0, 0);
-                } catch (IntentSender.SendIntentException e) {
-                    // The intent was canceled before it was sent.  Return to the default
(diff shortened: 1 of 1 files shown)


app/src/play/java/com/amaze/filemanager/activities/MainActivity.java
```diff
@@ -1640,30 +1640,25 @@ public void onErrorResponse(VolleyError error) {
     @Override
     public void onConnectionSuspended(int i) {
         Log.d("G+", "Connection suspended");
-        new Thread(() -> {
-            if (mGoogleApiClient != null)
-                mGoogleApiClient.connect();
-        }).run();
+        if (mGoogleApiClient != null)
+            mGoogleApiClient.connect();
```
EmmanuelMess: If this does any networking it will crash, have you tested?

Pavneet-Sing: I am unable to do it , couple of files are missing like "BaseFile" etc many Files has been renamed (with still old import), although as i mentioned during previous commit 

invoke run won't create a new thread it just call the run function of runnable passed as lambda on UI thread so there is no threading with `.run();` if you want threading then it should use `.start();`

        new Thread(() -> {
            if (mGoogleApiClient != null)
                mGoogleApiClient.connect();
        }).start();
So currently lambda thread modules are running everything on UI.
Here is a [Small test that you can use to verify the behavior](https://gist.github.com/Pavneet-Sing/23bd944e322ebf02f32ac2cfa9c53fee)

VishalNehra:
Make sure you're compiling play-debug

app/src/play/java/com/amaze/filemanager/activities/MainActivity.java
EmmanuelMess: Change your build flavour to playDebug.
Maybe ´mGoogleApiClient.connect();´was never being reached so the exception never poped up, but if it would have run, it would have chashed; nevermind though, as connect() doesn't seem to make the connection from the thread it was called from.
Just a small question though, are you using AS 3.0?

Pavneet-Sing: yes i am using 3.0 and if that's the case then you should close the PR , thanks for reply

EmmanuelMess: No need to close anything, it's OK.

Pavneet-Sing: Ok good , Frankly, i was little free and I've been doing dry run (for fun but with caution ), i imported the project after your first comment and then created a little test ,  so if there is anything problematic then feel free to do anything , i''ll make the project up and running , keeping your suggestion  in mind :)

VishalNehra:
The implementation seems to be broken, as the profile/cover pic are not being fetched successfully. However, this PR does have a point, so I'll merge.

PR #829, opened 2017-11-19
The file lists reload in onResume()

### LINKED ISSUE #631: If directory is changed while out of the app, app doesn't reflect
Open a directory, exit app, change directory (it may be an automated action by another app in the background), open app, directory is as before, reload dir, now changes can be seen.

EmmanuelMess:
Fixes #631.


VishalNehra:
Not a good approach, as list always reloads when user returns, leaving user to the top of the list again, which is a bad UX.
A better fix would be to work on a background thread to reload list while still showing current list to user, then making changes as per the new data to an existing adapter.
Already working on this [here](https://github.com/TeamAmaze/AmazeFileManager/compare/synchronize_files_list).

EmmanuelMess:
Will take note.

Issue #851, opened 2017-12-01
Crash when trying to add new SMB source


jpage4500:
to reproduce:
- start app
- click on "+" button (FAB)
- hit SMB
- app crashed

Testing on S7 running 7.0

NOTE: this is because of the progress spinner was trying to reference holder.txtTitle for the progressbar which is null

EmmanuelMess:
Logcat:
```
12-01 13:50:27.722 com.amaze.filemanager E/AndroidRuntime: FATAL EXCEPTION: main
                                                           Process: com.amaze.filemanager, PID: 16236
                                                           java.lang.NullPointerException: Attempt to invoke virtual method 'void android.widget.TextView.setText(java.lang.CharSequence)' on a null object reference
                                                               at com.amaze.filemanager.ui.dialogs.SmbSearchDialog$ListViewAdapter.onBindViewHolder(SmbSearchDialog.java:163)
                                                               at com.amaze.filemanager.ui.dialogs.SmbSearchDialog$ListViewAdapter.onBindViewHolder(SmbSearchDialog.java:116)
                                                               at android.support.v7.widget.RecyclerView$Adapter.onBindViewHolder(RecyclerView.java:6356)
                                                               at android.support.v7.widget.RecyclerView$Adapter.bindViewHolder(RecyclerView.java:6389)
                                                               at android.support.v7.widget.RecyclerView$Recycler.tryBindViewHolderByDeadline(RecyclerView.java:5335)
                                                               at android.support.v7.widget.RecyclerView$Recycler.tryGetViewHolderForPositionByDeadline(RecyclerView.java:5598)
                                                               at android.support.v7.widget.RecyclerView$Recycler.getViewForPosition(RecyclerView.java:5440)
                                                               at android.support.v7.widget.RecyclerView$Recycler.getViewForPosition(RecyclerView.java:5436)
                                                               at android.support.v7.widget.LinearLayoutManager$LayoutState.next(LinearLayoutManager.java:2224)
                                                               at android.support.v7.widget.LinearLayoutManager.layoutChunk(LinearLayoutManager.java:1551)
                                                               at android.support.v7.widget.LinearLayoutManager.fill(LinearLayoutManager.java:1511)
                                                               at android.support.v7.widget.LinearLayoutManager.onLayoutChildren(LinearLayoutManager.java:595)
                                                               at android.support.v7.widget.RecyclerView.dispatchLayoutStep2(RecyclerView.java:3583)
[...]

PR #852, opened 2017-12-01
- fix crash when adding SMB Connection

jpage4500:
> SmbSearchDialog was trying to reference holder.txtTitle for the progressbar which is Null
- center progress dialog when searching for SMB Connection

fixes https://github.com/TeamAmaze/AmazeFileManager/issues/851


EmmanuelMess:
This seems OK, although I haven't a woking SMB setup right now, so I can't test connections.

Issue #854, opened 2017-12-01
App Crash - ConcurrentModificationException


jpage4500:
not sure exactly how this occurred but I do see and have fixed the issue as described below:

Testing on S7 running 7.0

```
11-30 19:47:49.009    com.amaze.filemanager    27621    -    E    AndroidRuntime     java.util.ConcurrentModificationException
11-30 19:47:49.009    com.amaze.filemanager    27621    -    E    AndroidRuntime         at java.util.HashMap$HashIterator.nextEntry(HashMap.java:851)
11-30 19:47:49.009    com.amaze.filemanager    27621    -    E    AndroidRuntime         at java.util.HashMap$EntryIterator.next(HashMap.java:891)
11-30 19:47:49.009    com.amaze.filemanager    27621    -    E    AndroidRuntime         at java.util.HashMap$EntryIterator.next(HashMap.java:890)
11-30 19:47:49.009    com.amaze.filemanager    27621    -    E    AndroidRuntime         at com.amaze.filemanager.ui.icons.IconHolder$1.processResult(IconHolder.java:110)
11-30 19:47:49.009    com.amaze.filemanager    27621    -    E    AndroidRuntime         at com.amaze.filemanager.ui.icons.IconHolder$1.handleMessage(IconHolder.java:89)
11-30 19:47:49.009    com.amaze.filemanager    27621    -    E    AndroidRuntime         at android.os.Handler.dispatchMessage(Handler.java:102)
11-30 19:47:49.009    com.amaze.filemanager    27621    -    E    AndroidRuntime         at android.os.Looper.loop(Looper.java:154)
11-30 19:47:49.009    com.amaze.filemanager    27621    -    E    AndroidRuntime         at android.app.ActivityThread.main(ActivityThread.java:6642)
11-30 19:47:49.009    com.amaze.filemanager    27621    -    E    AndroidRuntime         at java.lang.reflect.Method.invoke(Native Method)
11-30 19:47:49.009    com.amaze.filemanager    27621    -    E    AndroidRuntime         at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1468)
11-30 19:47:49.009    com.amaze.filemanager    27621    -    E    AndroidRuntime         at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1358)
[...]

PR #855, opened 2017-12-01
- fix crash ConcurrentModificationException : IconHolder$1.processRes…

jpage4500:
…ult(IconHolder.java:110)

  > use iterator to remove from map
  > if going to synchronize, need to put it around all access of map and not just 1 place

fixes issue: https://github.com/TeamAmaze/AmazeFileManager/issues/854

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/ui/icons/IconHolder.java
+++ b/app/src/main/java/com/amaze/filemanager/ui/icons/IconHolder.java
@@ -48,9 +48,9 @@ import com.amaze.filemanager.utils.Utils;
 import com.cloudrail.si.interfaces.CloudStorage;
 
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.Map;
-import java.util.Objects;
 
 import jcifs.smb.SmbFileInputStream;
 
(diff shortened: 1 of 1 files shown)


EmmanuelMess:
Instead of this: 
```
synchronized (mRequests) {
      this.mRequests.clear();
}
```
Use `Collections.syncronizedList(new ArrayList<>)` for the list resulting in those operations being atomic.

app/src/main/java/com/amaze/filemanager/ui/icons/IconHolder.java
```diff
@@ -100,29 +100,20 @@ private synchronized void processResult(LoadResult result) {
...
             // find the request for it
             synchronized (mRequests) {
-
-                for (Map.Entry<ImageView, String> entry : mRequests.entrySet()) {
+                Iterator<Map.Entry<ImageView, String>> iterator = mRequests.entrySet().iterator();
+                while (iterator.hasNext()) {
+                    Map.Entry<ImageView, String> entry = iterator.next();
                     final ImageView imageView = entry.getKey();
                     final String fso = entry.getValue();
-                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
```
EmmanuelMess: Why did you remove this line?

EmmanuelMess: nevermind

app/src/main/java/com/amaze/filemanager/ui/icons/IconHolder.java
```diff
@@ -100,29 +100,20 @@ private synchronized void processResult(LoadResult result) {
...
 
             synchronized (mAppIcons) {
-
                 mAppIcons.put(filePath, result.result);
             }
 
             // find the request for it
             synchronized (mRequests) {
-
-                for (Map.Entry<ImageView, String> entry : mRequests.entrySet()) {
```
EmmanuelMess: Here the code the same to what you wrote, you just extended it. Read [here](https://docs.oracle.com/javase/1.5.0/docs/guide/language/foreach.html).

PR #858, opened 2017-12-07
Revert "- fix crash ConcurrentModificationException : IconHolder$1.processRes…"

EmmanuelMess:
Reverts TeamAmaze/AmazeFileManager#855.
I had made some comments on the pull request, it has some redundancy (foreach replaced with the same thing in a while loop) and syncronizes List.add() and List.put() instead of using Collections.synchronizedList().


VishalNehra:
Please make the changes you mentioned in a new PR, I accepted this as it synchronizes ````cancelLoad```` method which was originally missing.

EmmanuelMess:
Very well

PR #862, opened 2017-12-09
Made most Parcelables mostly immutable

EmmanuelMess:
So that one can get fields directly instead of using getter.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/adapters/CompressedExplorerAdapter.java
+++ b/app/src/main/java/com/amaze/filemanager/adapters/CompressedExplorerAdapter.java
@@ -195,51 +195,50 @@ public class CompressedExplorerAdapter extends RecyclerView.Adapter<CompressedIt
             holder.checkImageView.setBackgroundDrawable(new CircleGradientDrawable(compressedExplorerFragment.accentColor,
                     utilsProvider.getAppTheme(), compressedExplorerFragment.getResources().getDisplayMetrics()));
 
-        if (rowItem.getType() == CompressedObjectParcelable.TYPE_GOBACK) {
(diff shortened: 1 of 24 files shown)


VishalNehra:
This is against the [best practices](http://www.codejava.net/coding/java-getter-and-setter-tutorial-from-basics-to-best-practices) for java style coding, and shall not be utilized wherever possible.
However, you can try to tackle some problems mentioned with our implementation of getter/setter as given in this link.

EmmanuelMess:
You can't change a final value, setters make no sense, and getters are redundant, why would you need them? Check Pair and Java immutables

VishalNehra:
@EmmanuelMess that is purely subjective and depends on the requirement. 
For GUI programming, like ours, it's always better to keep objects mutable.

EmmanuelMess:
@VishalNehra Why? I makes for less code to have immutables.

VishalNehra:
Really? You'll have to create new objects each time user make any interaction with GUI. Not to mention, cloning existing data to new object will just extend the code more.  While a simple setter method can do this job. I do agree immutables are really helpful for concurrent programming, hence will be useful for some parts of our project, but that doesn't mean you make all your objects immutable. 
There's a thing called MVC for a reason. Otherwise they would also have simply used immutables instead of creating models for each component.

EmmanuelMess:
There's no need for setters in the majority of these. They are just created to hold some data and then get destroyed.

EmmanuelMess:
Maybe I can just Kotlinize... That sounds better?

VishalNehra:
Alright, can be pulled, but please rebase this branch first.

EmmanuelMess:
Done.

Issue #498, opened 2017-03-29
Add support for SFTP client


X-Ryl669:
There is another issue about a FTP client where it's said that SFTP client was on the way should be almost done.

I think it's easier to have separate issue, so anyway, can you add a SFTP (SSH) client so it's possible to connect to remote linux server's file tree over internet? 

Thanks

TranceLove:
As a refugee from another famous file manager app I'm interested in having this feature available in Amaze. However I also see some refactoring at the current codebase is necessary since the code related to remote server connection (at least at open source side) is mostly SMB oriented.

May see what I can do when I have time.

vishal0071:
@TranceLove I refractored most of the code to support multiple file types in past so it's really easy to implement new file types now...  And as a matter of fact, I've started working on a FTP/FTPS client implementation just now. So rest assured, it's coming very soon (:
BTW I don't think SFTP is worth it, as an SSH protocol is supposed to be used to tunnel authentication via secure medium, it's not ideal to establish secure data connection. Correct me if I'm wrong.
Also, I'm using Apache Commons net API to implement FTP/FTPS client, it doesn't support SFTP client afaik.

Limero:
SFTP is definitely worth it just by the convenience factor alone. It's working really well and only requires SSH to be installed on the server (which most distros do by default).

TranceLove:
I share the same thought with @Limero.

> I'm using Apache Commons net API to implement FTP/FTPS client, it doesn't support SFTP client afaik.

@vishal0071 Yes, you need external library like jsch or sshj to implement this.

vishal0071:
Very well, I'll include a SFTP client connection too! Thanks @TranceLove for pointing me to the library (:

scanno:
Is there any progress on implementing the SFTP client?

Using SFTP is the best way of accessing files Linux machines remotely. Using SFTP on a daily basis.

TranceLove:
I did had some groundwork based on the codebase of the last release, only been setback 'cos of my daytime work.

It had POC of integration with jssh, connection setup dialog, but it's far from complete; and hence, it's not even on my own fork on github yet.

Wouldn't give out any hard promise based on my current situation... but I'm trying to make way for it because I need it too. So let's say, I'll let Santa Claus test it before he sends it to your device? Adding the necessary work merge, code review, (initial) bugfixing... Christmas should be about the time.

VishalNehra:
That's great! @TranceLove 
[...]

Issue #856, opened 2017-12-01
Unable to copy multiple files


jpage4500:
Using an S7 running 7.0

- connect to SMB source
- select multiple files and hit Copy
- change to external SD card and hit Paste

I added some log statements to see what was happening.. only the first selected file gets copied

```
12-01 10:33:15.536	com.amaze.filemanager	8230	-	D	PrepareCopyTask	 startService: COPY: smb://192.168.0.100/MEDIA/TV/testfile.mkv to /storage/29A4-1410/MEDIA/TV
12-01 10:33:15.537	com.amaze.filemanager	8230	-	D	PrepareCopyTask	 startService: COPY: smb://192.168.0.100/MEDIA/TV/testfile2.mkv to /storage/29A4-1410/MEDIA/TV
...
12-01 10:33:16.016	com.amaze.filemanager	8230	8360	D	GenericCopyUtil	 startCopy: smb://192.168.0.100/MEDIA/TV/testfile.mkv (SMB) to /storage/29A4-1410/MEDIA/TV/testfile.mkv (FILE)
```

log statements can be seen here: https://github.com/jpage4500/AmazeFileManager/commit/9ffce62d05137ecbf2b184bb0e4355906d5cc60a

jpage4500:
here's where I *think* the issue is occurring.. in CopyService it iterates over each file to copy but there's a check here that's only true the first time through. I haven't dug into the code to figure out why this is here or where the other files should be copied over (I assume the idea is to wait until the first file is copied and *then* start the next one.. but it never happens)

```
                    for (int i = 0; i < sourceFiles.size(); i++) {
                        sourceProgress = i;
                        HybridFileParcelable f1 = (sourceFiles.get(i));

...
                            if (!progressHandler.getCancelled()) {
```

VishalNehra:
Is this still an issue?

PR #863, opened 2017-12-09
SFTP client support

TranceLove:
This is in response to #498.

Before going further, some disclaimer:
* For the purists: I was born as Java developer with dedication to server-side stuff (my past jobs were mainly work with Spring Framework (up to 3.0 though)), so my programming patterns may not be compatible with Android's, as well as some decisions I made may not be correct or appropriate
* I rely heavily on Android's API and Android Studio's feedbacks to tell me I was doing things wrong. I may have missed something I don't know about with Android
* My work was started on 3.1.2RC4 codebase, I may not have take advantage of some Java8 features available to Android (e.g. Lambdas which I saw on some code during merge)
* For the same reason, as well as I didn't merge with upstream from time to time, I may have missed some API changes on `master`. I'd say I did my best to make the merge, but I may still missed something
* I used the current SMB client's behaviour as reference, may become the cause of some unexpected behaviours in my implementation
* Sorry about the number of commits I made - I used small commits to track what did I do with the code, as well as I was working on different machines on implementing it
* Both SFTP and SSH are used in code and comments, they mean the same thing

Then, **couple of change warnings before this pull request is even opened**:
* To deal with SSH keys(public/private, client/server sides) properly - [SpongyCastle](https://github.com/rtyley/spongycastle) is used instead of Android's stock BouncyCastle (which was already long outdated). Therefore **the resulting APK had hit Dalvik's 65k method limit and multidex had been enabled**. The last time I build the master the APK was around 57k, so I think it's a matter of time hit the limit anyway
* Because of the way I wrote my SSH client implementation to reduce complexity and coupling, **I used `AppConfig` to access `UtilsHandler` and `MainActivity`'s `Context`**

Now back to the pull request.

As said in #498, my implementation was an integration with [sshj](https://github.com/hierynomus/sshj) which I saw it as simpler and a better alternative than [jsch](http://www.jcraft.com/jsch/).

The connection setup dialog was basically a rip-off from `SmbConnectDialog` ;) couldn't find an icon handy, so I used the Tux icon to represent SSH servers.

It supports both password and key authentication. I tested it against my Ubuntu Xenial machines (home desktop and VMs) on Lollipop and Nougat emulators.

Upload, download, mkdir, delete, rename should be working properly.

Now let the cruel code review begin... wish me luck.


VishalNehra:
Really nice and clean implementation.
Thanks for all the hard work (:
Please also define what you feel to be the areas of improvement or if there are any bugs so that we can also test and help.

VishalNehra:
Pl. increase the ````DATABASE_VERSION```` in ````UtilsHandler```` otherwise it'll crash for existing versions.

EmmanuelMess:
That'll clean the database, edit the onUpdate() accordingly, then bump the database version

app/src/main/java/com/amaze/filemanager/filesystem/ssh/tasks/AsyncTaskResult.java
```diff
@@ -0,0 +1,50 @@
...
+package com.amaze.filemanager.filesystem.ssh.tasks;
+
+/**
+ * Container for AsyncTask results. Allow either result object or exception to be contained.
+ *
+ * @param <T> Result type
+ */
+
+public class AsyncTaskResult<T> {
+    private T result;
```
EmmanuelMess: Immutable could be used here. Make fields public final and delete getters.

app/src/main/java/com/amaze/filemanager/filesystem/ssh/tasks/GetSshHostFingerprintTask.java
```diff
@@ -0,0 +1,112 @@
...
+ * connection settings.
+ *
+ * @see HostKeyVerifier
+ * @see SSHClient#addHostKeyVerifier(String)
+ * @see com.amaze.filemanager.ui.dialogs.SftpConnectDialog#onCreateDialog(Bundle)
+ */
+
+public class GetSshHostFingerprintTask extends AsyncTask<Void, Void, AsyncTaskResult<PublicKey>>
+{
+    final String mHostname;
```
EmmanuelMess: Private?

EmmanuelMess:
Can you rebase into less commits?

TranceLove:
@EmmanuelMess will do. Let me get back to you once it's done.

@VishalNehra 
Known issues:
* app will hang if device does not have enough storage space for the files being copied. I saw the same symptom when copying with SMB (current master), so I think it's the problem within Amaze itself :/
* possibly unknown behaviour if user wants to switch from password to key authentication, or vice versa, on existing SSH connection setups. Best bet is to create a new connection setup. Maybe fixed later on, but keep note on it
* For SFTP targets, disk space calculation, MD5/SHA hash calculations rely on running command on remote server (over SSH, of course). During development it had cases that disk space calculation did not return the result in time. Didn't crash the app, but doesn't look good
* maybe you'll want a new icon for SSH servers? By the way, I saw the Tux is smaller than it was when running on my Oneplus One (custom 7.0) and Galaxy S2 (custom 5.1); emulator does not have this problem

app/src/main/java/com/amaze/filemanager/database/UtilsHandler.java
```diff
@@ -79,13 +86,22 @@ public void onCreate(SQLiteDatabase db) {
...
         db.execSQL(queryHistory);
         db.execSQL(queryHidden);
         db.execSQL(queryList);
         db.execSQL(queryGrid);
         db.execSQL(queryBookmarks);
         db.execSQL(querySmb);
+        db.execSQL(querySftp);
     }
 
     @Override
```
EmmanuelMess: Change onUpgrade()'s code to:
```
    @Override
    public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
        switch(newVersion) {
            case 2:
                db.execSQL("CREATE TABLE IF NOT EXISTS " + TABLE_SFTP + " ("
                        + COLUMN_ID + " INTEGER PRIMARY KEY,"
                        + COLUMN_NAME + " TEXT,"
                        + COLUMN_PATH + " TEXT,"
                        + COLUMN_HOST_PUBKEY + " TEXT,"
                        + COLUMN_PRIVATE_KEY_NAME + " TEXT,"
                        + COLUMN_PRIVATE_KEY + " TEXT"
                        + ")");
        }
    }
```

And increment the `DATABASE_VERSION` to 2.

EmmanuelMess:
If you enter the wrong IP it seems that the dialog hangs.

TranceLove:
@EmmanuelMess requested changes committed in e0a0814 (if I understand correctly)

And app hang... do need to find a way to handle this. Will get back to you once it's solved.

TranceLove:
Additionally, squashed commits count down (to 70 currently).

app/src/main/java/com/amaze/filemanager/database/UtilsHandler.java
VishalNehra: Why would you change onUpgrade code to that? Consider a change in schema later and you would need to rewrite the code to drop and create those tables again . And creating table in onUpgrade is redundant, simply drop the table if exists and call ````onCreate```` again.

EmmanuelMess: That would delete any old data. Why would a schema change make you have to drop the tables? If you don't need some data, sure drop it, but if it is still needed why would you drop it?

EmmanuelMess: **Ok, I made a mistake in the code above.**

For version 2 the code would be:
´´´
    @Override
    public void onUpgrade(SQLiteDatabase db, int **oldVersion**, int newVersion) {
        switch(**oldVersion**) {
            case 1:
                db.execSQL("CREATE TABLE IF NOT EXISTS " + TABLE_SFTP + " ("
                        + COLUMN_ID + " INTEGER PRIMARY KEY,"
                        + COLUMN_NAME + " TEXT,"
                        + COLUMN_PATH + " TEXT,"
                        + COLUMN_HOST_PUBKEY + " TEXT,"
                        + COLUMN_PRIVATE_KEY_NAME + " TEXT,"
                        + COLUMN_PRIVATE_KEY + " TEXT"
                        + ")");
        }
    }
´´´
PS: No break, so that if you upgrade from 1 to n you get all changes needed.

VishalNehra: Alright, what would you suggest then, if I add a new column in a table, what do I do with existing rows? What entries will they contain for that column in their row?

TranceLove: Thanks for the explanation and clarifications. See 6a4951e if can suit your needs.

TranceLove:
To fix app hang on setting up the SSH connection, I moved the procedure of getting SSH host key fingerprint to run non-blocking (no `AsyncTask.execute().get()`) but with a callback ala JS style.

For SSH authentication, it is happening asynchronously anyway, so I didn't change it.

And added connection timeouts to 30 secs, in align with SMB client.

Finally, added a ProgressDialog on getting the SSH host key fingerprint. It's deprecated API, but seems the other parts of Amaze is still using, so I'll just follow suit.

app/src/main/java/com/amaze/filemanager/database/UtilsHandler.java
EmmanuelMess: @VishalNehra you need to add a new column? 
There are 2 possibilities:
* first, from the information on the other ones (and the device) you'll be able to reconstruct the column for old entries; 
* second, the user must enter the information manually, but, because the user doesn't want to enter 100 values, you must put a special value (on each cell of the new column) and work arround it, you add the new data for new entries.

app/src/main/java/com/amaze/filemanager/database/UtilsHandler.java
```diff
@@ -86,18 +102,18 @@ public void onCreate(SQLiteDatabase db) {
         db.execSQL(queryGrid);
         db.execSQL(queryBookmarks);
         db.execSQL(querySmb);
+//        db.execSQL(querySftp);
```
EmmanuelMess: I think you need this line to create the database for new users...

TranceLove: Oops...

EmmanuelMess:
Ok, probably needs a Toast to tell the user that the connection failed. Can't really test the actual connection but I'll give it the visto bueno.

VishalNehra:
I am failing to stream files, can you confirm if you implemented the support for streaming (on local host as we do in SMB/cloud or from this API itself)?

VishalNehra:
Alright... Regardless, it won't reach to that, as we've already implemented a standard http streaming API, you can simply implement that if there's no native support with the API you're using.

VishalNehra:
I'm trying to push commit but getting permission denied error @TranceLove 
Have you checked the tick for allowing adding changes to this PR?

TranceLove:
This tick?

screenshot from 2017-12-13 07-29-21

TranceLove:
CloudStreamer integration added in 5706b5d.

I also added support for PuTTY keys (*.ppk) and OpenSSH v1 keys (which has headers reading `BEGIN OPENSSH PRIVATE KEY`) in ee988af.

Note however, I found there are keys that can't be parsed by sshj but still can be used from Linux's `ssh`. I'd suppose if you are using newer key types, like those with smaller PEM file sizes having `ecdsa` in the header you'll most likely run into problems.

I thought sshj should have handled this, maybe just because I haven't read through the source codes, and there are already postings around the web suggesting their possible incompatibility.

Will look into it when I have time. Before it gets fixed, use RSA or DSA key types on the safe side.

TranceLove:
Acknowledged conflicts. Fix is on the way... need test before they appear here.

VishalNehra:
Can't seem to parse ppk PuTTY keys. Can you please confirm?

EmmanuelMess:
I would recommend not pulling this until the release 3.3.0 is published.

TranceLove:
Tested with keys generated with latest puttygen, codebase merged with upstream.

RSA works OK
DSA works OK
ECDSA parse failed
OpenSSHv1 parse failed
ED22519 parse failed

However, from [sshj's source code](https://github.com/hierynomus/sshj/blob/master/src/main/java/net/schmizz/sshj/userauth/keyprovider/PuTTYKeyFile.java#L79) and [sshj's test case](https://github.com/hierynomus/sshj/blob/master/src/test/java/net/schmizz/sshj/keyprovider/PuTTYKeyFileTest.java) it mentioned only RSA and DSA key types... so I think only PuTTY's RSA and DSA key types are supported. :/

Did you test with non-RSA/DSA keys generated by `puttygen`?

VishalNehra:
@TranceLove indeed. I tried with other GUI (WinSCP) they can't seem to parse their own generated PuTTY key :/
I'll give it a shot again, maybe I'm doing something wrong, I'm not used to using SFTP much.

VishalNehra:
@TranceLove still can't seem to parse neither OpenSSH generated RSA key, nor PuTTY generated RSA key.  Please find my private and public keys [here](https://drive.google.com/open?id=1DWPylxP3OSmclfpBUvIln7BVTpc-5slE)

````IOException reading PEM
                                                                         java.io.IOException: No converter available to parse selected PEM
                                                                             at com.amaze.filemanager.filesystem.ssh.tasks.PemToKeyPairTask.doInBackground(PemToKeyPairTask.java:110)
                                                                             at com.amaze.filemanager.filesystem.ssh.tasks.PemToKeyPairTask.doInBackground(PemToKeyPairTask.java:64)
                                                                             at android.os.AsyncTask$2.call(AsyncTask.java:333)
                                                                             at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:245)
                                                                             at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
                                                                             at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
                                                                             at java.lang.Thread.run(Thread.java:764)````

TranceLove:
Hmm... key with passphrase. Seems a bit hairy to tackle this... not to say difficult, but need to add something to the implementation... a dialog box to capture key passphrase, the way to store the passphrase, to begin with.

VishalNehra:
Ahh alright.. you didn't mention a key with passphrase isn't supported yet. 
My bit, don't just store it anywhere, asking the passphrase everytime during login would be better.

TranceLove:
OK, let's see e938da0 can help.

VishalNehra:
Most of the things are working correctly. Thanks for all your hard work. Much delay from my side, my apologies for that.
We're getting ready for beta versions so next production release should be soon.

FBachofner:
I just reinstalled Amaze after seeing that SFTP support had been added.

Even though I am registered as a beta tester, the Play Store serves up version 3.2.1 which does not seem to have the SFTP client built in.  What's the best way to test this?

TranceLove:
I shouldn't be the right person to answer this, but anyway...

For end users, you'll have to wait until the latest production version is released on Google Play and/or fdroid. (And in fact, I think they have a more serious problem to solve before release - see #887, #905, and possibly #856) 
For advanced users who don't mind getting hands dirty and know how to build apks, checkout 120dbc3 and build a `playDebug` apk. But be advised, current copy implementation has a few problems - see #905 and #887

Maybe they have beta channels for the impatient to play with, but I don't know.

EmmanuelMess:
@FBachofner [Here](https://github.com/TeamAmaze/AmazeFileManager/releases) are the releases, if a new one pops up you'll have the (new) changes in code up to the point of (that) release.

Limero:
Is there an ETA of when the next release/beta will be released? This is a fantastic thing to have been added.

PR #864, opened 2017-12-09
Added a black theme

PiwwowPants:
Structured the black theme around the current way of theme-ing.
Generally all checks are the same, with only background checks changing to apply the black color relative to the context that other colors are being applied.

* No maintenance was done on the theme management.
* The black theme has no dependencies on the dark theme, so the dark theme styles were just duplicated out to the black theme
* The black theme is identical to the dark theme with the major background colors changing to be solid black rather than holo dark


app/src/main/res/values/strings.xml
```diff
@@ -106,6 +106,7 @@
     <string name="light">Material Light</string>
     <string name="dark">Material Dark</string>
     <string name="daytime">Daytime</string>
+    <string name="black">Black</string>
```
EmmanuelMess: It would be better if you explain that it would reduce the battery consumption in AMOLED devices.

PiwwowPants: If you think that'd be fitting but it's a bit wordy. 
It's just a black theme on non-OLED-based screens. I don't know if it's worth explaining as if it's a feature of the theme when that's device dependent.

EmmanuelMess: If you can't add a smaller explanatory text (like "saves battery on AMOLED") "Black (for AMOLED)" is good enough.

VishalNehra: Pitch black does save battery for AMOLED devices... It's a universal truth... 
As for PR, I'll review later.
Thanks for the contribution @PiwwowPants

VishalNehra:
A few minor corrections and this will be merged! Thanks for your contribution ^_^ a lot of people were asking for a black theme.

Issue #875, opened 2017-12-15
Last point in ProcessWatcherFrament has copy speed that's incoherent


EmmanuelMess:
It always happens that the copy speed goes up→down→up→...→absurdly up.
<img height="250px" src="https://user-images.githubusercontent.com/10991116/34055386-b987415a-e1ad-11e7-82c9-ae81a04528a6.png" />

VishalNehra:
Not really...
There *is* this thing that the update interval is 1 sec throughout but at last sec... If progress completes before a second, then it's updated at that moment only instead of waiting for a whole second to update which I guess is an expected behavior.

EmmanuelMess:
Where's the waiting code?

VishalNehra:
There's no waiting code, I'm saying that we don't wait for a second to complete and post update whenever the progress actually completes. Maybe the copy speed actually was that much for you in that last second?

EmmanuelMess:
It happens always, I don't think it's my phone.

> There is this thing that the update interval is 1 sec throughout but at last sec...

Where's the waiting code for the "update interval of 1 sec".

VishalNehra:
In [ServiceWatcherUtil](https://github.com/TeamAmaze/AmazeFileManager/blob/master/app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java#L97)

EmmanuelMess:
One day, I'll have to edit that class...

EmmanuelMess:
Also happens encripting, I suppose it should be fixed, it doesn't seem real:

<img src="https://user-images.githubusercontent.com/10991116/34059106-5173437a-e1bc-11e7-8054-864d2d466b7a.png" height="250px"/>

VishalNehra:
Can't reproduce, please try to reproduce on a different device.

VishalNehra:
Alright, I reproduce this issue with the changes in your PR #876 
Can't reproduce on master.

EmmanuelMess:
Why?

EmmanuelMess:
Nope, on master also occurs, where did you copy the file? I did it on internal memory.

VishalNehra:
Yep, in internal memory. Can't reproduce on master but in your PR.

EmmanuelMess:
> Haven't really changed the way the code works in PR #876.

EmmanuelMess:
@TranceLove could you test?

TranceLove:
[...]

PR #876, opened 2017-12-15
Code cleanup for Services and ProgressViewerFragment

EmmanuelMess:
* Uses a generalizing abstract class and a simple binder child to reduce repeated code in services.
* Had to separate Encryption and Decryption. 
* Moved notification IDs to NotificationConstants.
* Fixed decryption failing bug.
* Better javadoc in DatapointParcelable.
* ProcessViewerFragment.ServiceType enum to int final vars (see [this](https://android.jlelse.eu/android-performance-avoid-using-enum-on-android-326be0794dc3)).
* Small optimization of chart data loading.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -169,6 +169,10 @@
             android:label="@string/crypt_encrypting"
             />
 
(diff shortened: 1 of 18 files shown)


TranceLove:
Accepting PR following discussion at #875.

PR #880, opened 2017-12-18
Vectorized drawer and bottom bar

EmmanuelMess:
* Vectorized drawer (used new, better icons, library type) and bottom bar.
* Deleted unused image.
* Rewrote Icons class.
* Deleted grid views.


VishalNehra:
There's no placeholder icon for images/videos/other generic types in case the thumbnail fails to load.

EmmanuelMess:
Where was that? The thing is I believe that the code keeps working as it did before, where's the 'transcription' error?

EmmanuelMess:
Could you retest (and not put issues that aren't created with this pr)?

EmmanuelMess:
As per @VishalNehra agreement this PR will be pulled.

PR #885, opened 2017-12-21
Fixed last drawer items not getting selected and aligned to Android Guidelines 

### LINKED ISSUE #298: Issues report, features request and UI/UX suggestions
Hi

First of all, thanks for this awesome explorer. I like the things it has, and how well designed it is.

Anyways, I would like to suggest and request some things. I didn't want to make a post for each thing as they will be kinda many.

Let's start:
#### 7. Documents icon is not from Material Design iconset. (I think is the only one not from there).

Please use [this icon](https://materialdesignicons.com/icon/file-document)

EmmanuelMess:
As per [Android Guidelines](https://material.io/guidelines/patterns/navigation-drawer.html#navigation-drawer-content) the drawer list should contain Settings and other formerly 'pinned' items.
Changed the way the items are handled (used a single immutable class) and documented drawer constants.
Fix #298 point 6.


VishalNehra:
I realized this after merging. Your hack probably introduced a bug. When you press 'recents' key on device, and resume Amaze later, it automatically sets the primary color of views to left panel even when you are on the right one.

EmmanuelMess:
Ok, that's why it's a hack, I will fix.

Issue #890, opened 2017-12-26
Glide doesn't show placeholder or failure images


EmmanuelMess:
When Glide is loading or when it failed to load it doesn't show a different image.

PR #902, opened 2018-01-04
Grid view now has rounded background behind icons

### LINKED ISSUE #894: Why do the file images show  with a background in list view and colorized in grid view?
Why do the file images show  with a background in list view and colorized in grid view?

Here are some of the lines that cause this:
On list view:
https://github.com/TeamAmaze/AmazeFileManager/blob/8a956c516843aeb4c5dee3d251997b29abd49784/app/src/main/java/com/amaze/filemanager/adapters/RecyclerAdapter.java#L566-L578
On grid view:
https://github.com/TeamAmaze/AmazeFileManager/blob/8a956c516843aeb4c5dee3d251997b29abd49784/app/src/main/java/com/amaze/filemanager/adapters/RecyclerAdapter.java#L627-L660

EmmanuelMess:
Fixes #894.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/adapters/RecyclerAdapter.java
+++ b/app/src/main/java/com/amaze/filemanager/adapters/RecyclerAdapter.java
@@ -29,6 +29,7 @@ import com.amaze.filemanager.ui.ItemPopupMenu;
 import com.amaze.filemanager.ui.icons.Icons;
 import com.amaze.filemanager.ui.icons.MimeTypes;
 import com.amaze.filemanager.ui.views.CircleGradientDrawable;
+import com.amaze.filemanager.ui.views.RoundedImageView;
 import com.amaze.filemanager.utils.GlideConstants;
 import com.amaze.filemanager.utils.Utils;
 import com.amaze.filemanager.utils.color.ColorUsage;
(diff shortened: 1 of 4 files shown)


VishalNehra:
Images and video thumbnails aren't circular. Also, the "use circular icons" option in settings has no effect here. Seems like you would need to keep both this circular, and the square one, sorry that I completely overlooked that option and failed to inform about it earlier.
Similar to the one you demonstrated [here](https://user-images.githubusercontent.com/10991116/34470770-ce4185a2-ef17-11e7-948e-f4314b565e35.png)

EmmanuelMess:
> Images and video thumbnails aren't circular.

Cicular images and videos are a waste of space, they look nice in list because list's main purpose is to provide action on files, but grid's primary function is to provide a preview (file name is ridiculously small and images use more than half the phisical space).

> Seems like you would need to keep both this circular, and the square one, sorry that I completely overlooked that option and failed to inform about it earlier.

I thought that the circular one looked nice on Photoshoped images, but implemented it doesn't seem as nice, will implement the "square" one too.

VishalNehra:
Hmm... Alright, not even worth a try you presume? There's an option just for that so it's upto user to decide if a square would look good on grid or circular.

EmmanuelMess:
No, there would be an option for the user to choose rounded in list AND grid. Also, square images is what Material Design recomends

EmmanuelMess:
This is a precursor to the PR that will fix #890.

PR #907, opened 2018-01-07
Added a new system to choose colors

EmmanuelMess:
New color selection system, needs more colors though. If you have suggestions leave them below.


VishalNehra:
Awesome implementation!
Can modify the green color choice's accent to something light as it's too dark to read.
And please don't add more colors, specially don't add a color picker. 
This many are just fine.
Ps. random skin doesn't seem to work too.

EmmanuelMess:
> specially don't add a color picker
There already is a color picker, choose "custom" and select colors.
> Can modify the green color choice's accent to something light as it's too dark to read.
There are also problems when the theme changes, good colors in white background aren't always the best in black backgrounds

VishalNehra:
I mean, don't add new colors or a hex color picker. What we have is just fine.

EmmanuelMess:
You misunderstood, other color combinations I meant.

VishalNehra:
Ahh, alright... I'll try some and let you know.

EmmanuelMess:
> Ps. random skin doesn't seem to work too.

You sure? just tested, it seems to work fine.

EmmanuelMess:
> Can modify the green color choice's accent to something light as it's too dark to read.

I think you should just change the colors in the palette, I chose from there and that's the best green I could find.

VishalNehra:
Let's leave more colors for later. I think this can be merged without problems.

EmmanuelMess:
very well

PR #818, opened 2017-11-08
Shortcut feature for devices >= 7.1

preet90:
Current way of creating shortcut is not working from android 7.1. So, I have updated the code to create dynamic shortcuts for android devices using 7.1 and later. After creating the shortcuts, user has to long press the app icon to see all the shortcuts. Image below.

https://developer.android.com/images/guide/topics/ui/shortcuts.png

https://github.com/TeamAmaze/AmazeFileManager/issues/771

PR #911, opened 2018-01-08
Android 8 support for launcher shortcuts

### LINKED ISSUE #771: No shortcut generated
Hi there. Recently updated Nexus 6P to Oreo. I'm trying to generate a shortcut to an audio file. I open Amaze, navigate to the file, and long press to select. Tapping the overflow menu, the drop down shows the option to "Add shortcut". Tapping that option, the program pauses or seems to do something, but no shortcut is created (at least, none that I can find)

Thanks!

dmitrio95:
This pull request is another option to fix an issue #771. It differs from the previously suggested fix (#818) mainly in two points:
1. It is implemented using appcompat and avoids explicit implementation of different behavior for different Android API levels;
2. It uses [pinned shortcuts](https://developer.android.com/reference/android/content/pm/ShortcutManager.html#requestPinShortcut%28android.content.pm.ShortcutInfo,%20android.content.IntentSender%29) instead of [dynamic shortcuts](https://developer.android.com/reference/android/content/pm/ShortcutManager.html#addDynamicShortcuts%28java.util.List%3Candroid.content.pm.ShortcutInfo%3E%29) used in #818. Pinned shortcuts are [recommended](https://developer.android.com/about/versions/oreo/android-8.0-changes.html#as) to use instead of those previously created by `INSTALL_SHORTCUT` broadcast and seem to provide the same user experience. Appcompat flavor of this feature uses the same broadcast to provide this feature on lower Android versions.

There is also a question concerning the state of license headers. It seems that most of files (at least those I worked with) contain a lot of modifications done, at least, in 2017 — 2018, while all the authors in license headers state their copyright only for 2014. As far as I understand, the copyright notice should contain information on all the years when modifications were done. Am I right with it or should I have just put my name in that list of 2014-year authors? For now, I left the license headers unchanged.
 
Fixes #771.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/fragments/MainFragment.java
+++ b/app/src/main/java/com/amaze/filemanager/fragments/MainFragment.java
@@ -43,6 +43,9 @@ import android.preference.PreferenceManager;
 import android.support.design.widget.AppBarLayout;
 import android.support.v4.app.FragmentActivity;
 import android.support.v4.app.FragmentManager;
+import android.support.v4.content.pm.ShortcutInfoCompat;
+import android.support.v4.content.pm.ShortcutManagerCompat;
+import android.support.v4.graphics.drawable.IconCompat;
 import android.support.v4.widget.DrawerLayout;
 import android.support.v4.widget.SwipeRefreshLayout;
 import android.support.v7.view.ActionMode;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/amaze/filemanager/fragments/MainFragment.java
```diff
@@ -1594,18 +1597,30 @@ public String getCurrentPath() {
     private void addShortcut(LayoutElementParcelable path) {
         //Adding shortcut for MainActivity
         //on Home screen
-        Intent shortcutIntent = new Intent(getActivity().getApplicationContext(),
-                MainActivity.class);
+        final Context ctx = getActivity().getApplicationContext();
```
EmmanuelMess: Reason for not using getContext()?

EmmanuelMess:
State of license headers should be a separate issue.

app/src/main/java/com/amaze/filemanager/fragments/MainFragment.java
dmitrio95: In fact, the reason was my intent to change as little as possible. It seems this really does not matter. I will change it to `getContext()`.

VishalNehra:
Works great with Nova. Thanks for your contribution ^_^

Issue #913, opened 2018-01-09
SFTP/SMB file icons are used to display directories


TranceLove:
With current `master` branch, when listing files on a SMB or SFTP server, folders are displayed as files.

Tap into item can still list the files inside the directory; it's just the display problem.

![screenshot_1515483274](https://user-images.githubusercontent.com/251819/34709558-aac6ad0a-f552-11e7-865a-d98272d23f8d.png)

TranceLove:
With #918 merged I think this can be mark as closed.

PR #914, opened 2018-01-09
Force using folder icon if isDirectory is true

TranceLove:
Fix for #913.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/adapters/data/LayoutElementParcelable.java
+++ b/app/src/main/java/com/amaze/filemanager/adapters/data/LayoutElementParcelable.java
@@ -24,6 +24,7 @@ import android.os.Parcel;
 import android.os.Parcelable;
 import android.support.annotation.DrawableRes;
 
+import com.amaze.filemanager.R;
 import com.amaze.filemanager.filesystem.HybridFileParcelable;
 import com.amaze.filemanager.ui.icons.Icons;
 import com.amaze.filemanager.utils.OpenMode;
@@ -55,7 +56,7 @@ public class LayoutElementParcelable implements Parcelable {
                                    String symlink, String size, long longSize, boolean header,
                                    String date, boolean isDirectory, boolean useThumbs) {
         filetype = Icons.getTypeOfFile(new File(path));
-        @DrawableRes int fallbackIcon = Icons.loadMimeIcon(path);
+        @DrawableRes int fallbackIcon = (isDirectory) ? R.drawable.ic_grid_folder_new : Icons.loadMimeIcon(path);
 
         if(useThumbs) {
             if (filetype == Icons.IMAGE || filetype == Icons.VIDEO || filetype == Icons.APK) {


app/src/main/java/com/amaze/filemanager/adapters/data/LayoutElementParcelable.java
```diff
@@ -55,7 +56,7 @@ public LayoutElementParcelable(String title, String path, String permissions,
                                    String symlink, String size, long longSize, boolean header,
                                    String date, boolean isDirectory, boolean useThumbs) {
         filetype = Icons.getTypeOfFile(new File(path));
-        @DrawableRes int fallbackIcon = Icons.loadMimeIcon(path);
+        @DrawableRes int fallbackIcon = (isDirectory) ? R.drawable.ic_grid_folder_new : Icons.loadMimeIcon(path);
```
EmmanuelMess: Why not edit `Icons.loadMimeIcon()`?

TranceLove: `Icons.loadMimeIcon()` uses `java.io.File` to parse the path, alone it cannot detect SMB and SFTP paths as directories.

One approach I used was passing `isDirectory` into `Icons.loadMimeIcon()`, but this will also affect `CompressedExplorerAdapter`, but just pass in `CompressedObjectParcelable.directory` would be fine.

I'm good to go this way if it's better.

EmmanuelMess: Can't we generify the file parser (say, with HybridFile)?

TranceLove: `HybridFileParcelable` would be a better fit, as we will need `HybridFile.generateMode(context)` to parse the path for the correct `OpenMode` since we only provide the file's path.

Generify the file parser in `Icons.loadMimeIcon()` can be a better alternative as it encapsulates logic, but then `LayoutElementParcelable` (or its callers) may want to have access to `Context` or provide `HybridFileParcelable` instance, and may result in a new `LayoutElementParcelable` constructor values can be provided by `HybridFile` itself.

VishalNehra: Yes, it's not a good idea to instantiate ````HybridFileParcelable```` in ````LayoutElementParcelable````
As former one is the raw base used to form the other in later stages. We need to use what we already have in ````LayoutElementParcelable````. Instantiating ````File``` from path in ````Icons.loadMimeIcon```` ain't a good idea either, but we didn't (also, shouldn't) use any of it's method because of obvious reasons.
Given this, I think we shouldn't parcel icon data at all. Let adapter handle it all. This is creating lots of other inconsistency in icons, not just for directory icons.

EmmanuelMess: We don't need to instantiate a new ``HybridFileParcelable``, just ask for one, each time a ``LayoutElementParcelable`` is created I'm sure a ``HybridFileParcelable`` can be created or already exists.

> Given this, I think we shouldn't parcel icon data at all. Let adapter handle it all. This is creating lots of other inconsistency in icons, not just for directory icons.

Not putting these icons inside ``HybridFileParcelable`` means creating redundant instantiations down the line to compute the correct icons.

TranceLove: However, `CompressedExplorerAdapter`, the other user of `Icons.loadMimeIcon(path)`, doesn't seem to use `LayoutElementParcelable`.

VishalNehra: ````Not putting these icons inside HybridFileParcelable means creating redundant instantiations down the line to compute the correct icons.````
Loading icons inside adapter isn't redundant. Rewriting whole code based on type of file system again in ````Icons.loadMimeIcon```` or ````LayoutElementParcelable```` would be redundant.

````However, CompressedExplorerAdapter, the other user of Icons.loadMimeIcon(path), doesn't seem to use LayoutElementParcelable.````
CompressedExplorerAdapter only supports ````File```` filesystem. Hence it works fine with ````Icons.loadMimeIcon````

EmmanuelMess: > Loading icons inside adapter isn't redundant. Rewriting whole code based on type of file system again in Icons.loadMimeIcon or LayoutElementParcelable would be redundant.

It requires redundant reinstantiation of File objects.

EmmanuelMess: We are overthinking this, look at [commit Icons.loadMimeIcon(file) to loadMimeIcon(path, isDir)](https://github.com/EmmanuelMess/AmazeFileManager/commit/b6fe3fde4fd5703fb21a9eca9a2f17d30afbdd0f).

TranceLove: Good. Though the changeset was a bigger than I thought...

TranceLove:
Closing in favor of b6fe3fde by @EmmanuelMess.

EmmanuelMess:
Will PR tomorrow when I see the horrible mistakes I made, in the meantime you can rebase to my branch so that you have a copy of the commit.

Issue #905, opened 2018-01-05
SFTP file support limited to a file at a time


avuton:
Briefly: Thank you so much for the SFTP support, it helped remove the final external app blob I had left on my phone.

With that said, it's worth mentioning that I've been using it with a private key,  and I have noticed an odd bug:

Behaviour:
1. Go to SFTP directory, several directories in.
2. Copy directory which contains multiple files.
3. Paste directory to local storage.
4. Notification pops up notifying the correct size and number of files which are downloading.
5. Now, the part that is the issue: Only one file is downloaded, no other notification is given.

This is revision de2153d built using the installPlayDebug task.

avuton:
@TranceLove Perhaps you would like to be tagged with this?

avuton:
This might be related to #887.

TranceLove:
I suppose so, solving #887 should help.

My tests:

Device: Oneplus One (Bacon) with CandyRom (7.1.2)

Case 1: copying a file worth 346MB (which guarantees copy time > 5 mins)
Progress dialog disappeared after started copying. It was still copying (as seen in network indicator), but was seen a `DummyFile` first

Case 2: Copy two folders, one with 150 files and one with 55 files
Same, progress dialog disappeared after start copying. Refresh the file list and network indicator proved copy was still in progress.

VishalNehra:
@TranceLove I managed to pinpoint the problem. SFTP fails to get total size for directory. When being called ````FileUtils.getTotalBytes```` from ````CopyService```` and apparently other services too. When copying a file this isn't the case, and progress goes in smoothly. Can you please check and verify if ````HybridFile.folderSize```` returns correctly during copying?

TranceLove:
Result of `HybridFile.folderSize` for SFTP targets looks fine to me. I did made a little change to the codebase to delegate `HybridFile.folderSize()` to `HybridFile.folderSize(context)` just to be sure.

Server is running on Xenial, openssh version string `OpenSSH_7.2p2 Ubuntu-4ubuntu2.2, OpenSSL 1.0.2g  1 Mar 2016`.

See TranceLove/AmazeFileManager@df900b9d.

VishalNehra:
@TranceLove no success. Fodler size shows up 0 bytes in properties dialog too, and copy progress also fails to show progress because of this.
Are you able to get progress for both files and folders?

Also, this is the debug output
````
[...]

PR #916, opened 2018-01-11
Fix watcher progress

### LINKED ISSUE #785: Crash on copy
While copying a single file from internal to sd, crash occurred after line 102 of PrepareCopyTask.

```
com.amaze.filemanager E/AndroidRuntime: FATAL EXCEPTION: service_startup_watcher
                                                           Process: com.amaze.filemanager, PID: 6949
                                                           java.lang.ArrayIndexOutOfBoundsException: length=0; index=-1
                                                               at java.util.ArrayList.remove(ArrayList.java:405)
                                                               at com.amaze.filemanager.utils.ServiceWatcherUtil$2.run(ServiceWatcherUtil.java:173)
                                                               at android.os.Handler.handleCallback(Handler.java:746)
```

### LINKED ISSUE #874: If file is copied in sd card, copy complete appears prematurely
If file is copied in sd card, copy complete appears prematurely and the folder reloads after the file is copied. The notification goes from '0B copied' to 'Copy complete'; if folder is reloaded manually the file will be half copied (a file that should be 80 MB weights less).

### LINKED ISSUE #887: SMB/Samba issues when copying
When I try to copy large files or large numbers of file from my desktop to my mobile phone over SMB using Amaze, it seems to fail (without clearly indicating that it's failed). Copying small single files seems to work.  I am able to copy the same files over SMB using Ghost Commander, so it doesn't seem to be an issue with my SMB network.

VishalNehra:
Notification won't show 'Copy completed' before actually completing it. In case there's no progress, the notification spinner will go indeterminate before waiting 5 secs for progress to resume. It'll resume from there when there's a progress.
Should also fix the race condition for starting the first service instance which some people faced earlier.

Fixes #785?
Fixes #874
Fixes #887


VishalNehra:
I manually removed it in 4b6a11b (last one)

app/src/main/java/com/amaze/filemanager/asynchronous/services/CopyService.java
```diff
@@ -85,7 +85,9 @@
 
     private long totalSize = 0L;
     private int totalSourceFiles = 0;
+    private String notificationID;
```
EmmanuelMess: Why String if you are going to parse it to int every time?

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
```diff
@@ -192,4 +213,24 @@ public void run() {
 
         handler.postDelayed(runnable, 0);
     }
+
+    public interface ServiceWatcherInteractionInterface {
+
+        enum STATE {
```
EmmanuelMess: Use final int vars.

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
```diff
@@ -192,4 +213,24 @@ public void run() {
 
         handler.postDelayed(runnable, 0);
     }
+
+    public interface ServiceWatcherInteractionInterface {
```
EmmanuelMess: It seems to me that each time this interface is used almost the same code is inside the functions.

EmmanuelMess:
@VishalNehra You can dismiss [this](https://github.com/TeamAmaze/AmazeFileManager/pull/916#pullrequestreview-88201630) then.

app/src/main/java/com/amaze/filemanager/asynchronous/services/CopyService.java
VishalNehra: Too much time consuming to rewrite everything. Will do later.

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
VishalNehra: No, enum s are cool !!

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
VishalNehra: Depends on the context

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
EmmanuelMess: They have [a significant overhead](https://android.jlelse.eu/android-performance-avoid-using-enum-on-android-326be0794dc3), not needed here.

EmmanuelMess:
@VishalNehra You dismissed all reviews, not just the one...

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
EmmanuelMess: Abstract class or stating functions then?

app/src/main/java/com/amaze/filemanager/asynchronous/services/CopyService.java
EmmanuelMess: #876 has cleaned most of it, will the string through...

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
VishalNehra: Abstract class it is, added new commits which reflect the changes.

avuton:
While this PR has [regressions](https://github.com/TeamAmaze/AmazeFileManager/issues/905#issuecomment-357181386), this does appear to fix #905 as well.

EmmanuelMess:
I made a mistake and now you have two merge commits, sorry (trying to fix).

VishalNehra:
Don't pull yet. It seems I messed up in last commit when trying to change notification id from string to int. Which causes two duplicate notifications to be generated. Will fix it soon.

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
EmmanuelMess: Also, the superclass implements the original interface, why?

EmmanuelMess: And the superclass is technically a service, it should not be in utils, it should be under asynchronous.services.

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
```diff
@@ -192,4 +213,24 @@ public void run() {
...
+         * Progress has been halted for some reason
+         */
+        void progressHalted();
+
+        /**
+         * Future extension for possible implementation of pause/resume of services
+         */
+        void progressResumed();
+
+        <T extends Service> T getServiceType();
```
EmmanuelMess: This method is redundant because of type erasure, just use the instance you are calling the method from (unless you can't access the Service instances, then you can have something like `Service getAttachedService();`).

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherProgressAbstract.java
```diff
@@ -33,4 +50,130 @@ public void progressResumed() {
...
+    /**
+     * Returns the {@link #dataPackages} list which contains
+     * data to be transferred to {@link ProcessViewerFragment}
+     * Method call is synchronized so as to avoid modifying the list
+     * by {@link com.amaze.filemanager.utils.ServiceWatcherUtil#handlerThread} while {@link MainActivity#runOnUiThread(Runnable)}
+     * is executing the callbacks in {@link ProcessViewerFragment}
+     *
+     * @return
+     */
+    public synchronized CopyDataParcelable getDataPackage(int index) {
```
EmmanuelMess: Why the object synchronization?

VishalNehra: Because this call is not atomic

EmmanuelMess: Why is it not atomic and why is it not synchronized on the list object?

VishalNehra: Please test and see if there are any problems.

VishalNehra: Currently this seems to be working fine and also serves the fix for original problem... So can be merged.. the only problem is sftp this PR wasn't focused to fix at the time.. 
Raise an issue later if you find any problem... We need to start pushing betas soon.

EmmanuelMess: > Currently this seems to be working fine and also serves the fix for original problem... So can be merged.. 

Good code isn't just fixing the existent issues, it is also fixing new issues that you haven't found yet.

Fix [this](https://github.com/TeamAmaze/AmazeFileManager/pull/916#discussion_r161290245), [this](https://github.com/TeamAmaze/AmazeFileManager/pull/916#discussion_r161062222), [this](https://github.com/TeamAmaze/AmazeFileManager/pull/916#discussion_r161287976), [this](https://github.com/TeamAmaze/AmazeFileManager/pull/916#discussion_r161288241) and [this](https://github.com/TeamAmaze/AmazeFileManager/pull/916#discussion_r161288695) (half of those are questions) with well written, efficient and non repetitive code and I'll merge.

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
VishalNehra: >You should use a constructor to set the superclass's fields.

Constructor would be the first thing called while Service would be instantiated. How would I be able to set them at that time.

>Also, the superclass implements the original interface, why?

 To avoid writing same code in all services

>And the superclass is technically a service, it should not be in utils, it should be under asynchronous.services.

Done

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
VishalNehra: Done

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
EmmanuelMess: > To avoid writing same code in all services

How does that work?

EmmanuelMess: > Constructor would be the first thing called while Service would be instantiated. How would I be able to set them at that time.

Ok, my mistake, you shouldn't set the variables in the constructor, but somewhere... It is too easy to leave variables initialized, maybe override `onStartCommand()`?

app/src/main/java/com/amaze/filemanager/asynchronous/services/ServiceWatcherProgressAbstract.java
```diff
@@ -31,9 +32,25 @@
...
+    public void onCreate() {
+        super.onCreate();
+
+        initVariables();
+    }
+
+    @Override
+    public int onStartCommand(Intent intent, int flags, int startId) {
+
+        initVariables();
```
EmmanuelMess: Twice for good measure?

VishalNehra: Ahh, my bad. Forgot to remove one in onCreate

TranceLove:
May I ask this PR be completed and included by the 3.3.0 milestone? I think this will make the user experience much more better.

EmmanuelMess:
This fixes issues that halt release, so yes.

VishalNehra:
Already working on a separate branch for this.

TranceLove:
This should be closed in favour of #1108.

PR #917, opened 2018-01-11
Fix save as

### LINKED ISSUE #867: If app is paused 'Save as...' doesn't work
If app is paused 'Save as...' intent will not pass through onCreate(), resulting in the FAB not changing.

### LINKED ISSUE #868: File saving can never succeed
This can never return true:

https://github.com/TeamAmaze/AmazeFileManager/blob/3f6c063942579d983a382e0c010fdc0be351df60/app/src/play/java/com/amaze/filemanager/activities/MainActivity.java#L594

`FileUtil.writeUriToStorage(..., getCurrentMainFragment().getCurrentPath()); ` 
checks:
`FileOutputStream output = new FileOutputStream(file, true);` 
on a folder resulting in the exception:
`java.io.FileNotFoundException: [the current directory]: open failed: EISDIR (Is a directory)`
that is catched and returns false:
https://github.com/TeamAmaze/AmazeFileManager/blob/3f6c063942579d983a382e0c010fdc0be351df60/app/src/main/java/com/amaze/filemanager/filesystem/FileUtil.java#L643-L656

VishalNehra:
Fixes saving file from external source
Fixes #867, fixes #868

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/filesystem/FileUtil.java
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/FileUtil.java
@@ -204,7 +204,7 @@ public abstract class FileUtil {
                         switch (hFile.getMode()) {
                             case FILE:
                             case ROOT:
-                                if (!FileUtil.isWritable(new File(currentPath))) {
+                                if (!FileUtil.isWritable(new File(finalFilePath))) {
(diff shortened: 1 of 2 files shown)


EmmanuelMess:
Trying to save an image the name is 'null', also, after saving the app throws me to the launcher, not the last app (Firefox app). These don't seem as bugs you caused with this PR, I'll make them into issues after pulling this.

PR #919, opened 2018-01-11
Fix folderSize() for SFTP targets

TranceLove:
Related to #905.

Changes:
- Added FileUtils.folderSizeSftp() if getting size using du failed
- Force use of HybridFile.folderSize(context)

The implementation is a bit different than that mentioned in #905, with the attempt to restrict all operations (original/fallback) within one `SSHClient` session.


TranceLove:
Seems to have broken copy function at all. Closing, until it is fixed.

💔

VishalNehra:
It was working fine for me though.

PR #924, opened 2018-01-13
Add FileUtils.folderSizeSftp() for recursive folder size calculation

TranceLove:
A redo of #919, related to #905. Use this to calculate folder size for SFTP targets if `du` command fails.

`SshClientUtils.execute(SSHClientTemplate)` also changed to capture all Exceptions, including RuntimeExceptions, which caused app to hang when copying files/directories over SFTP due to `NumberFormatException` thrown from `HybridFile.folderSize(Context)` when `du` execution failed or is not available at all.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/filesystem/HybridFile.java
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/HybridFile.java
@@ -592,29 +592,10 @@ public class HybridFile {
 
         switch (mode){
             case SFTP:
-                return SshClientUtils.execute(new SshClientTemplate(path, false) {
+                return SshClientUtils.execute(new SFtpClientTemplate(path) {
                     @Override
-                    public Long execute(SSHClient client) throws IOException {
-                        Session session = client.startSession();
-                        String remotePath = SshClientUtils.extractRemotePathFrom(path);
-                        Session.Command cmd = session.exec(String.format("du -b -s \"%s\"", remotePath));
-
-                        String result = new String(IOUtils.readFully(cmd.getInputStream()).toByteArray());
-                        cmd.close();
-                        session.close();
-                        result = result.substring(0, result.indexOf('/') - 1).trim();
-
-                        // It is possible du may still return valid value even getting permission problems,
-                        // so we check if result is a valid number.
-                        try {
-                            return Long.parseLong(result);
-                        } catch (NumberFormatException whenParseFailed){
-                            Log.w(TAG, "Error running du, fallback to FileUtils.folderSizeSftp()");
-                            SFTPClient sftpClient = client.newSFTPClient();
-                            Long retval = FileUtils.folderSizeSftp(sftpClient, remotePath);
-                            sftpClient.close();
-                            return retval;
-                        }
+                    public Long execute(SFTPClient client) throws IOException {
+                        return client.size(SshClientUtils.extractRemotePathFrom(path));
                     }
                 });
             case SMB:
--- a/app/src/main/java/com/amaze/filemanager/filesystem/HybridFile.java
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/HybridFile.java
@@ -558,14 +558,7 @@ public class HybridFile {
 
         switch (mode) {
             case SFTP:
-                return SshClientUtils.execute(new SFtpClientTemplate(path, false) {
-                    @Override
-                    public Long execute(SFTPClient client) throws IOException {
-                        Long retval = client.size(SshClientUtils.extractRemotePathFrom(path));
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/amaze/filemanager/filesystem/HybridFile.java
```diff
@@ -599,20 +592,28 @@ public long folderSize(Context context) {
...
-                return SshClientUtils.execute(new SshClientSessionTemplate(path) {
+                return SshClientUtils.execute(new SshClientTemplate(path, false) {
                     @Override
-                    public Long execute(Session session) throws IOException {
-                        Session.Command cmd = session.exec(String.format("du -b -s \"%s\"",
-                                SshClientUtils.extractRemotePathFrom(path)));
+                    public Long execute(SSHClient client) throws IOException {
+                        Session session = client.startSession();
+                        String remotePath = SshClientUtils.extractRemotePathFrom(path);
+                        Session.Command cmd = session.exec(String.format("du -b -s \"%s\"", remotePath));
```
EmmanuelMess: Isn't there a method that does the same thing as the command?

TranceLove: `net.schmizz.sshj.sftp.SFTPClient.size(path)`?
That seems to return predictable values like 4096, 20480 (or similar) for folders.

EmmanuelMess: Bear in mind that we are trying not to use commands because the responses are oriented towards users and can be unpredictable (in cases of error). But I don't know if this is true for SFTP.

TranceLove: True indeed - hence #905 and this PR. Just not sure if it's fine by simply return the reported value by `SFTPClient.size(path)`.

Let's test the alternate method back at #905 and see how it goes.

TranceLove: True indeed.

9ea93ed will use SFTPClient.size(path) alone. Copy is still functioning properly, but expect copy progress will report incorrect size.

You may want to refactor recursive copy's code so that each folder it descends to will update the total copy size to keep things beautiful.

screenshot_1516030505

VishalNehra: @TranceLove please test with #916 
I think with latest changes made, should fix all the problems with copy.

EmmanuelMess: Except the erroneous copy size.

TranceLove: @VishalNehra yes, screenshot above is test with #916.

Alexander--: @TranceLove @EmmanuelMess 

> `net.schmizz.sshj.sftp.SFTPClient.size(path)`?
> That seems to return predictable values like 4096, 20480 (or similar) for folders.

This is the size of *folder itself* — number of bytes, taken by *filenames* of it's direct descendants. I don't think, that there is any popular filesystem (on Linux or Windows), that can track total size of directory descendants. 

> Except the erroneous copy size. Maybe say: "x MB of unknown".

I second this. Trying to recursively compute directory size over network protocol have potential to take more time than copy itself (imagine copying deeply nested directory tree with thousands of individually small files — making full network roundtrip for each `stat()`).

Alternatively, if the remote directory is writable, Amaze can try to upload small script/executable (for example PowerShell program), to perform the job of `du`.

TranceLove: > Alternatively, if the remote directory is writable, Amaze can try to upload small script/executable (for example PowerShell program), to perform the job of `du`.

Sorry, but allowing the app to inject programs/scripts to execute on remote servers is **always a bad idea**.

Alexander--: @TranceLove you are already running `du` on remote server, aren't you?

Anyway, there is no need to create an executable — simply passing a script via PowerShell's `-Command` argument should do.

TranceLove: > @TranceLove you are already running `du` on remote server, aren't you?

No denying. So I understand why I need to change this - though things maybe a bit safe in the UNIX world (relatively, compared to Windows).

Judging this, it is possible that the SFTP parts may no longer be able to report remaining spaces, used spaces... correctly because of the remote commands involved. :/

PR #933, opened 2018-01-18
Use statvfs@openssh.com extension to query disk size/available space for SFTP targets

TranceLove:
Concern of running commands on the remote server, whether it's command already available on the remote server or injecting scripts, in #924, which is dangerous.

In previous implementation of SFTP support, remote command was used to query disk size and free space on the remote SSH server.

Luckily, there is a `statvfs@openssh.com` extension available at OpenSSH and Bitvise SSH servers, the two most common SSH server implementations, allowing SFTP clients to query disk space and free space without running arbitary commands.

As suggested [here](https://serverfault.com/questions/301953/how-to-check-diskspace-on-sftp-server) and [here](https://stackoverflow.com/questions/19769776/sftp-check-free-space-available#19775455). Specification is [here](https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/ssh/PROTOCOL?annotate=HEAD), at section 3.4.

This commit implements such support, and yes, it's tested with OpenSSH and Bitvise SSH servers. On SSH server implementations that doesn't support `statvfs@openssh.com` extension (which would give incorrect replies), `HybridFile.getTotal(Context)` and `HybridFile.getUsableSpace()` will return 0 (not tested though).

`BigInteger` was used in the first implementation, however changed back to `Long` in align with `HybridFile`...and actually I don't know if there is any implication on using `BigInteger`. Based on calculation, the caveat on using `Long` is app may crash if disk size on the SSH server is above 8191 petabytes (`Long.MAX_VALUE`). `BigInteger` on the other hand can support up to 16383 petabytes.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/filesystem/ssh/Statvfs.java
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/ssh/Statvfs.java
@@ -95,10 +95,10 @@ public class Statvfs
          */
 
         // f_bsize
-        public final Long fileSystemBlockSize;
+        public final Integer fileSystemBlockSize;
 
         // f_frsize
-        public final Long fundermentalFileSystemBlockSize;
+        public final Integer fundamentalFileSystemBlockSize;
 
         // f_blocks
         public final Long fileSystemBlocks;
@@ -122,10 +122,10 @@ public class Statvfs
         public final BigInteger fileSystemId;
 
         // f_flag
-        public final Long fileSystemFlag;
+        public final Integer fileSystemFlag;
 
         // f_namemax
-        public final Long filenameMaxLength;
+        public final Integer filenameMaxLength;
 
         public Response(net.schmizz.sshj.sftp.Response response) throws SFTPException, Buffer.BufferException
         {
@@ -137,8 +137,8 @@ public class Statvfs
 
             mResponse = response;
 
-            fileSystemBlockSize = mResponse.readUInt32();
-            fundermentalFileSystemBlockSize = mResponse.readUInt64();
+            fileSystemBlockSize = Long.valueOf(mResponse.readUInt32()).intValue();
+            fundamentalFileSystemBlockSize = Long.valueOf(mResponse.readUInt64()).intValue();
             fileSystemBlocks = mResponse.readUInt64();
             freeFileSystemBlocks = mResponse.readUInt64();
             availableFileSystemBlocks = mResponse.readUInt64();
@@ -146,8 +146,8 @@ public class Statvfs
             freeFileInodes = mResponse.readUInt64();
             availableFileInodes = mResponse.readUInt64();
             fileSystemId = mResponse.readUInt64AsBigInteger();
-            fileSystemFlag = mResponse.readUInt64();
-            filenameMaxLength = mResponse.readUInt64();
+            fileSystemFlag = Long.valueOf(mResponse.readUInt64()).intValue();
+            filenameMaxLength = Long.valueOf(mResponse.readUInt64()).intValue();
         }
 
         /**
--- a/app/src/main/java/com/amaze/filemanager/filesystem/ssh/Statvfs.java
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/ssh/Statvfs.java
@@ -1,0 +1,179 @@
+/*
+ * Statvfs.java
+ *
+ * Copyright © 2018 Raymond Lai <airwave209gt at gmail.com>.
+ *
+ * This file is part of AmazeFileManager.
+ *
+ * AmazeFileManager is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * AmazeFileManager is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/amaze/filemanager/filesystem/ssh/Statvfs.java
```diff
@@ -0,0 +1,179 @@
...
+        public final net.schmizz.sshj.sftp.Response mResponse;
+
+        /*
+            Regarding choice of number types:
+
+            - fileSystemFlag is only 0x0, 0x1 and 0x2
+            - filenameMaxLength should be well within Long.VALUE_MAX
+
+            Others, can use BigInteger to prepare for the ever increasing disk size for years to come.
+         */
```
Alexander--: Actually several of those container types are hilariously oversized.

Maximum length of filename (a component of file path between two slashes) is directly tied to size of the memory buffer, used for listing files in directory. I strongly suspect, that we won't *need* filenames longer than 2Gb for couple hundred years at least. Even if we somehow start using such names, an Android file manager won't be able to allocate an array for storing those names, because maximum length of Java array is slightly below `Integer.MAX_VALUE`.

File system block size is another example — if file systems with such ridiculously huge blocks become commonplace, knowing the size of block will be useless, because using temporary buffers of such sizes is going to be impractical. May as well cap them to 2Gb.

By the way, it is not necessary to use `BigInteger` for storing unsigned 64-bit values — `long` can store them fine. If necessary, you can allocate an instance of `BigInteger` lazily before actually operating on it.

TranceLove: Thanks for the suggestion.

ffca0bb changes them. Also changed `fileSystemFlag` to `Integer` as valid values are actually 0, 1, 2 (per specification) or 3 even if `|` flag is applied.

However, OpenSSH on Linux (at least on my Xenial machines) still requires `BigInteger`. Sorry about this.

Alexander--: @TranceLove It is largely inconsequential (really, how many time would that method be called during a copy? one?), but let's clarify that point: since you are calling `readUInt64AsBigInteger`, it means, that the size of value should be the same as `long` — 64 bits. The relevant methods in jssh seem to read exactly same amount of bytes from socket: [1][1], [2][2]. So there can't be any objective reason, requiring use of BigInteger.

The only difference between long and BigInteger is treatment of sign bit: later preserves original unsigned value, former may "overflow" into negative for sufficiently high filesystem IDs. But you will only notice the difference if you *print the data to user* or *perform math on it*. There is nothing wrong with simply *storing* unsigned integer in signed Java data type and [converting to BigInteger before printing it to user](https://stackoverflow.com/questions/7031198).

Btw, based on the style of your code, it looks like you have some experience programming for .Net or other "big" managed platforms. In most of those platforms using object-like data types for primitives is preferable (or the only) choice. In Java primitive scalar types (`long`, `int`, 'char') are more widely used for that purpose. There is zero benefit to using `Long.valueOf` and taking `intValue` from it the way you do. Just cast to int:

````java
filenameMaxLength = (int) mResponse.readUInt64();
````

[1]: https://github.com/hierynomus/sshj/blob/8ca6451d5dee2667d1a5/src/main/java/net/schmizz/sshj/common/Buffer.java#L351
[2]: https://github.com/hierynomus/sshj/blob/8ca6451d5dee2667d1a5/src/main/java/net/schmizz/sshj/common/Buffer.java#L343

Alexander--: @TranceLove Of course the check behind your exception is meaningful — as I explained above, it might be bad idea to expose an unsigned value stored in `long`, because it may overflow when you print it out or operate on it (as it did in your example). But the check is mainly there to warn you about dangers of overflow. If you are already conscious of that, you can make that part of contract.

Personally I would do it like this:

1. Make `fileSystemId` long
2. Make `fileSystemId` private
3. Create two getters: `BigInteger getFsId()` and `long getRawFsId()`
4. `toString()` shall print result of the first getter

This will avoid creating and storing instance of `BigInteger` unless somebody actually uses the field.

Old versions of Java used to have no good utility methods for working with unsigned values. But since Java 8, there are multiple static methods in `Long` class (such as [toUnsignedString](https://docs.oracle.com/javase/8/docs/api/java/lang/Long.html#toUnsignedString-long-)), that allow you to treat scalar `long` values as unsigned. If you need to work with 64-bit unsigned numbers, BigInteger is not a must anymore.

As for the silly check, here is relevant source code:

````java
public long readUInt64()
            throws BufferException {
        long uint64 = (readUInt32() << 32) + (readUInt32() & 0xffffffffL);
        if (uint64 < 0) // this line has tripped you
            throw new BufferException("Cannot handle values > Long.MAX_VALUE");
        return uint64;
}
````

You can just read the value yourself without checking for negativity:

````java
public long readUInt64FromBuffer(net.schmizz.sshj.common.Buffer buffer) {
        long uint64 = (buffer.readUInt32() << 32) + (buffer.readUInt32() & 0xffffffffL);
        return uint64;
}
````

TranceLove: Excellent. Never looked that deep into sshj's source code - thanks for pointing out!

PR #938, opened 2018-01-21
Moved drawer open/close logic to post event

### LINKED ISSUE #692: [known bug] On tablets, if rotated twice the drawer opens and doesn't close
If you rotate to landscape the drawer opens, but after rotating back to portrait, the drawer doesn't close.

PiwwowPants:
Fixes #692

Another low priority bug fix to get comfortable with the code base a little more.

The close/openDrawer calls in the MainActivity's intialiseView method were being performed before the drawer finished laying out. This meant when closeDrawer got called (exiting a "tablet" layout), openDrawer got called after in the DrawerLayout.onRestoreInstanceState method.
Moving the calls to open/close out to run when the view posts resolved this. The drawer will now always be closed when rotating out of "tablet" orientation.

Also removed redundant conditionals on the if-else. If it's not not null, it's null. If that is somehow wrong here, let me know.

I tested:
- Rotating into tablet orientation (on my Pixel 2 XL, landscape is tablet) opens the drawer like before and forces it to stay open
- Rotating into non-tablet orientation, closes the drawer
- The drawer can still be opened and is usable like before
- There is no animation jitter between the view post and the action

### DIFF
--- a/app/src/play/java/com/amaze/filemanager/activities/MainActivity.java
+++ b/app/src/play/java/com/amaze/filemanager/activities/MainActivity.java
@@ -1754,8 +1754,7 @@ public class MainActivity extends ThemedActivity implements OnRequestPermissions
             isDrawerLocked = false;
             initDrawerOpen = false;
         }
-        mDrawerLayout.post(() ->
-        {
+        mDrawerLayout.post(() -> {
             if (initDrawerOpen) {
                 mDrawerLayout.openDrawer(mDrawerLinear);
             } else {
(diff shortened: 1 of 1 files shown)


PiwwowPants:
Had to consume your CI machine again to fix a formatting issue with the curly brace. Didn't mean to put it on a newline on the post call and mess up that funky inline flow you have going.

TranceLove:
Looks good to me, and the fix did what's described (as tested with Lollipop tablet emulator). I'd vote a +1 (after code review).

app/src/play/java/com/amaze/filemanager/activities/MainActivity.java
```diff
@@ -1743,17 +1743,24 @@ void initialiseViews() {
...
-        } else if (findViewById(R.id.tab_frame) == null) {
-
+            initDrawerOpen = true;
+        } else {
             mDrawerLayout.setDrawerLockMode(DrawerLayout.LOCK_MODE_UNLOCKED, mDrawerLinear);
-            mDrawerLayout.closeDrawer(mDrawerLinear);
             isDrawerLocked = false;
+            initDrawerOpen = false;
         }
+        mDrawerLayout.post(() -> {
```
EmmanuelMess: Maybe split the function in two, so as to remove the `initDrawerOpen` variable and reduce code length?

VishalNehra:
Nice catch, this was quite an old bug and I failed to find the root cause. Can be merged.

PR #939, opened 2018-01-21
Prevent the PreferencesActivity from being restarted when irrelevant preferences are changed

PiwwowPants:
I do not believe there is an issue linked to this fix. Let me know how you feel about my handling of this.

Bug fixed: When changing a preference value in FolderPrefs or QuickAccessPrefs, a double back press is required to exit the fragment.

- Currently, every time a preference fragment makes a change, it sets the "changed" Boolean in PreferencesActivity. onBackPressed, if this Boolean is true, the activity ends up getting restarted. On restart, it will attempt to open the same fragment again. This behavior is to handle the multiple screens of ColorPref.
- While the logic of telling the activity that the fragment changed a preference makes sense, it does not seem to have any reason to care that preference changed, it just wants to know if it needs to restart.

This change refactors the "changed" logic to convert it into "restartActivity" logic. This new concept tells the activity when it needs to restart rather than when a preference changes. In the future, if a single screen needs to handle restarting the activity, this same problem will come up unless onBackPressed is overridden appropriately for the fragment.

I tested:
- Closing QuickAccessPref fragment (closes to activity)
    - Preferences still set appropriately immediately
- Closing FolderPref fragment (closes to activity)
    - Preferences still set appropriately immediately
- Closing ColorPref from the first screen (closes to activity)
- Closing ColorPref from second screen (closes to first ColorPref screen)
    - Activity is restarted appropriately with any color changes

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/activities/PreferencesActivity.java
+++ b/app/src/main/java/com/amaze/filemanager/activities/PreferencesActivity.java
@@ -62,7 +62,7 @@ public class PreferencesActivity extends ThemedActivity {
     public static final int QUICKACCESS_PREFERENCE = 3;
     public static final int ADVANCEDSEARCH_PREFERENCE = 4;
 
-    private boolean changed = false;
+    private boolean restartActivity = false;
     //The preference fragment currently selected
     private int selectedItem = 0;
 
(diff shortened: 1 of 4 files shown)


EmmanuelMess:
Will test tomorrow, we needed this, thanks for your contributions!

TranceLove:
Looks good to me, and the fix did what's described (as tested with Lollipop emulator). I'd vote a +1.

app/src/main/java/com/amaze/filemanager/activities/PreferencesActivity.java
```diff
@@ -144,12 +144,12 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
 
-    public void setChanged() {
-        changed = true;
+    public void setRestartActivity() {
+        restartActivity = true;
     }
 
-    public boolean getChanged() {
-        return changed;
+    public boolean getRestartActivity() {
```
TranceLove: Not sure if others would agree... but to my knowledge for boolean getters, it would be better to say `isRestartActivity()` here.

PR #940, opened 2018-01-21
Added accent colors to available colors

PiwwowPants:
Not related to an existing issue:

Bugfix: After setting color preferences explicitly through settings (notably to "Default), killing the app and attempting to reopen causes the app to crash on startup due to colors set in theme missing from available colors array

I'm not sure how this was missed, so let me know if I am missing something. Each theme contains an "accent color" (R.color.accent*) but that accent color is not present in the ColorPreference.availableColors array. This causes an ArrayOutOfBounds exception to be thrown after the app is killed and get recreated completely, since the initial color setter requires the color to be in that array.

Fix: Added all accent colors to available colors array

I tested:
- All color preferences to verify they started appropriately
     - The color themes are unmodified
- A crashing exception is not thrown on start up

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/utils/color/ColorPreference.java
+++ b/app/src/main/java/com/amaze/filemanager/utils/color/ColorPreference.java
@@ -41,7 +41,11 @@ public class ColorPreference {
             R.color.primary_brown,
             R.color.primary_grey_900,
             R.color.primary_blue_grey,
-            R.color.primary_teal_900
+            R.color.primary_teal_900,
+            R.color.accent_pink,
+            R.color.accent_amber,
+            R.color.accent_light_blue,
+            R.color.accent_light_green
     );
 
     private static final String TAG = "ColorPreference";
--- a/app/src/main/java/com/amaze/filemanager/utils/color/ColorPreference.java
+++ b/app/src/main/java/com/amaze/filemanager/utils/color/ColorPreference.java
@@ -6,11 +6,13 @@ import android.content.res.Resources;
 import android.graphics.drawable.ColorDrawable;
 import android.support.annotation.ColorInt;
 import android.support.annotation.ColorRes;
+import android.support.v4.content.ContextCompat;
 import android.util.Log;
 
 import com.amaze.filemanager.R;
 import com.amaze.filemanager.utils.Utils;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


TranceLove:
I'm able to reproduce the problem you mentioned, and the code did fix it.

However, is it possible to filter out the "duplicated" colors in the color picker dialog? We know they have actually different color codes, but in the UI it's not, and maybe confusing to end users.

Before applying the fix
screenshot_1516586977

After applying, 
screenshot_1516587425

PiwwowPants:
I do not have an elegant answer right now because it is late, but I will look into it tomorrow.

My kneejerk guess is to convert the resource values to hex codes, run that through a LinkedHashSet, then convert back to resource values as an array without duplicate colors. The problem is the hex strings won't match hashes even if the values match without overriding equals or hashCode on Set.

app/src/main/java/com/amaze/filemanager/utils/color/ColorPreference.java
```diff
@@ -214,5 +220,18 @@ public ColorDrawable getDrawable(ColorUsage usage) {
     public String getColorAsString(ColorUsage usage) {
         return String.format("#%06X", (0xFFFFFF & getColor(usage)));
     }
+    
+    public static List<Integer> getUniqueAvailableColors(Context context) {
```
EmmanuelMess: O(n) for getting the color for a ui? isn't it easier to have two arrays?

TranceLove:
e7012ab worked on my Oneplus One 7.1.2, but not working on Lollipop emulator.

```
java.lang.NoSuchMethodError: No virtual method getContext()Landroid/content/Context; in class Lcom/amaze/filemanager/fragments/preference_fragments/ColorPref; or its super classes (declaration of 'com.amaze.filemanager.fragments.preference_fragments.ColorPref' appears in /data/app/com.amaze.filemanager-1/split_lib_slice_1_apk.apk)
 at com.amaze.filemanager.fragments.preference_fragments.ColorPref.onPreferenceClick(ColorPref.java:119)
 at android.preference.Preference.performClick(Preference.java:985)
 at android.preference.PreferenceScreen.onItemClick(PreferenceScreen.java:214)
 at android.widget.AdapterView.performItemClick(AdapterView.java:305)
 at android.widget.AbsListView.performItemClick(AbsListView.java:1146)
 at android.widget.AbsListView$PerformClick.run(AbsListView.java:3053)
 at android.widget.AbsListView$3.run(AbsListView.java:3860)
 at android.os.Handler.handleCallback(Handler.java:739)
 at android.os.Looper.loop(Looper.java:135)
 at android.app.ActivityThread.main(ActivityThread.java:5254)
 at java.lang.reflect.Method.invoke(Method.java:372)
 at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:903)
 at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:698)
```

TranceLove:
I had verified the changeset again, with both My Oneplus One (7.1.2) and Lollipop emulator, both acted fine without crash. Adding ae65d5a fixing compatibility with Lollipop, I think it's safe to let this in now.

PR #941, opened 2018-01-21
Modified disableButton logic for the shortcuts dialog to watch both edit texts for validity

PiwwowPants:
Not tied to a current issue.

Bug: If a newly created or modified shortcut has either a title OR a valid file path, it can be created/modified, allowing invalid shortcuts to be created

Previously, shortcut validity was being checked on the title edit text and the path edit text separately. This change consolidates the check into one method with two listeners that both check for validity of changed text and the other edit text's value

I kept the enable logic in the listener to allow it to be more easily extended later and only broke the "isValid" check out to its own method.

I tested:
- Modifying either edit text functions as expected
- A shortcut without both valid values cannot be created/modified
- A shortcut with bot valid values can be created/modified


EmmanuelMess:
I think this just adds lines... If in the future someone needs to add more checks they can add them to the corresponding listener.

EmmanuelMess:
Nah, just leave it as is, that would make the code more difficult to understand at first glance.

PR #961, opened 2018-01-24
Fix encryption input problems

### LINKED ISSUE #950: Password Entry is Hidden on Encryption but Visible on Decryption
#### Expected Behavior
Password entry is Hidden for both the Encryption and Decryption process.

#### Actual Behavior
Password entry is Hidden for the Encryption process but Visible for the Decryption process.

#### How to Reproduce
1. Launch Amaze File Manager 3.2.1 on Android 8.1.0
2. Click the 3 Vertical Dots icon to the right of a File
3. Select 'Encrypt'
4. Select 'Got It'
5. Enter the same password into the top and bottom fields
6. Select 'OK'
7. Select the newly encrypted file
8. Type the password from Step 5

---
    Device: Pixel 2 XL
    OS: Android 8.1.0
    App Version: 3.2.1
---

#### Recording of the Bug
inconsistentpasswordhiding

<br /><hr/><em>Posted on <a href="https://utopian.io/utopian-io/@thewizard.studio/password-entry-is-hidden-on-encryption-but-visible-on-decryption">Utopian.io -  Rewarding Open Source Contributors</a></em><hr/>

### LINKED ISSUE #951: File Encryption Form is Submittable w/ Blank Password Fields
#### Expected Behavior
Submitting blank password fields when Encrypting a File results in an error, or the form cannot be submitted until entries exist in both fields.

#### Actual Behavior
Submitting blank password fields when Encrypting a File causes the dialog window to close without creating an encrypted file, and with no error message explaining what just happened.

#### How to Reproduce
1. Launch Amaze File Manager 3.2.1 on Android 8.1.0
2. Click the 3 Vertical Dots icon to the right of a File
3. Select 'Encrypt'
4. Select 'Got It'
6. Select 'OK'

---
    Device: Pixel 2 XL
    OS: Android 8.1.0
    App Version: 3.2.1
---

#### Recording of the Bug

<br /><hr/><em>Posted on <a href="https://utopian.io/utopian-io/@thewizard.studio/file-encryption-form-is-submittable-w-blank-password-fields">Utopian.io -  Rewarding Open Source Contributors</a></em><hr/>

VishalNehra:
Fixes #950 
Fixes #951


TranceLove:
Did fixed #950 and #951 as advertised. I'd vote +1.

EmmanuelMess:
I rebased for clarity.

VishalNehra:
It's not necessary at all. As many apps I've seen, this is the expected behavior.

EmmanuelMess:
I mean it is nicer to know that "OK" isn't available because of some error, before you press it. Check WiFi in Android Settings, when inputing a password for a WPA2 net it greys out the "OK" button until the pass has enough characters.

VishalNehra:
There's a difference. In WiFi settings the ok button disappear because of password length.. we have no such constraint on password. The only one is that repeat password should match.. I rarely find any UX where repeat password is matched against original one as the user types.

EmmanuelMess:
As per the [Material Guidelines: Text Fields - Text field content](https://material.io/guidelines/components/text-fields.html#text-fields-states):

> If a text field is invalid, an error message is displayed when the user shifts focus to another element. To indicate an error, the input line and label use a color with a warmer hue, such as red or orange. 

In this case the focus shift would be pressing "ok" button. I still think that this is not the best approach UX-wise.

EmmanuelMess:
Three small issues with this PR on the light of [what the Material Guidelines say](https://material.io/guidelines/components/text-fields.html#text-fields-states):
* When retesting I noticed that if you select the second text field after not entering text in the first the "password cannot be empty" error doesn't appear (although it should as it is a focus change).
* Inputting "abs" in the first text field and "123" in the second, then returning to the first should prompt the "missmatch" error.
* If you press OK after not entering a password the "empty" error appears, if you enter text in the first text field it dissappears and deleting the text makes it appear again (this last thing shouldn't happen as there's no focus change for the error to appear).

I realize this is a bit too much hair splitting for a single dialog, and I won't stop the pulling of this PR for these.

VishalNehra:
I've fixed the first two, as for the third one, I think it's okay to remind user that password can't be empty.
As you start typing, and then decides to remove it, an empty field will trigger this. The usecase you mentioned is same as this.

EmmanuelMess:
> I think it's okay to remind user that password can't be empty.

[The guidelines disagree](https://material.io/guidelines/components/text-fields.html#text-fields-states).

Let me make make a PR to this branch. I don't want to bother you with something I didn't realize then was wrong.

VishalNehra:
Thanks! Latest changes should fix all problems.

PR #963, opened 2018-01-24
Fix opening content URIs in TextEditor

### LINKED ISSUE #426: Saving to content:// is still broken
I have downloaded the latest apk and looked at the current code. It seems, that saving to `content://` was not fixed and the broken behavior was made even more broken in recent changes by @martincz 

1) The text viewer still assumes that `uri.getPath` == path to local file. This is obviously not the case: if, say, opencloud ContentProvider exposes uri where `uri.getPath` == `/mnt/storage/example.txt`, that does not mean that you should mess with `/mnt/storage/example.txt` on local drive... What if the uri path accidentally happens to match something on local system? Say, `/dev/block/vdc`? Trying to blindly open that file with root shell (before even asking ContentProvider for file descriptor!) is not a good idea...

2) The same goes for filename. [The documentation](https://developer.android.com/reference/android/provider/OpenableColumns.html#DISPLAY_NAME) of OpenableColumns#FILE_NAME clearly states:
    > The human-friendly name of file. If this is not provided then the name should default to the the last segment of the file's URI.

    E.g. you should query the ContentProvider for `OpenableColumns#FILE_NAME` before assuming that `getLastPathSegment()` can be used as file name.

3) No check for null from `openFile`: if ContentProvider does not return a file, your app crashes.

4) If something happen to process (see p. 3), the file manager starts to crash upon reopening:

````
    java.lang.NullPointerException: Attempt to invoke virtual method 'boolean eu.chainfire.libsuperuser.Shell$Interactive.isRunning()' on a null object reference

### LINKED ISSUE #1014: Crash in TextReader with ACTION_VIEW and a null Uri
With Amaze 3.2.1, if another app uses `ACTION_VIEW` with a MIME type of `text/plain` but a `null` `Uri`, Amaze crashes with this:

```
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String com.amaze.filemanager.filesystem.BaseFile.getName()' on a null object reference
    at com.amaze.filemanager.activities.TextReader.onCreate(TextReader.java:207)
```

That is because [here](https://github.com/TeamAmaze/AmazeFileManager/blob/v.3.2.1/app/src/main/java/com/amaze/filemanager/activities/TextReader.java#L197-L207), you skip initializing `mFile` if the `Uri` is `null`, but then try calling `getName()` on `mFile` immediately thereafter.

Another app should not be providing a `null` `Uri` in this case, of course. But, still, ideally, you check if `mFile` is `null` before referencing it here.

The attached project can reproduce the error, though you will need to place a file named `hello_world.txt` in the `Download/` directory of external storage for it to work.
[MyApplication-20180211a.zip](https://github.com/TeamAmaze/AmazeFileManager/files/1714370/MyApplication-20180211a.zip)

### PR DESCRIPTION
**Pull request by EmmanuelMess**
Couldn't test if content URIs actually work. If someone knows how to create one, please tell me. The code should be [here](https://github.com/TeamAmaze/AmazeFileManager/pull/963/files#diff-706d9e9b88925a039b9492e107b67288R84). @Alexander--?
Fixes #426 
FIxes #1014
Also adds tests (#959)

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/WriteFileAbstraction.java
+++ b/app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/WriteFileAbstraction.java
@@ -1,0 +1,126 @@
+package com.amaze.filemanager.asynchronous.asynctasks;
+
+import android.content.ContentResolver;
+import android.content.Context;
+import android.os.AsyncTask;
+
+import com.amaze.filemanager.activities.superclasses.ThemedActivity;
+import com.amaze.filemanager.exceptions.ShellNotRunningException;
+import com.amaze.filemanager.exceptions.StreamNotFoundException;
+import com.amaze.filemanager.filesystem.EditableFileAbstraction;
+import com.amaze.filemanager.filesystem.FileUtil;
+import com.amaze.filemanager.filesystem.HybridFileParcelable;
+import com.amaze.filemanager.utils.OnAsyncTaskFinished;
+import com.amaze.filemanager.utils.RootUtils;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.lang.ref.WeakReference;
+
+/**
+ * @author Emmanuel Messulam <emmanuelbendavid@gmail.com>
+ *         on 16/1/2018, at 18:36.
+ */
+
+public class WriteFileAbstraction extends AsyncTask<Void, String, Integer> {
+
+    public static final int NORMAL = 0;
+    public static final int EXCEPTION_STREAM_NOT_FOUND = -1;
+    public static final int EXCEPTION_IO = -2;
+    public static final int EXCEPTION_SHELL_NOT_RUNNING = -3;
+
+    private WeakReference<Context> context;
+    private ContentResolver contentResolver;
+    private EditableFileAbstraction fileAbstraction;
+    private File cachedFile;
+    private OnAsyncTaskFinished<Integer> onAsyncTaskFinished;
+
+    private String dataToSave;
+
+    public WriteFileAbstraction(Context context, ContentResolver contentResolver,
+                                EditableFileAbstraction file, String dataToSave, File cachedFile,
+                                OnAsyncTaskFinished<Integer> onAsyncTaskFinished) {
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by Alexander--** [comment]
> Couldn't test if content URIs actually work. If someone knows how to create one, please tell me.

@EmmanuelMess

Have you tried Google Drive? IIRC, they create bonna-fide `content://` Uris without `_path` column. Alternatively, try OwnCloud (or whatever they have forked it into).

**Comment by EmmanuelMess** [comment]
@TranceLove Mind explaining add257b?

**Comment by EmmanuelMess** [comment]
Okay, content URIs work OK to read, don't know to write.

File: app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/WriteFileAbstraction.java [hunk 1]
```diff
@@ -0,0 +1,128 @@
...
+    @Override
+    protected Integer doInBackground(Void... voids) {
+        try {
+            OutputStream outputStream;
+
+            switch (fileAbstraction.scheme) {
+                case EditableFileAbstraction.SCHEME_CONTENT:
+                    if(fileAbstraction.uri == null) throw new NullPointerException("Something went really wrong!");
+
+                    outputStream = contentResolver.openOutputStream(fileAbstraction.uri);
```
Alexander--: You should always catch RuntimeException when you interact with ContentProviders.

ContentProviders are essentially untrusted third-party apps, written by often clueless people. There is little isolation to protect your code from bugs in them.

ContentResolver offers some degree of isolation to you, but it will throw SecurityException if remote ContentProvider denies you access. It will also rethrow most RuntimeExceptions, thrown by remote content provider. E.g. if remote provider accidentally throws NullPointerException, and you don't handle it, *your* app will crash.

TranceLove:
> @TranceLove Mind explaining add257b?

Never mind. Still work in progress, far from complete - yes, tests are broken. While Roboletric is cool, still need time to figure out how it can replace Android's recommended UI test practices.

app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/WriteFileAbstraction.java
EmmanuelMess: Okay, will do, thanks.

PR #972, opened 2018-01-29
Zip filename dialog should allow only values of *.zip

### LINKED ISSUE #954: Crash Occurs When Naming a Compressed File '/'
#### Expected Behavior
Naming a Compressed File '/' results in an invalid name error.

#### Actual Behavior
Naming a Compressed File '/' results in a Crash.

#### How to Reproduce
1. Launch Amaze File Manager 3.2.1 on Android 8.1.0
2. Select a File
3. Click the 3 Vertical Dots icon at the very top right
4. Select 'Compress'
5. Name the file '/'
6. Select 'Create'

---
    Device: Pixel 2 XL
    OS: Android 8.1.0
    App Version: 3.2.1
---

#### Recording of the Bug
crashcompressedslash

<br /><hr/><em>Posted on <a href="https://utopian.io/utopian-io/@thewizard.studio/crash-occurs-when-naming-a-compressed-file">Utopian.io -  Rewarding Open Source Contributors</a></em><hr/>

TranceLove:
Use regex to check zip filename input, and blocks Create button until valid filename (`*.zip`) is entered.

Fixes #954.


EmmanuelMess:
This is Linux, why the need for extensions?

TranceLove:
As far as I can see from the compress file dialog, it asks user the filename of the zip, so I suppose it should be *.zip per current behaviour.

Outside Android/Linux file extension will be important to them - they don't have `file` to detect the content.

EmmanuelMess:
As Linux isn't so widespread I'll leave this in.

Alexander--:
@TranceLove This sounds like a bad idea to me. What if I want to unpack an apk archive? Microsoft docx archive?

Android does not have a `file` program, but it is possible to compile it as shared library and use it in the app (see https://github.com/chdir/fdutil/blob/master/provider/src/main/cpp/javamagic.c)

More importantly, there is no need to validate whether each file is a zip archive before opening it — as long as you  can open it as zip archive it IS a zip archive. And if file has *.zip extension, but is not a valid zip archive, a file manager should be able to handle that correctly anyway.

TranceLove:
> @TranceLove This sounds like a bad idea to me. What if I want to unpack an apk archive? Microsoft docx archive?

But I suppose this is a create zip file dialog?

Alexander--:
@TranceLove the same argument applies both to creation and unpacking. You have been talking about `file` command, so I believe, that you should understand my point here.

TranceLove:
But I only changed `GeneralDialogCreation.showCompressDialog()`... I'm not sure it will affect unpacking any other archives at all.

EmmanuelMess:
What he's saying is that if someone wants to save a zip file as "something.apk" they can't (and should be able to).

TranceLove:
Sold. See 2e6cccbe.

Alexander--:
> I'm not sure it will affect unpacking any other archives at all.

After reading the title and your comments here, I have been under impression, that you intend to apply the same approach to other places as well. That was before actually reading the code of this PR, so I am not going to argue about that. But I retain the same opinion about these changes after looking at commits.

> What he's saying is that if someone wants to save a zip file as "something.apk" they can't (and should be able to).

Indeed, — sometimes users may have to unpack non-zip files, so it makes sense to allow packing them back without additional step to change file extension.

@TranceLove as for the particular changes in this PR, — if you want to sanity-check a filename, consider at least checking for forbidden symbols (`'/'`, `'\0'`, and possibly few others).

app/src/main/java/com/amaze/filemanager/ui/dialogs/GeneralDialogCreation.java
```diff
@@ -757,7 +757,7 @@ public void onNegative(MaterialDialog materialDialog) {
         b.show();
     }
 
-    static final Pattern ZIP_FILE_REGEX = Pattern.compile(".*\\.zip", Pattern.CASE_INSENSITIVE);
+    static final Pattern ZIP_FILE_REGEX = Pattern.compile("[\\\\\\/:\\*\\?\"<>\\|\\x01-\\x1F\\x7F]", Pattern.CASE_INSENSITIVE);
```
EmmanuelMess: Does this translate to `[\\\/:\*\?"<>\|\x01-\x1F\x7F]`? Because it's not getting the matches in [RegExr](https://regexr.com/).

TranceLove: The regex requires a global flag to work, hence line 792 uses `find()` instead of `matches()`.

```
!ZIP_FILE_REGEX.matcher(etFilename.getText().toString()).find()
```

EmmanuelMess: Ok, tested in emulator, added an "a" to ".zip", deleted the dot, ("azip") and pressed create, it crashed (twice, with two different files): 
```
E/AndroidRuntime: FATAL EXCEPTION: main
                  Process: com.amaze.filemanager, PID: 7757
                  java.lang.NullPointerException: Attempt to invoke virtual method 'android.text.Editable android.widget.EditText.getText()' on a null object reference
                      at com.amaze.filemanager.ui.dialogs.GeneralDialogCreation.lambda$showCompressDialog$18$GeneralDialogCreation(GeneralDialogCreation.java:777)
                      at com.amaze.filemanager.ui.dialogs.GeneralDialogCreation$$Lambda$18.onClick(Unknown Source)
                      at com.afollestad.materialdialogs.MaterialDialog.onClick(MaterialDialog.java:410)
                      at android.view.View.performClick(View.java:5637)
                      at android.view.View$PerformClick.run(View.java:22429)
                      at android.os.Handler.handleCallback(Handler.java:751)
                      at android.os.Looper.loop(Looper.java:154)
                      at android.app.ActivityThread.main(ActivityThread.java:6119)
                      at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886)
                      at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776)
```

TranceLove: Oops, my bad. 2fc2f6b.

EmmanuelMess:
I just realized, you should show an error, so the user knows why the button is greyed out. 
But I think not allowing the user to use any filename is not the correct approach, it would be better to warn them instead, maybe show  extra text under the TextField saying 'You may want to add .zip to the end' with a '?' button to a longer explanation.
Also you should show the error when user changes focus from the TextField (that means not greying out OK button (see PR #961)

TranceLove:
> maybe show extra text under the TextField saying 'You may want to add .zip to the end' with a '?' button to a longer explanation.

Personally I have some reservation on displaying too much help text, which can be annoying to some users; just use red with validation error messages to alert user should be sufficient. So I didn't implement the ? button at the hint text - and in fact it may make a simple dialog too complicated.

(And in fact, couldn't find an example for reference... ;) )

But I did implemented them anyway.

screenshot_1517392360

screenshot_1517392349

screenshot_1517392355

app/src/main/java/com/amaze/filemanager/ui/dialogs/GeneralDialogCreation.java
```diff
@@ -753,32 +758,103 @@ public void onNegative(MaterialDialog materialDialog) {
         b.show();
     }
 
+    static final Pattern ZIP_FILE_REGEX = Pattern.compile("[\\\\\\/:\\*\\?\"<>\\|\\x01-\\x1F\\x7F]", Pattern.CASE_INSENSITIVE);
+
+    static final class CompressFileDialogTextValidator extends SimpleTextWatcher implements View.OnFocusChangeListener
+    {
+        final WeakReference<Context> mContext;
```
EmmanuelMess: Why use a weak reference?

app/src/main/java/com/amaze/filemanager/ui/dialogs/GeneralDialogCreation.java
```diff
@@ -753,32 +758,103 @@ public void onNegative(MaterialDialog materialDialog) {
         b.show();
     }
 
+    static final Pattern ZIP_FILE_REGEX = Pattern.compile("[\\\\\\/:\\*\\?\"<>\\|\\x01-\\x1F\\x7F]", Pattern.CASE_INSENSITIVE);
+
+    static final class CompressFileDialogTextValidator extends SimpleTextWatcher implements View.OnFocusChangeListener
```
EmmanuelMess: This class should probably be in views (maybe extending TextView?).

app/src/main/java/com/amaze/filemanager/ui/dialogs/GeneralDialogCreation.java
```diff
@@ -753,32 +758,103 @@ public void onNegative(MaterialDialog materialDialog) {
...
+    {
+        final WeakReference<Context> mContext;
+        final EditText mEditText;
+        final TextInputLayout mTextInputLayout;
+        final View mPositiveButton;
+
+        CompressFileDialogTextValidator(Context context, EditText editText, TextInputLayout textInputLayout, View positiveButton)
+        {
```
EmmanuelMess: Brackets in the same line as the signature.

app/src/main/java/com/amaze/filemanager/ui/dialogs/GeneralDialogCreation.java
```diff
@@ -753,32 +758,103 @@ public void onNegative(MaterialDialog materialDialog) {
...
+
+        @Override
+        public void afterTextChanged(Editable s) {
+            doValidate();
+        }
+
+        @Override
+        public void onFocusChange(View v, boolean hasFocus)
+        {
+            if(!hasFocus)
```
EmmanuelMess: As per guidelines do:
```
if(!hasFocus) {
    doValidate();
}
```

EmmanuelMess:
I'll do a PR shortly.

app/src/main/java/com/amaze/filemanager/ui/dialogs/GeneralDialogCreation.java
TranceLove: Agreed. Maybe a `View` composing `TextInputLayout` and `EditText` but reusable is a better approach.

EmmanuelMess: An XML layout!

TranceLove: Sure. On the way.

TranceLove:
Let's see if the changes from e309595 can fit.

app/src/main/res/layout/dialog_singleedittext.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
EmmanuelMess: #983.

app/src/main/res/layout/dialog_singleedittext.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
+<android.support.constraint.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
EmmanuelMess: The ".zip" in the dialog EditText dissapeared. And it seems like there's too much space between the EditText and the buttons. See here the space between the inner text and the buttons: 


PS: May be `android:layout_height="wrap_content"`?

TranceLove: See b70dfd5.

However I'm sure if I should or able to fix the buttons in the dialog... as it's from the material-dialogs library. Going further may led us end up constructing the whole dialog on our own but giving up the convenience of material-dialogs.

Alexander--: @EmmanuelMess

> And it seems like there's too much space between the EditText and the buttons.

This is inherent flaw of `TextInputLayout`, — it reserves space for error message underneath it's EditText. If you remove the space, the buttons will jump around each time error message visibility changes.  IMO, if you want the validation feedback to appear quickly, a bit of extra empty space is the lesser of evils.

EmmanuelMess: @Alexander-- Hold my beer.

TranceLove: With 92dc754 @EmmanuelMess introduced shrinking space for EditText's error/warning messages when it's not in use.

While I personally have reservation on this behaviour, if everyone think it's OK, I'll be OK with it.

EmmanuelMess:
If everyone is fine with it I will merge into a feature branch untill all TextEdits show the same kind of behavior, then that branch can be pulled into master.

TranceLove:
Should I close this in favor of #1011?

EmmanuelMess:
Yes

TranceLove:
Closing in favor of #1011.

PR #991, opened 2018-02-04
Added TAR viewing

EmmanuelMess:
* Abstracted compressed or archived viewing
* Abstracted complete and partial decompression of zip, rar and untaring
* Fixed compressed icons
* Simplified code to check if file viewing and decompressing is supported


TranceLove:
I tested with ubuntu's rootfs tarball for ARM which is floating around my dev machine. It crashed at filenames having colons (`/var/lib/dpkg/info/libudev1:armhf.symbols` actually).

```
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String android.content.Intent.getStringExtra(java.lang.String)' on a null object reference
    at com.amaze.filemanager.asynchronous.services.ExtractService.onStartCommand(ExtractService.java:91)
```

I too think it's a atypical case that nobody would be extracting such tarballs like me. See if you'd want to handle this... otherwise I'd give a go first.

TranceLove:
Sorry for removing a comment made yesterday since the comment was made while testing on another changeset, thereby making the comment invalid.

The PR is tested OK and good to go - approved status is still valid. However, may I ask the reason behind removing .gz from the extractables, which was working properly before this PR? Perhaps another PR to address this change?

EmmanuelMess:
GZIP is not extractable by TAR nor ZIP, it uses another stream, it errored.

EmmanuelMess:
Okay, found my mistake, don't pull yet.

EmmanuelMess:
Issues have been fixed and gzipped tarballs can be extracted.

EmmanuelMess:
Issues have been fixed and gzipped tarballs can be extracted.

TranceLove:
Tested OK with both tarballs and gzipped tarballs. I'm letting this through.

EmmanuelMess:
@shubhwicked Please open an issue indicating Android version and Amaze version.

PR #1006, opened 2018-02-09
Traditional Chinese translation update

TranceLove:


EmmanuelMess:
+1 (FastHub merge broke (404 error), FastHub review is premium feature and black out so can't use PC)

Issue #1020, opened 2018-02-12
Warn on input checklist


EmmanuelMess:
This is all the EditFields that should be modified to recommend a specific extension:

- [x] Compress dialog's
- [x] Text file creation
- [x] Encrypted file creation ~~?~~

TranceLove:
Will look into them after sunrise.

However, out of curiosity and in fact didn't look into the insides of file encryption yet... but do we need to allow users to specify another file extension on encrypting files? I thought files encrypted by Amaze had been designated a fixed file extension, although I think it is possible to decrypt the files outside Amaze?

EmmanuelMess:
Encryption is horrendously broken by my standards, and nothing identifies the encrypted file but the extension, so it would be obligatory.

VishalNehra:
'broken'?  take an example of old blackberry devices. They could only understand their own extension (.rem). There is no standard to this unless you find some library extension which doesn't make sense since we've already implemented it.

VishalNehra:
Decryption outside amaze is not possible

EmmanuelMess:
[Magic number](https://en.wikipedia.org/wiki/List_of_file_signatures)

VishalNehra:
Making folder a single file before encryption? Doesn't make sense to me... The only thing that I think can be done is to change encrypted file names in directories to some random gibberish and only the name of directory encrypted to end with .aze.

VishalNehra:
Ps. .aze stands for amaze encrypted. Tell this to the guy who prepared that wiki page to add it to the list.

EmmanuelMess:
>Making folder a single file before encryption? Doesn't make sense to me... The only thing that I think can be done is to change encrypted file names in directories to some random gibberish and only the name of directory encrypted to end with .aze.

You mean "can" as in is possible or should be done? Because tarballing a folder is easy. And I think I already explained that gibberish filenames are horrendously confusing when you have 20 text files and need to open *the* one.

> Ps. .aze stands for amaze encrypted. Tell this to the guy who prepared that wiki page to add it to the list.

Please tell me this is a joke, that list references extension, file type and *header*.

TranceLove:
Ah... the discussion goes way off the topic. I'd suggest leaving file encryption to a separate issue (which seems very big).

However, naming an encrypted folder with an `.aze` ala how XCode/OSX handles xcode project folders seems a good way to go.

EmmanuelMess:
Yes

[...]

PR #1021, opened 2018-02-13
Use generic dialog layout and WarnableTextInputLayout for mkfile, mkdir and rename (#1020)

TranceLove:
While working on the dialogs for create files and folders, I also noticed the rename dialog was using its own dialog, so I migrated it too.

Tried to keep `GenericDialogCreation.showNameDialog()` as generic as possible, hence passing the callbacks in the method.

See if you liked this refactoring.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/utils/MainActivityHelper.java
+++ b/app/src/main/java/com/amaze/filemanager/utils/MainActivityHelper.java
@@ -135,13 +135,13 @@ public class MainActivityHelper {
     }
 
     private void mk(@StringRes int newText, final OnClickMaterialListener l) {
-        final MaterialDialog materialDialog = GeneralDialogCreation.showNameDialog(mainActivity,
-                new String[]{mainActivity.getResources().getString(R.string.entername),
-                        "",
-                        mainActivity.getResources().getString(newText),
-                        mainActivity.getResources().getString(R.string.create),
-                        mainActivity.getResources().getString(R.string.cancel),
-                        null});
+        final MaterialDialog materialDialog = GeneralDialogCreation.createNameDialog(mainActivity,
+                mainActivity.getResources().getString(R.string.entername),
+                "",
+                mainActivity.getResources().getString(newText),
+                mainActivity.getResources().getString(R.string.create),
+                mainActivity.getResources().getString(R.string.cancel),
+                null);
 
         materialDialog.getActionButton(DialogAction.POSITIVE).setOnClickListener(v -> l.onClick(materialDialog));
         materialDialog.show();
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/amaze/filemanager/utils/MainActivityHelper.java
```diff
@@ -127,28 +141,40 @@ void mkdir(final OpenMode openMode, final String path, final MainFragment ma) {
      * @param ma       {@link MainFragment} current fragment
      */
     void mkfile(final OpenMode openMode, final String path, final MainFragment ma) {
```
EmmanuelMess: The ".txt" is missing when the dialog shows.

app/src/main/java/com/amaze/filemanager/utils/Utils.java
```diff
@@ -38,6 +39,7 @@
     private static final String INPUT_INTENT_BLACKLIST_PIPE = "\\|";
     private static final String INPUT_INTENT_BLACKLIST_AMP = "&&";
     private static final String INPUT_INTENT_BLACKLIST_DOTS = "\\.\\.\\.";
+    private static final Pattern FILENAME_REGEX = Pattern.compile("[\\\\\\/:\\*\\?\"<>\\|\\x01-\\x1F\\x7F]", Pattern.CASE_INSENSITIVE);
```
EmmanuelMess: Off topic: I can't really read this RegEx, but it seems to be filtering [more than it should](https://unix.stackexchange.com/a/230299/274342). Not sure how Android treats valid filenames though.

app/src/main/java/com/amaze/filemanager/utils/Utils.java
```diff
@@ -231,4 +233,16 @@ public static int nthToLastCharIndex(int elementNumber, String str, char element
         return -1;
     }
 
+    /**
+     * Validate given text is a valid filename.
+     *
+     * @param text
+     * @return true if given text is a valid filename
+     */
+    public static boolean isValidFilename(String text) {
```
EmmanuelMess: Also this should be in FileUtils, no (more) file stuff in Utils.

app/src/main/java/com/amaze/filemanager/utils/Utils.java
Alexander--: That regexp is for Windows compatibility. Even if Linux filesystems allow for most of those symbols, such files can't be copied to FAT-formatted SD card — Windows filesystem drivers won't able to interact with them.

EmmanuelMess: Very well

PR #1023, opened 2018-02-13
Duplicated items (second) solution

EmmanuelMess:
This (as opposed to #822) synchronizes on the final concurrent Lists to fix the duplication issue.


VishalNehra:
As I can test, items still get duplicated once you scroll around in the same screen.

Issue #977, opened 2018-01-31
Update Proguard and enable it


EmmanuelMess:
Proguard helps maintain apk size down, but for it to work some libs have special requirements, I think CloudRAIL was one of them.

VishalNehra:
For Cloudrail, [this](https://github.com/CloudRail/cloudrail-si-android-sdk/issues/34) should be able to bypass it.

EmmanuelMess:
The snippet:
`-keep class com.cloudrail.** { *; }`

EmmanuelMess:
@TranceLove Please update proguard with rules relating to sshj and other networking libs you have added as I don't know how those libs work.

PR #1025, opened 2018-02-14
Proguard configuration of BouncyCastle/SpongyCastle (for sshj)

TranceLove:
From request at #977. To reiterate, **it's for sshj to work, may break file encryption/decryption**.

sshj itself ~~seems doesn't~~ needs proguard configurations, ~~but~~ and its dependency, SpongyCastle and BouncyCastle does too.

Verifications:
 - Setup SMB/SFTP connections with username/password
 - Setup SFTP connections with SSH keys with/without passphrases
 - Open SFTP connection, copy file from SSH server
 - SMB/SFTP connections still work after restarting app

P.S. At some point I was thinking of omitting BouncyCastle since SpongyCastle can do exactly the same thing, but sshj's behaviour (as of 0.23.0) that [it'd jump to default JCE provider](https://github.com/hierynomus/sshj/blob/v0.23.0/src/main/java/net/schmizz/sshj/common/SecurityUtils.java#L247-L260) forced me to rollback. It's fixed in their master (hierynomus/sshj#392), but we shall need to wait for a new version published in the repos.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -38,8 +38,7 @@ android {
 
     buildTypes {
         debug {
-            minifyEnabled true
-            shrinkResources true
+            minifyEnabled false
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard.cfg'
 
         }
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -38,7 +38,8 @@ android {
 
     buildTypes {
         debug {
-            minifyEnabled false
+            minifyEnabled true
+            shrinkResources true
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard.cfg'
 
         }
--- a/app/proguard.cfg
+++ b/app/proguard.cfg
@@ -166,5 +166,28 @@
     public static final int define_*;
 }
 
+#From SpongyCastle, sshj dependency
+-keep class org.spongycastle.**
+-keep class org.bouncycastle.**
+
+#We are not using these anyway
+-dontwarn org.spongycastle.jce.provider.X509LDAPCertStoreSpi
+-dontwarn org.spongycastle.x509.util.LDAPStoreHelper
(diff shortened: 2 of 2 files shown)


TranceLove:
Wow, travis log showed a lot of things that I didn't see even with a clean build from my machine.

Surely you won't pull this...

app/proguard.cfg
```diff
@@ -166,5 +166,28 @@
...
+-dontwarn net.schmizz.sshj.SSHClient
+
+#ignore test classes
+-dontwarn android.test.**
+-dontwarn org.junit.**
+
 # for DexGuard only
--keepresourcexmlelements manifest/application/meta-data@value=GlideModule
\ No newline at end of file
+#-keepresourcexmlelements manifest/application/meta-data@value=GlideModule
```
EmmanuelMess: Why comment this line?

app/build.gradle
```diff
@@ -38,7 +38,8 @@ android {
 
     buildTypes {
         debug {
-            minifyEnabled false
+            minifyEnabled true
```
EmmanuelMess: Move these lines to release so debugging is faster.

app/proguard.cfg
TranceLove: Had problem running from Android studio with this enabled.

```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:transformClassesWithAndroidGradleClassShrinkerForDebug'.
> ProGuard configuration parser error: /home/trancelove/git/AmazeFileManager/app/proguard.cfg line 193:0 missing EOF at '-keepresourcexmlelements'
```

And I suppose we are not using the paid Dexguard but the free Proguard when building releases? Anyway, will uncomment this and let you decide.

EmmanuelMess:
Don't pull just because I accepted the changes, remember I don't yet know how to test external connections.

EmmanuelMess:
Related #1093

TranceLove:
Closing, in favor of #1093.

PR #1026, opened 2018-02-14
 Migrate PEM passphrase dialog to use generic dialog layout

TranceLove:
Also merged `master`; actual commit is at f8b5c9f.

Though no warning message would come up from PEM passphrase dialog (PEM without passphrase won't bring this dialog up), migrate anyway for user interface uniformity.


EmmanuelMess:
No idea how to test this, @VishalNehra?

TranceLove:
> No idea how to test this, @VishalNehra?

Back at #863 @VishalNehra handed in a set of keys for testing. [Here](https://drive.google.com/open?id=1DWPylxP3OSmclfpBUvIln7BVTpc-5slE), but not `authorized_keys` which is for the server side.

I didn't know the passphrase... and I don't need to know this either ;)

And for a positive case, try this, [id_rsa.encrypted.txt](https://github.com/TeamAmaze/AmazeFileManager/files/1726251/id_rsa.encrypted.txt). Passphrase = 123456

(And I suppose you'd know you won't put these keys into any SSH servers you'd be using)

VishalNehra:
It doesn't seem to show error in edit text if the password was wrong just popup again asking for passphrase. I thought this PR was for fixing this behavior since you've used TIL in the commit. Or this wasn't the intention of this PR?

TranceLove:
Oops. Will fix.

VishalNehra:
OK from my side

PR #1038, opened 2018-02-20
Fix ColorAdapter

EmmanuelMess:
* Moved ColorAdapter to its own file, removed unnecessary fields
* Reuse recycled view, delete always gone TextView and cleanup
* Simplified and rewrote view for ColorPicker


EmmanuelMess:
Forgot to apply antialiasing to the circles.

PR #1051, opened 2018-02-23
 All custom View measurements are now percentual 

### LINKED ISSUE #1045: Constants in Views must be percentual
Constants in Views must be percentual, so that changing ppi doesn't change the relative size of things.

EmmanuelMess:
All custom View measurements are now percentual instead of by pixel measurements.
Fixes #1045


TranceLove:
Looks good. I'd vote a +1.

PR #1048, opened 2018-02-23
Change size of Linux tux icon for hdpi to 36x36 (#1031)

TranceLove:
The Tux icon for hdpi resolution was 24x24 which should be 36x36 (referencing other icons), so I made one by resizing the xxxhdpi one which was 96x96 to 36x36.

Weights 1297 bytes which is bigger than `ic_remote_grey600` which weights just 831 bytes, not sure I'm doing the icon in the correct way...

Fixes #1031.

PR #1056, opened 2018-02-24
Replace raster icons in cloud connection fragment with vector versions

TranceLove:
As addendum to #1048, replacing icons in Cloud Connection Fragment with vector icons found at [Material Design Icons](https://materialdesignicons.com/).

Note: you will need Amaze's Cloud plugin to verify the correctness of the icon display (if I don't hack the code to force the icons to appear).


EmmanuelMess:
Looks OK in my phone:

PR #1060, opened 2018-02-26
Fix drawer width

VishalNehra:
Fixes drawer width in landscape mode
screenshot_20180226-115159

### DIFF
--- a/app/src/main/res/layout-v21/main_toolbar.xml
+++ b/app/src/main/res/layout-v21/main_toolbar.xml
@@ -72,7 +72,7 @@
 
     <android.support.design.widget.NavigationView
         android:id="@+id/navigation"
-        android:layout_width="wrap_content"
+        android:layout_width="@dimen/drawer_width"
         android:layout_height="match_parent"
         android:layout_gravity="start"/>
 
--- a/app/src/main/res/layout-w720dp/main_toolbar.xml
+++ b/app/src/main/res/layout-w720dp/main_toolbar.xml
@@ -88,7 +88,7 @@
         android:fitsSystemWindows="true">
         <android.support.design.widget.NavigationView
             android:id="@+id/navigation"
-            android:layout_width="wrap_content"
+            android:layout_width="@dimen/drawer_width"
             android:layout_height="match_parent"
             android:layout_gravity="start"/>
     </android.support.v4.widget.DrawerLayout>
--- a/app/src/main/res/layout/main_toolbar.xml
+++ b/app/src/main/res/layout/main_toolbar.xml
@@ -75,7 +75,7 @@
     </android.support.design.widget.CoordinatorLayout>
     <android.support.design.widget.NavigationView
         android:id="@+id/navigation"
-        android:layout_width="wrap_content"
+        android:layout_width="@dimen/drawer_width"
         android:layout_height="match_parent"
         android:layout_gravity="start"/>
 


EmmanuelMess:
AFAIK NavigationView has the standard width, why make it wider? Making the white space smaller would be better.

TranceLove:
OK, I read [the spec](https://material.io/guidelines/patterns/navigation-drawer.html) via [here](https://android.jlelse.eu/navigation-drawer-styling-under-material-design-f0767882e692) saying

> Maximum width: The maximum width of the nav drawer is 280dp on mobile and 320dp on tablet. This is calculated by multiplying the standard increment by five (the standard increment is 56dp on mobile and 64dp on tablet).

I think it's the squeezed drawer header image that made us feel uncomfortable?

VishalNehra:
@EmmanuelMess navigation view is also a view just like any other. You set it to wrap_content it would just wrap around the view bounds. You gotta set it to a defined width. And ````@dimen/drawer_width```` is a standard width followed by many material design apps including Google's own apps.

EmmanuelMess:
Google doesn't really follow the Material Guidelines...
Also as per [Documentation](https://developer.android.com/reference/android/support/design/widget/NavigationView.html):

> Represents a standard navigation menu for application. The menu contents can be populated by a menu resource file.
> 
> NavigationView is typically placed inside a DrawerLayout.
> ```xml
>  <android.support.v4.widget.DrawerLayout xmlns:android="http://schemas.android.com/apk/res/android"
>      xmlns:app="http://schemas.android.com/apk/res-auto"
>      android:id="@+id/drawer_layout"
>      android:layout_width="match_parent"
>      android:layout_height="match_parent"
>      android:fitsSystemWindows="true">
> 
>      
> 
>      <android.support.design.widget.NavigationView
>          android:id="@+id/navigation"
>          android:layout_width="wrap_content"
>          android:layout_height="match_parent"
>          android:layout_gravity="start"
>          app:menu="@menu/my_navigation_items" />
>  </android.support.v4.widget.DrawerLayout>
> ```

VishalNehra:
But we don't use menu resource for navigation drawer, do we?

EmmanuelMess:
This is the view, not the menu resource, and we do use the view (in XML).

VishalNehra:
It won't... Regardless, I set that width in past after consulting various sources and through various feedbacks of users. Changing width to 280dp might be according to guidelines as told by @TranceLove but it might ruin UX.

EmmanuelMess:
And what about changing the size of the white section?

VishalNehra:
We didn't create the white section. It's because of the difference between the default width of fragment and ````wrap_content```` width of navigation drawer. This gets fixed with a default width which we've been using for long enough.
What else do you propose?

VishalNehra:
Looks too narrow.

EmmanuelMess:
Does it cut text? If that isn't the case I would vote for the guidelines' measures.

VishalNehra:
It doesn't but looks ugly. Like I said, many google apps themselves use this instead of guidelines's.

EmmanuelMess:
The other option seems to be `screen width - appbar height`, this is what you prefer, right? Could you provide screenshots?

VishalNehra:
Please create another PR if you feel like changing the drawer width, we'll discuss about it there. This one is focused on fixing what seems to be broken.

EmmanuelMess:
Ok, I'll merge as a fix, and create an issue to follow the Design Guidelines, because I don't think I'll have the time to create a PR.

Issue #1070, opened 2018-02-28
Tab 1 is right and tab 2 is left.


EmmanuelMess:
Tab 1 is right and tab 2 is left. They should be inverted.
When fixing this bug beware of database data.

VishalNehra:
Why should this be inverted? Right tab is default and primary, I don't plan on changing that.

EmmanuelMess:
I usually start counting from the left, but I suppose we can leave as is.

PR #1071, opened 2018-02-28
Fixed default tab storage

### LINKED ISSUE #1043: Both tabs start with root set as home
When Amaze is installed for the first time, both tabs are set to root as home. I suspect this is due to new changes in rewritten navigation drawer code.
Please make changes so that left one is set to root and right tab which is set by default is set to storage.

### LINKED ISSUE #1067: Home for right tab set to '/' instead of internal storage
The default home for right tab is set to root instead of internal storage.

EmmanuelMess:
* Code cleanup in TabFragment
* Fixed storage paths not getting saved correctly in Drawer
* Better handling of null storages when creating Tabs
* If the home for a tab is set to root the rootmode pref is activated

Fixed #1043 
Fixed #1067
and not #1070

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/fragments/TabFragment.java
+++ b/app/src/main/java/com/amaze/filemanager/fragments/TabFragment.java
@@ -9,6 +9,7 @@ import android.os.Build;
 import android.os.Bundle;
 import android.preference.PreferenceManager;
 import android.support.annotation.ColorInt;
+import android.support.annotation.NonNull;
 import android.support.v4.app.Fragment;
 import android.support.v4.app.FragmentManager;
 import android.support.v4.app.FragmentStatePagerAdapter;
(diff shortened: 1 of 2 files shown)


TranceLove:
Seems it can fix mentioned problems when tested on Lollipop emulator.

If the conclusions of #1043 and #1067 is to make both tabs as internal storage as said

> Neither should be root, as on first open the "rootmode" preference wouldn't be set to true.

I'd vote a +1.

EmmanuelMess:
It will resort to making both internal storage if it doesn't currently have another storage. If there is no internal storage it will activate "rootmode" and set both to root.

EmmanuelMess:
If plug a mass storage device on first start of Amaze it can be set as home (if no SD is present), if the device is then removed I have no idea what path the tab will show.

EmmanuelMess:
@TranceLove?

TranceLove:
Yeap, I'm good. +1.

PR #1085, opened 2018-03-04
Save the scroll state of PreferenceFragments

### LINKED ISSUE #1075: Back button brings the screen at the top of menu
I am using Amaze for some time now and it is really the nicest one I have used till. While using it I noticed a small glitch. I was tweaking some settings and went into a sub menu, when I pressed back it sure got out of the sub menu but went at the beginning of setting tab, not where I was working. I had to scroll back all the way to bottom to see the change again. I checked again for a few time and the result is same.
I am using v3.2.1. Hope this issue will be solved soon.
Thanks !

EmmanuelMess:
* Save the scroll state of PreferenceFragments in the activity so that they can be restored, currently only implemented for start PreferenceFragment.

Fixes #1075.


TranceLove:
Verified working with Lollipop emulator and my Oneplus One running 7.1.2. I'd vote a +1.

PR #1086, opened 2018-03-04
SharedPreferences booleans are obtained each time they have to be used

EmmanuelMess:
* SharedPreferences booleans are obtained each time they have to be used.  
[Kurtis Nusbaum says I don't have to cache values](https://stackoverflow.com/a/8055131/3124150).
* SharedPreferences booleans have centralized default values.


VishalNehra:
Seems like a cleaner approach, still I'll wait for better source.

EmmanuelMess:
[Here](http://grepcode.com/file/repository.grepcode.com/java/ext/com.google.android/android/5.1.1_r1/android/app/SharedPreferencesImpl.java/) is the source code for the SharedPreference implementation.

PR #1087, opened 2018-03-05
Use FolderChooser to choose FTP server share path & archive extract/zip creation paths

### LINKED ISSUE #811: Change FTP server shared path with dir chooser
Small improvement request:
Currently changing the shared path of the FTP server is done by typing/editing the path in a text field. It would be a lot easier to have a directory chooser widget (or whatever this is called in the Android SDK) where I can browse the FS and choose the target directory.
Thanks!

TranceLove:
Fixes #811, using MaterialDialog's FolderChooser.

While working on #811, also took the time to change archive extraction and zip creation paths to use FolderChooser too.

Sorry for scattered callbacks of the folder chooser dialogs, but seems it's the way for FolderChooser to work.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -96,6 +96,7 @@ dependencies {
     }
 
     implementation 'com.afollestad.material-dialogs:core:0.9.6.0'
+    implementation 'com.afollestad.material-dialogs:commons:0.9.6.0'
 
     //FTP Server
     /*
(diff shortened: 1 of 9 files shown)


EmmanuelMess:
I do not recommend pulling this until after 3.3.0 is released to prevent bugs that have to be fixed at the last time or after release of v3.3.0.

Issues I found:
* Archive preferences don't seem to do anything.
* There is no way of returning to the 'default mode' (extracting in compressed folder).
* I propose changing the '...' (for going to parent folder) to '..'.
* Hidden folders show even if the option (in settings) is deactivated.
* Don't leave `@param`s empty, complete or delete.
* Shortcuts still has the old system for selecting paths.

Haven't tested FTP server.

TranceLove:
Will be superceded by #1895. Closing.

PR #1089, opened 2018-03-06
Preserve overwriting extraction path by service

### LINKED ISSUE #1088: Can't open files from CompressedExplorerFragment
Unable to open file from fragment, it just extracts to parent of compressed file.
Bug introduced in #837

VishalNehra:
Fixes #1088

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/activities/TextEditorActivity.java
+++ b/app/src/main/java/com/amaze/filemanager/activities/TextEditorActivity.java
@@ -44,6 +44,7 @@ import android.view.ViewGroup;
 import android.view.Window;
 import android.view.WindowManager;
 import android.view.animation.AccelerateDecelerateInterpolator;
+import android.view.inputmethod.EditorInfo;
 import android.view.inputmethod.InputMethodManager;
 import android.widget.EditText;
 import android.widget.FrameLayout;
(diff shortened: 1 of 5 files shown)


TranceLove:
After applying it is now possible to open a file inside the zip without extracting first that #1088 mentioned.

I'd vote a +1.

VishalNehra:
Commit added to fix #1090

VishalNehra:
I think user should be expecting that much. Regardless, I'm adding new changes to reflect the same.

EmmanuelMess:
Most of our users won't know what " (r/o)" means. My recommendation, a dialog that says something like "This file can only be read, you cannot make changes".

VishalNehra:
Let's not consider users this dumb. Save button is disabled by default and it won't be visible unless you make any change. Showing dialog to user everytime he opens a text file from compressed file would be too reduntant, user would get annoyed. Let's keep it simple.

EmmanuelMess:
It's computational jargon, users aren't dumb, it's just that the meaning of "r/o" doesn't concern them.

EmmanuelMess:
Maybe a Snackbar intead of dialog?

VishalNehra:
@EmmanuelMess I agree and have made the changes suggested by you.

EmmanuelMess:
Will test

EmmanuelMess:
What do you think of the last commit @TranceLove ?

PR #1093, opened 2018-03-07
Proguard

### LINKED ISSUE #977: Update Proguard and enable it

EmmanuelMess:
* Added proguard rule for CloudRail, SpongyCastle
* Added @TranceLove 's rules for jssh
* Added proguard rules for tests (in another file)
* Removed '-dontwarn' rule and unnecessary lines in `proguard.cfg`
* Activated ProGuard for releases

Fixes #977


TranceLove:
Unable to connect to SSH servers with these proguard rules.

Having SpongyCastle alone is not sufficient for the SCP/SFTP part to function, since BouncyCastle is (as of 0.23.0, still) a dependency of [sshj](https://github.com/hierynomus/sshj), hence the regression.

Please wait, while I check if it's just copy-and-pasting SpongyCastle's rules to BouncyCastle would do.

EmmanuelMess:
I don't understand, why would Proguard break BouncyCastle, isn't it already in Android?

TranceLove:
> I don't understand, why would Proguard break BouncyCastle, isn't it already in Android?

Surely yes. However unfortunately from my testing without keeping BouncyCastle available sshj cannot won't be able to find them, hence not able to connect.

VishalNehra:
SFTP seems to be working fine for me for these rules. Can't test decryption as it's currently broken on master.

VishalNehra:
@TranceLove are you positive SFTP doesn't work for you after applying these rules?

EmmanuelMess:
I merged @TranceLove's fix.

TranceLove:
> @TranceLove are you positive SFTP doesn't work for you after applying these rules?

After 178dce5 it worked.

EmmanuelMess:
@VishalNehra?

PR #1099, opened 2018-03-09
Fix load SMB connections on app startup

TranceLove:
Steps to reproduce the symptom:

1. Fresh install
2. Create a new SMB connection
3. Verify SMB connection is loaded correctly, and drawer has the connection item
4. Close app by keep tapping back or force quit
5. Launch app again
6. The SMB connection created in 2 disappeared

The second, third... other SMB connections are not affected, only the first one. This fix is going to fix this glitch.

Fix is borrowed from code when I worked on #863, while keeping the `try/catch` block.


VishalNehra:
There are many such similar errors in this and other handler classes, please correct them also while you're at it. @TranceLove

TranceLove:
> There are many such similar errors in this and other handler classes, please correct them also while you're at it. @TranceLove

Noted.

PR #1108, opened 2018-03-13
Fix watcher progress (again)

### LINKED ISSUE #785: Crash on copy

### LINKED ISSUE #874: If file is copied in sd card, copy complete appears prematurely

### LINKED ISSUE #887: SMB/Samba issues when copying

VishalNehra:
Remake of #916 which was halted due to change in codebase and conflicts with master.

<repeat>
Notification won't show 'Copy completed' before actually completing it. In case there's no progress, the notification spinner will go indeterminate before waiting 5 secs for progress to resume. It'll resume from there when there's a progress.
Should also fix the race condition for starting the first service instance which some people faced earlier.

Fixes #785 
Fixes #874
Fixes #887


app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
```diff
@@ -190,4 +225,22 @@ public void run() {
...
+         * Progress has been halted for some reason
+         */
+        void progressHalted();
+
+        /**
+         * Future extension for possible implementation of pause/resume of services
+         */
+        void progressResumed();
+
+        <T extends Service> T getServiceType();
```
EmmanuelMess: Type erasure makes this getter unnecessarily complex, try `Service getService();`, although, seeing this interface is used solely in `ProgressiveService` it seems redundant.

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
```diff
@@ -120,26 +155,26 @@ public static synchronized void runService(final Context context, final Intent i
         /*if (handlerThread==null || !handlerThread.isAlive()) {
```
EmmanuelMess: Commented code, if not needed, delete.

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
```diff
@@ -120,26 +155,26 @@ public static synchronized void runService(final Context context, final Intent i
...
+         * in {@link #init(Context)}. If older service has returned, we already have the runnable
+         * waiting to execute in #init, and user is in app, and starts another service, and
+         * as this block executes the {@link android.app.Service#onStartCommand(Intent, int, int)}
+         * we end up with a context switch to 'service_startup_watcher' in #init, it also starts
+         * a new service (as {@link #progressHandler} is not alive yet).
+         * Though chances are very slim, but even if this condition occurs, only the progress will
+         * be flawed, but the actual operation will go fine, due to android's native serial service
+         * execution. #nough' said!
+         *//*
             context.startService(intent);
```
EmmanuelMess: More commented code.

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
```diff
@@ -25,26 +30,27 @@
...
     private static ArrayList<Intent> pendingIntents = new ArrayList<>();
 
     // position of byte in total byte size to be copied
-    public static long POSITION = 0L;
+    public static volatile long POSITION = 0L;
+
+    private static int HALT_COUNTER = -1;
 
-    private static int HAULT_COUNTER = -1;
+    public static final int ID_NOTIFICATION_WAIT =  9248;
```
EmmanuelMess: Shouldn't this go in `NotificationConstants`?

app/src/main/java/com/amaze/filemanager/asynchronous/services/ProgressiveServiceAbstract.java
```diff
@@ -0,0 +1,277 @@
...
+import com.amaze.filemanager.utils.ServiceWatcherUtil;
+
+import java.util.ArrayList;
+
+/**
+ * @author Emmanuel Messulam <emmanuelbendavid@gmail.com>
+ *         on 28/11/2017, at 19:32.
+ */
+
+public abstract class ProgressiveServiceAbstract extends Service implements ServiceWatcherUtil.ServiceWatcherInteractionInterface {
```
EmmanuelMess: Why rename to ...Abstract? There is no additional useful info added... Also, renaming all abstract classes to ...Abstract makes no sense.

EmmanuelMess: I still don't understand the use of ` ServiceWatcherUtil.ServiceWatcherInteractionInterface`, you can make the functions abstract and do away with the interface.

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
```diff
@@ -190,4 +225,22 @@ public void run() {
 
         handler.postDelayed(runnable, 0);
     }
-}
+
+    public interface ServiceWatcherInteractionInterface {
+
+        int STATE_HALTED = 0;
```
EmmanuelMess: Naming conventions, this variable in not final.

app/src/main/java/com/amaze/filemanager/asynchronous/services/ProgressiveServiceAbstract.java
```diff
@@ -0,0 +1,277 @@
...
+    private boolean isNotificationTitleSet = false;
+
+    @Override
+    public int onStartCommand(Intent intent, int flags, int startId) {
+
+        initVariables();
+        return super.onStartCommand(intent, flags, startId);
+    }
+
+    public abstract void initVariables();
```
EmmanuelMess: I don't think this is a good idea... Subclasses are initializing the superclass' fields, the superclass then uses the fields blindly, it seems to me that there has to be a better way. @TranceLove ?

app/src/main/java/com/amaze/filemanager/asynchronous/services/ProgressiveServiceAbstract.java
```diff
@@ -0,0 +1,277 @@
...
+    public Context context;
+    // list of data packages, to initiate chart in process viewer fragment
+    public ArrayList<DatapointParcelable> dataPackages;
+    public ProgressListener progressListener;
+
+    private boolean isNotificationTitleSet = false;
+
+    @Override
+    public int onStartCommand(Intent intent, int flags, int startId) {
+
```
EmmanuelMess: Please stop adding new lines after the '{'. There are quite a few...

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
```diff
@@ -25,26 +30,27 @@
     private Handler handler;
     private static HandlerThread handlerThread;
     private ProgressHandler progressHandler;
-    long totalSize;
     private Runnable runnable;
 
+    public static int STATE = -1;
```
EmmanuelMess: Naming conventions, this variable in not final.

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
```diff
@@ -25,26 +30,27 @@
...
+
     private static ArrayList<Intent> pendingIntents = new ArrayList<>();
 
     // position of byte in total byte size to be copied
-    public static long POSITION = 0L;
+    public static volatile long POSITION = 0L;
```
EmmanuelMess: Naming conventions, this variable in not final. Also, volatile is a strong word, are you sure there isn't another way? (Maybe 'atomic' is what you want).

EmmanuelMess:
There is a possibility that the notification will remain as "Processing..." after a cut has finished.

TranceLove:
Tested on my Oneplus One running Slim7 stable 2.0 (7.1.2).

Unzip: working
Copy a file over SFTP: working
Copy files over SMB: working
Copy a big file (ISO weighing 765MB over SMB, then a video of 830MB over SFTP): operation failed, status bar was also incorrect.

screenshot_20180314-173254

Exception thrown:
```
 java.io.IOException: Map failed
     at sun.nio.ch.FileChannelImpl.map(FileChannelImpl.java:929)
     at com.amaze.filemanager.utils.files.GenericCopyUtil.copyFile(GenericCopyUtil.java:289)
     at com.amaze.filemanager.utils.files.GenericCopyUtil.startCopy(GenericCopyUtil.java:239)
     at com.amaze.filemanager.utils.files.GenericCopyUtil.copy(GenericCopyUtil.java:283)
     at com.amaze.filemanager.asynchronous.services.CopyService$DoInBackground$Copy.copyFiles(CopyService.java:426)
     at com.amaze.filemanager.asynchronous.services.CopyService$DoInBackground$Copy.execute(CopyService.java:326)
     at com.amaze.filemanager.asynchronous.services.CopyService$DoInBackground.doInBackground(CopyService.java:196)
     at com.amaze.filemanager.asynchronous.services.CopyService$DoInBackground.doInBackground(CopyService.java:160)
     at android.os.AsyncTask$2.call(AsyncTask.java:305)
     at java.util.concurrent.FutureTask.run(FutureTask.java:237)
     at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:243)
     at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
     at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
     at java.lang.Thread.run(Thread.java:761)
 Caused by: java.lang.OutOfMemoryError: Map failed
     at sun.nio.ch.FileChannelImpl.map0(Native Method)
     at sun.nio.ch.FileChannelImpl.map(FileChannelImpl.java:926)
 	... 13 more
```
Followed by another exception at GenericCopyUtil
```
D/GenericCopyUtil: I/O Error!
 java.io.IOException
     at com.amaze.filemanager.utils.files.GenericCopyUtil.startCopy(GenericCopyUtil.java:249)
     at com.amaze.filemanager.utils.files.GenericCopyUtil.copy(GenericCopyUtil.java:283)
     at com.amaze.filemanager.asynchronous.services.CopyService$DoInBackground$Copy.copyFiles(CopyService.java:426)
     at com.amaze.filemanager.asynchronous.services.CopyService$DoInBackground$Copy.execute(CopyService.java:326)
     at com.amaze.filemanager.asynchronous.services.CopyService$DoInBackground.doInBackground(CopyService.java:196)
     at com.amaze.filemanager.asynchronous.services.CopyService$DoInBackground.doInBackground(CopyService.java:160)
     at android.os.AsyncTask$2.call(AsyncTask.java:305)
     at java.util.concurrent.FutureTask.run(FutureTask.java:237)
     at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:243)
     at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
     at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
     at java.lang.Thread.run(Thread.java:761)
```
Copy a folder containing the above big ISO over SMB did worked, but the remaining time was displayed incorrectly. I'd accept the progress curve went like below in this case, blame it the network/storage latency

screenshot_20180314-170212

Sorry about the Japanese (since my cell was set to Japanese locale), but I think you should get the picture.

VishalNehra:
@TranceLove that is essentially another issue, not a problem relating to this PR.

app/src/main/java/com/amaze/filemanager/asynchronous/services/ProgressiveServiceAbstract.java
VishalNehra: It helps when we're taking a quick look. Just like 'MainActivity' or 'CopyService'.

app/src/main/java/com/amaze/filemanager/asynchronous/services/ProgressiveServiceAbstract.java
VishalNehra: Yes, I was directly using superclasses' field in subclass by declaring them public and you said opposite to this.

app/src/main/java/com/amaze/filemanager/asynchronous/services/ProgressiveServiceAbstract.java
VishalNehra: Really? This had to come in a review?

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
VishalNehra: Yes, volatile fits the usage here.

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
VishalNehra: Passing service instances within different threads will only lead to complexities, that's why I created a simple interface as the dependency on service was not that major.
Also, the interface is not used in ````ProgressiveService```` but in ````ServiceWatcherUtil````.

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
VishalNehra: Will delete later when I'll be working on a fix to the broken ````ProcessViewerFragment```` when returning to Amaze from progress notification.

app/src/main/java/com/amaze/filemanager/asynchronous/services/ProgressiveServiceAbstract.java
VishalNehra: ````ServiceWatcherUtil```` uses it to interact with the pause/resume of service.

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
VishalNehra: Interface variables are final by default

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
VishalNehra: Yes, will do the needful.

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
VishalNehra: It is not final, but refers to final values and is used with another class too. So it's easy to differentiate from a public method.

app/src/main/java/com/amaze/filemanager/ui/notifications/NotificationConstants.java
```diff
@@ -30,6 +30,7 @@
 
     public static final String CHANNEL_NORMAL_ID = "normalChannel";
     public static final String CHANNEL_FTP_ID = "ftpChannel";
+    public static final int ID_NOTIFICATION_WAIT =  9248;
```
EmmanuelMess: First, "NOTIFICATION" is redundant. Second, "WAIT_ID" already exists, why not use that?

VishalNehra: Will do, but change their name convention later. It's always more generalized to specific not the other way around. For eg. we use ````ID_WAIT```` not ````WAIT_ID````.

VishalNehra:
Fixed issues with processing notification disappearing before actually deleting files from origin.

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
EmmanuelMess: Why not atomic?

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
EmmanuelMess: >  but refers to final values

A lot of variable only have specific states (this is Android so instead of enums, ints are used, as a general example), that doesn't mean they are constant.

> So it's easy to differentiate from a public method.

No one will mistake a public variable for a method :/

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
EmmanuelMess: Ups, didn't know that.

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
EmmanuelMess: > Passing service instances within different threads will only lead to complexities, that's why I created a simple interface as the dependency on service was not that major.

When you "pass an interface" you are passing an instance of a class that implements it, in this case the only class that implements it is ProgressiveService, that is a Service, so you are passing a Service around. 

If you want to only pass a certain operations on the service around (and not all of them) you can make an object that contains the Service but only allows certain functions.

But type erasure is still a thing, and `<T extends Service> T getServiceType();` is ridiculous.

> Also, the interface is not used in ProgressiveService but in ServiceWatcherUtil.

By 'used' I meant "only ProgressiveService implements ServiceWatcherUtil".

app/src/main/java/com/amaze/filemanager/asynchronous/services/ProgressiveServiceAbstract.java
EmmanuelMess: Are you using Android Studio, because it changes the icon for the class. Also, this makes the name even longer. @TranceLove thoughts?

EmmanuelMess: It is passing the `ProgressiveService` as a `ServiceWatcherUtil.ServiceWatcherInteractionInterface`, it can use it directly (if you move the methods).

app/src/main/java/com/amaze/filemanager/asynchronous/services/ProgressiveServiceAbstract.java
EmmanuelMess: Missing an initialization is not breaking enough in both, (it should lead to compile error or instant crash).

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
EmmanuelMess: OK, but next time add a comment explaining that.

app/src/main/java/com/amaze/filemanager/asynchronous/services/ProgressiveServiceAbstract.java
EmmanuelMess: I've seen a LOT of these... Should I make an issue?

app/src/main/java/com/amaze/filemanager/asynchronous/services/ProgressiveServiceAbstract.java
TranceLove: If I'm coding in Java I'd call this `AbstractProgressiveService`; some may call it `BaseProgressiveService` or `ProgressiveServiceBase`... however name it `...Abstract` is a bit awkward to me.

> `ServiceWatcherUtil.ServiceWatcherInteractionInterface`

I'd prefer a shorter name for this. I'm used to use long variable/method names so the code can document itself when everyone are too lazy to write docs (some colleagues hate me doing this ;)); but cases like this doesn't fulfill such argument.

app/src/main/java/com/amaze/filemanager/asynchronous/services/ProgressiveServiceAbstract.java
TranceLove: > I don't think this is a good idea... Subclasses are initializing the superclass' fields, the superclass then uses the fields blindly, it seems to me that there has to be a better way. @TranceLove ?

I tend to use the DRY approach, as long as __I know what I'm doing__ and I know I have initialized the members properly (whenever necessary). Surely it risks difficulty in tracing when code has bugs, but subclassing with declaring fields on our own would also lead to the same result. Also, when other people wants to extend the class, they'd have to walk through a long trail figuring out why the subclass didn't behave as they'd expected.

(OK, with documentation this is not a big problem, but I'd say it's still a bad idea.)

TranceLove:
With today's changeset it's now possible to copy big files that I couldn't copy yesterday, even when it's standalone (not copying with the folder containing it)

Earlier today before 1302af4 and 9e01fc0 arrived I was getting the same pattern at process viewer as the image above; but as I test right now - copying 10 files of sizes from 35-70MB over SFTP - it seems it goes smooth...

screenshot_20180315-232747

Will test once more tomorrow.

VishalNehra:
That's great! The timer, however, seems to be flawed in your screenshot... will look into it.

app/src/main/java/com/amaze/filemanager/asynchronous/services/ProgressiveServiceAbstract.java
EmmanuelMess: The other solution that comes to mind is a lot of abstract getters in superclass, or one or more different objects that can be obtained by the superclass (via abstract getters) and maintain a series of variables.

TranceLove:
Tried again with the same server, same single file, but with SMB and SFTP respectively.

Over SMB
screenshot_20180316-072848
Over SFTP
screenshot_20180316-073900

Then I tried to copy a series of files over SMB and SFTP, and I got this today
screenshot_20180316-105600

Not sure I need to complain this though, as I'd consider the result is not always reproducible, and the network topology at my home and office is entirely different (completely wireless vs ethernet/AP). So just FYI.

> There is a possibility that the notification will remain as "Processing..." after a cut has finished.

Yes, it indeed happened, after I cut and paste the above screenshots back to my desktop over SMB.

VishalNehra:
Processing notification should end after files have successfully been deleted from origin. Please check back to your origin of files were deleted, it not then it's probably waiting for them to be deleted. 
There's another bug that delete task might fail when we're not inside the app during deletion. Which can lead to indefinite processing notification too. But fixing that won't count in this PR, as it would require converting deletion to a service.
There is no problem left with this PR, let's leave further testing to beta testers.

app/src/main/java/com/amaze/filemanager/asynchronous/services/ProgressiveServiceAbstract.java
VishalNehra: Will try to work this around.

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
VishalNehra: I'm not passing a service instance around The interface is just fine, and does the job it's supposed to do.

app/src/main/java/com/amaze/filemanager/asynchronous/services/ProgressiveServiceAbstract.java
VishalNehra: Alright, will rename it to 'AbstractProgressiveService'

VishalNehra: No it can't. There's no need to pass around whole service when you need only some part of it.

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
EmmanuelMess: Ok, pass the object as an instance of the interface, so as to make more difficult unwanted access to other functions.

But the original issue remains: 

> Type erasure makes this [`getServiceType()`] getter unnecessarily complex

In addition `getServiceType()` doesn't actually return a type.

app/src/main/java/com/amaze/filemanager/asynchronous/services/ProgressiveServiceAbstract.java
EmmanuelMess: Ok, pass it as an instance of the interface, so as to make more difficult unwanted access to other functions.

VishalNehra:
Alright, added documentation to initialize variables and renamed the class. I'll be merging this now.

VishalNehra:
Are you positive you tested this right. As I can test, master has this bug not this PR.

app/src/main/java/com/amaze/filemanager/asynchronous/services/ProgressiveServiceAbstract.java
EmmanuelMess: Will PR tomorrow (check the branch [here](https://github.com/EmmanuelMess/AmazeFileManager/tree/fix-vars)) with a solution, it is not the best, but it's late and just fixed a bug where AS would load apk to device.

EmmanuelMess:
> I'll be merging this now.

That wouldn't lead to a better app :disappointed:.

EmmanuelMess:
Copying 133MB file from internal to SD in Android 6.0 got:
```
W/System.err: java.nio.BufferUnderflowException
W/System.err:     at java.nio.Buffer.checkGetBounds(Buffer.java:171)
W/System.err:     at java.nio.DirectByteBuffer.get(DirectByteBuffer.java:113)
                  at java.nio.ByteBuffer.get(ByteBuffer.java:376)
W/System.err:     at com.amaze.filemanager.utils.files.GenericCopyUtil.copyFile(GenericCopyUtil.java:343)
                  at com.amaze.filemanager.utils.files.GenericCopyUtil.startCopy(GenericCopyUtil.java:242)
W/System.err:     at com.amaze.filemanager.utils.files.GenericCopyUtil.copy(GenericCopyUtil.java:283)
                  at com.amaze.filemanager.asynchronous.services.CopyService$DoInBackground$Copy.copyFiles(CopyService.java:457)
                  at com.amaze.filemanager.asynchronous.services.CopyService$DoInBackground$Copy.execute(CopyService.java:357)
                  at com.amaze.filemanager.asynchronous.services.CopyService$DoInBackground.doInBackground(CopyService.java:227)
                  at com.amaze.filemanager.asynchronous.services.CopyService$DoInBackground.doInBackground(CopyService.java:191)
                  at android.os.AsyncTask$2.call(AsyncTask.java:295)
                  at java.util.concurrent.FutureTask.run(FutureTask.java:237)
                  at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:234)
                  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1113)
                  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:588)
                  at java.lang.Thread.run(Thread.java:818)
```

VishalNehra:
@EmmanuelMess please create a new issue for this and assign it to me... This PR doesn't address problems with copying itself but the watcher for it's progress.

VishalNehra:
Anymore plans on adding new code improvement? @EmmanuelMess

EmmanuelMess:
I don't think [this](https://github.com/TeamAmaze/AmazeFileManager/pull/1108#discussion_r175120479) is fixed, want me to PR?

VishalNehra:
Alright go on ahead with a fix for ````getServiceType```` without passing whole service instance.
As for copy problems, I'm not going to test these under this PR. Create new issue and assign it to me. If the copy itself goes wrong I don't think the progress would be to blame if it also shows wrong results.

EmmanuelMess:
> Alright go on ahead with a fix for getServiceType without passing whole service instance.

Will PR ASAP.

VishalNehra:
Approval guys? @EmmanuelMess @TranceLove

EmmanuelMess:
Nope: [this](https://github.com/TeamAmaze/AmazeFileManager/pull/1108#discussion_r174787684) and [this](https://github.com/TeamAmaze/AmazeFileManager/pull/1108#discussion_r174786502).

app/src/main/java/com/amaze/filemanager/utils/ServiceWatcherUtil.java
VishalNehra: Methods are atomic, not variables. Variables are volatile. As for using wrapper classes of Atomic behaviour, I don't see the need.

VishalNehra:
You can change the name of that variable later.

EmmanuelMess:
A. That only addresses one issue and
B. I could have PR'd a fix from the start, why not?

VishalNehra:
I addressed your another question in it's respective comment.
Please don't hold important PRs like this for silly things. It's pushing update schedules.

TranceLove:
> Approval guys?

Functional-wise I see no big problem here after testing, and there are other PRs that depend on this. I'll agree push this through and let code quality problems raised be addressed later, so other problems can be worked on.

Voting +1.

EmmanuelMess:
Okay, I'll move the Atomic vs volatile discussion to an issue.

PR #1120, opened 2018-03-28
Allow specifying filename on encrypting files

TranceLove:
Insisted as last task for #1020. Sorry for working on this so late - hope I understand the requirements correctly.

Main changes happened on aa7032d while other commits were to make `fix-warnoninput` keep up with `master`. Also added size check on pending intents in `ServiceWatcherUtil` to prevent app crash from NPE when `ServiceWatcherUtil` tries to consume Intents in the queue.


VishalNehra:
Why so many commits :/ it's difficult to find your specific changes.
BTW in commit ````aa7032d```` why did you create a new separate dialog for file names. You could just take filename in the same dialog as password one.
Also, please revert changes made in ServiceWatcherUtil, as they'll conflict with PR #1108

TranceLove:
> BTW in commit aa7032d why did you create a new separate dialog for file names. You could just take filename in the same dialog as password one.
Also, please revert changes made in ServiceWatcherUtil, as they'll conflict with PR #1108

Good idea. Will do.

Maybe I should prepare a more clean branch before submitting the PR again... closing first, in favour of new, refreshed branch and work.

VishalNehra:
That'd be great! Appreciate your work on this though.

PR #1121, opened 2018-03-28
Merge current master into fix-warnoninput

TranceLove:
No actual changes were done but to let `fix-warnoninput` to catch up with current `master`, so future changes can be tracked more easily.

Seems there was no easy way out but need to use `git merge` instead of `git rebase`.


TranceLove:
CI build failed was caused by failure in resolving dependencies. Couldn't get the build log now, but IIRC it was leakcanary

`com.squareup.leakcanary:leakcanary-android:1.5.3`

The jar is on maven's repo, but gradle instead tried to download from bintray. Not sure if it's now fixed, but it was what happened during CI check.

Merging anyway.

PR #1126, opened 2018-03-30
Allow specifying name of encrypted file/directory (Rework)

TranceLove:
Insisted as task on #1020. Also migrated the dialog to use constraint layout.


TranceLove:
> Is there a dialog to change the encryped file's name when encrypting? Because I am getting only the password dialog.

OK, missed out the path of encrypting files using master password/fingerprint... will work on it.

VishalNehra:
Encryption is broken on master, but is fixed with #1108 
Code seems ok, but can only successfully test after above PR is merged.

EmmanuelMess:
Tested, works on my Moto E2.

PR #1134, opened 2018-03-31
refactor UtilsHandler to centralize saving to database and fix #1128

### LINKED ISSUE #1128: Adding folder to bookmarks make app crashes
## **Expected behavior**
adding folder to bookmark show toast the app not crash

## **Actual behavior**
adding folder to bookmark show toast and app crashes

### How to reproduce

- Launch Amaze filemanager
- Press on 3 dots to show menu of actions 
- Press on Add to Bookmark

### TESTING ENVIRONMENT
Operating system: ANDROID V7.0 
App Version: v3.2.1

michaelsam94:
Fix #1128


EmmanuelMess:
You created an empty interface `OperationData`, and then implemented it in different `*OperationData`. Wouldn't it be better to have one immutable OperationData with an int indicating which type of operation it should be. Then extend that class for `SMPOperationData`, `BookmarksOperationData` and `SSHOperationData` (I know immutables are generally final, but let's make an exception, or use an abstract class and three extensions).

michaelsam94:
@EmmanuelMess yes I agree with you it's better to reduce number classes and casts I will make this change and push it.

michaelsam94:
I will make abstract class called operationdate and include path as field to be overided in subclasses but still need to extend 7 types because every type has path

EmmanuelMess:
You don't need 7 subclases, if the code is the same just add an int to know which one of the bunch it is, and go from there. `GridOperationData`, `HiddenOperationData`, `HistoryOperationData` and `ListOperationData` have the same code.

michaelsam94:
@EmmanuelMess i make change delete four classes and create one class instead but I need more clarification about immutable part can you clarify to me more ?

EmmanuelMess:
[The Java™ Tutorials: A Strategy for Defining Immutable Objects](https://docs.oracle.com/javase/tutorial/essential/concurrency/imstrat.html)

michaelsam94:
yes I know mutability but here what objects need to be immutable OperatoinData is interface

michaelsam94:
okay enum can be replaced with constants I will define 4 constants in UtilsHandler for each type

michaelsam94:
if i keep enum I can remove switch statement and call setPath directly

EmmanuelMess:
> yes I know mutability but here what objects need to be immutable OperatoinData is interface

It's subclasses.

> if i keep enum I can remove switch statement and call setPath directly

Don't keep enum, add `setPath()` to interface, all subclasses seem to have it.

EmmanuelMess:
You forgot to finalize the fields of immutable classes, then you can remove the getters. And please squash the commits into one.

michaelsam94:
But removing getters require the fields to be public to access them in utilshandler

EmmanuelMess:
Make them public.

michaelsam94:
there no problem @EmmanuelMes as long we care about quality of code and to make it easy in maintaining later but the method in interface getPath keep it or remove I see there will not be consistent with access to fields directly

EmmanuelMess:
>  the method in interface getPath keep it or remove I see this will not be consistent with access to fields directly

I think you missed a word in that sentence. Remove the getters, use public final fields.

EmmanuelMess:
I think it'll be better to remove it, I don't like to end up with an empty interface, but it'll be better to remove it, I think.

EmmanuelMess:
Give me a few hours I will PR some fixes into you branch

michaelsam94:
okay still there is an issue regarding to adding duplicate bookmarks

michaelsam94:
I try to squash commits on sourcetree but it's seems I don't it right

EmmanuelMess:
Ok, after restart bookmarks are broken, path appears instead of name and clicking crashes app.

michaelsam94:
ok I will fix it and make PR again

EmmanuelMess:
I will close this as you will PR again. I can reopen later if you fix the issues in this branch.

PR #1140, opened 2018-04-04
Corrected FAB colors on OLED theme

EmmanuelMess:
* Changed titles background to black
* Added black overlay (9 patch)


TranceLove:
Looks clean, though don't know how to comment on the change since I don't have an OLED device on hand... but the unified black background does looks good. Will vote +1.

Before
screenshot_1522852323
After
screenshot_1522852625

PR #1141, opened 2018-04-04
Fix drawer width

VishalNehra:
Just now noticed drawer is too wide right now.
As per [material design guidelines](https://material.io/guidelines/layout/metrics-keylines.html#metrics-keylines-keylines-spacing) the navigation drawer right margin should be equal to toolbar height.
For us, toolbar height is 56 + 72dp (additional height)

Compare Screenshots for yourself and see, current one is much wider, covers up most of the screen and even the header is not in correct aspect ratio of 16:9, it's too much stretched.

screenshot_amaze_20180405-023718
screenshot_amaze_20180405-023609


EmmanuelMess:
I am very pedantic, so I will merge bug, for the record:

I think you found a bug in the documentation:

> Side nav width: The screen width minus the height of the **action bar**. Here, the width is 56dp from the right screen edge.

-- [Material design guidelines](https://material.io/guidelines/layout/metrics-keylines.html#metrics-keylines-keylines-spacing) (bold added).

[Action bar](https://developer.android.com/design/patterns/actionbar.html) is not [toolbar](https://developer.android.com/reference/android/widget/Toolbar.html) is not [appbar](https://developer.android.com/training/appbar/index.html). 

See [here](https://developer.android.com/design/patterns/actionbar.html#considerations-split-action-bars):

> Layout Considerations for Split Action Bars
> 
> When splitting up content across multiple action bars, you generally have three possible locations for action bar content:
> 
>     1. Main action bar
>     2. Top bar
>     3. Bottom bar
> 
> If the user can navigate up the hierarchy from a given screen, the main action bar contains the up caret, at a minimum.
> 
> To allow the user to quickly switch between the views your app provides, use tabs or a spinner in the top bar.
> 
> To display actions and, if necessary, the action overflow, use the bottom bar.
> 

The 3 parts are part of the action bar. Leading to a possible ridiculous reduction in drawer size.

VishalNehra:
Google themselves doesn't seem to follow the same.

EmmanuelMess:
Google never followed anything.

PR #1145, opened 2018-04-05
Catch ActivityNotFoundException and launch web browser for app

### LINKED ISSUE #1139: Crashes if Play store not installed
App Manager -> [more] popup -> Play store
It crashes by ActivityNotFoundException

TranceLove:
Fixes #1139, app package's webpage on Google Play Store (if any) will be opened if user device does not have Google Play Store installed, which should be sufficient as almost all Android devices should have webview implementations.

Additionally
 * remove deprecated MaterialDialog.Builder.callback() to .onNegative() and .onPositive()
 * change HybridFileParcelable.getParent() to HybridFileParcelable.getParent(Context).
 * change hard-coded SDK version 21 to Build.VERSION_CODES.LOLLIPOP

Tested with Nougat emulator (no Google Play Services installed).

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/adapters/AppsAdapter.java
+++ b/app/src/main/java/com/amaze/filemanager/adapters/AppsAdapter.java
@@ -20,6 +20,7 @@
 package com.amaze.filemanager.adapters;
 
 import android.app.Activity;
+import android.content.ActivityNotFoundException;
 import android.content.Context;
 import android.content.Intent;
 import android.content.pm.ApplicationInfo;
@@ -212,8 +213,13 @@ public class AppsAdapter extends ArrayAdapter<AppDataParcelable> {
                         return true;
                     case R.id.play:
                         Intent intent1 = new Intent(Intent.ACTION_VIEW);
-                        intent1.setData(Uri.parse(String.format("market://details?id=%s", rowItem.packageName)));
-                        app.startActivity(intent1);
+                        try {
+                            intent1.setData(Uri.parse(String.format("market://details?id=%s", rowItem.packageName)));
+                            app.startActivity(intent1);
+                        } catch (ActivityNotFoundException ifPlayStoreNotInstalled) {
+                            intent1.setData(Uri.parse(String.format("https://play.google.com/store/apps/details?id=%s", rowItem.packageName)));
+                            app.startActivity(intent1);
+                        }
                         return true;
                     case R.id.properties:
                         app.startActivity(new Intent(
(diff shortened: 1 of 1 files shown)


EmmanuelMess:
Can't test right now, but shouldn't open in Play Store be removed if it is not installed?

TranceLove:
> Can't test right now, but shouldn't open in Play Store be removed if it is not installed?

It's indeed one of the alternatives... see 2bacd0a, but I'm open to both options.

app/src/main/java/com/amaze/filemanager/adapters/AppsAdapter.java
```diff
@@ -249,9 +242,23 @@ public void onPositive(MaterialDialog materialDialog) {
...
             popupMenu.show();
         });
 
     }
 
+    private final boolean isGooglePlayStoreAvailable() {
+        PackageManager pm = context.getPackageManager();
+        try {
+            pm.getPackageInfo(COM_ANDROID_VENDING, PackageManager.GET_ACTIVITIES);
+            return true;
```
EmmanuelMess: When I test on emulator, the app crashes as before. I think the package is present but not the activity. Changing these lines to:
```java
    PackageInfo packageInfo = pm.getPackageInfo(COM_ANDROID_VENDING, PackageManager.GET_ACTIVITIES);
    return packageInfo.activities != null;
```
Seems to fix it.

app/src/main/java/com/amaze/filemanager/adapters/AppsAdapter.java
```diff
@@ -249,9 +242,23 @@ public void onPositive(MaterialDialog materialDialog) {
...
+        } catch (PackageManager.NameNotFoundException ifNotInstalled) {
+            ifNotInstalled.printStackTrace();
```
EmmanuelMess: I think this just clutters the logcat, when printed it doesn't help, hence we shouldn't print it.

app/src/main/java/com/amaze/filemanager/adapters/AppsAdapter.java
TranceLove: Being a cautious one I'd check both.

app/src/main/java/com/amaze/filemanager/adapters/AppsAdapter.java
```diff
@@ -253,10 +254,9 @@ private void showPopup(View v, final AppDataParcelable rowItem){
     private final boolean isGooglePlayStoreAvailable() {
         PackageManager pm = context.getPackageManager();
         try {
-            pm.getPackageInfo(COM_ANDROID_VENDING, PackageManager.GET_ACTIVITIES);
-            return true;
+            PackageInfo appInfo = pm.getPackageInfo(COM_ANDROID_VENDING, PackageManager.GET_ACTIVITIES);
+            return appInfo != null || appInfo.activities != null;
```
EmmanuelMess: Or? If `appInfo` is null `appInfo.activities` will NPE.

TranceLove: Oops.

VishalNehra:
I think opening in browser is a good option in case play store is not installed.
A fix would be to not show this option in case the application is not signed with a playstore key (example system apps or debug/third party apps). However it doesn't concern this PR.

EmmanuelMess:
> I think opening in browser is a good option in case play store is not installed.
> A fix would be to not show this option in case the application is not signed with a playstore key (example system apps or debug/third party apps). However it doesn't concern this PR.

This should go to an issue.

VishalNehra:
Yes, but don't hide the option in this PR, open it with browser.

PR #1166, opened 2018-04-12
OLED: Black appbar overflow, ActionBar and BottomBar (when ActionBar is active)

EmmanuelMess:
* Renamed styles overflow_white and overflow_white2 to overflow_light a…
* Black appbar overflow and action bar
* Black BottomBar (when ActionBar is active)


EmmanuelMess:
[ActionBar](https://developer.android.com/design/patterns/actionbar.html)? I can't replicate, don't you mean [Navigation Bar](https://developer.android.com/training/system-ui/navigation.html)?

TranceLove:
> ActionBar? I can't replicate, don't you mean Navigation Bar?

Oops, my bad.

VishalNehra:
Pitch black action bar it's matching up the UI. We can't differentiate between the bounds of action bar to where the list actually starts as in other theme.

EmmanuelMess:
You say that from a purely aesthetic standpoint or also functional?

VishalNehra:
I don't think this hinders any function. Aesthetically it's not looking good. And like i said, differs from other themes where we can actually see the action mode.

EmmanuelMess:
I see your point, I'll remove that commit, what about the overflow?

EmmanuelMess:
Do test now, I've removed the other modification.

VishalNehra:
I'm still getting pitch black appbar

EmmanuelMess:
Please delete your copy of the branch and retry

EmmanuelMess:
@TranceLove ?

TranceLove:
I'm fine with it. Voting +1.

PR #1202, opened 2018-04-29
Sollution Issue 1176

### LINKED ISSUE #1176: ActionBar's color is broken in API 15

beastlyCoder:
I was able to add the correct styles code to API Level 15.

EDIT: fixes #1176. - By @TranceLove

### DIFF
--- a/app/src/main/res/values-v15/styles.xml
+++ b/app/src/main/res/values-v15/styles.xml
@@ -1,0 +1,56 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <resources>
+
+        <!-- appCompat -->
+        <!-- light -->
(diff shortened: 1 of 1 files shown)


app/src/main/res/values-v15/styles.xml
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <resources>
```
TranceLove: Actually I doubt if you had test the changes in Android Studio, since I also couldn't get the build to run with my Android Studio after applying this change.

To begin with, remove this redundant `<resources>` tag for the build to pass.

And of course, the closing tag too.

EmmanuelMess:
Why is this is values-v15? I think errors in API 15 are frequently replicable in API 14, but I can't test the latter. This should be in the base values.xml.

PR #1207, opened 2018-04-30
Converted copy dialog layout to ConstraintLayout and replaces endline with two TextViews

EmmanuelMess:
Converted copy dialog layout to ConstraintLayout and replaces endline with two TextViews.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -154,6 +154,7 @@ dependencies {
     }
 
     implementation 'com.android.support:multidex:1.0.2'//Multiple dex files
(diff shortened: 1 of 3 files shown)


VishalNehra:
Sadly, my phone is cooked, bad battery and damaged daughterboard. Will take a few days to set it up again. Will try setting up emulator in the mean time.

VishalNehra:
Got my device working. 
Looks nice. Also, while I was testing this, found out that 'do for all items' option is broken. Basically just shows one duplicate file even when we've many, and regardless of checking above mentioned option, it applies the selected operation to all the files. This may go to a new issue.

PR #1201, opened 2018-04-28
Merge fix-warnoninput to master

TranceLove:
Fixes #1020.

Also updated file license header for the files touched per #985 recommended.

PR #1209, opened 2018-05-02
Add support for save files to external storage via ACTION_SEND intent

### LINKED ISSUE #1073: Cannot save file to SD card
Cannot save file to SD card via share to Amaze. It says "Operation not allowed".
The problem seems to be here:
https://github.com/TeamAmaze/AmazeFileManager/blob/e928b4c2ac14befa11ab04821533c7f863dc884e/app/src/main/java/com/amaze/filemanager/filesystem/FileUtil.java#L643-L666
`boolean result = file.canWrite();`
Returns false, incorrectly.

TranceLove:
Fixes #1073.

Note: this won't fix sharing some selected text to device via Amaze - can't find the exact issue #, that will be fixed in separate PR once #1201 is merged. However on such intents which has no Uri but data content this PR should prevent Amaze from crashing.

Test case for verification:
1. Open a web page with images with Firefox/Chrome
2. Long tap an image
3. Select Amaze "Save As..."
4. Amaze FM pops up, FAB menu button shows a download symbol
5. Select external storage via the drawer or enter path
6. On devices running Android versions that have SAF enabled, dialog prompting user to select external device's directory will be displayed
7. Select the external storage device's directory and the file will be saved there

Not easy to write test case, but did tested on Oneplus One running Slim7 (7.1.2), Nougat emulator (Pixel 2 with external SD card), Galaxy S2 running SlimLP (5.1.1) with 64GB micro SD and GPD XD running LegacyROM (4.4.4) with 64GB micro SD.

One thing that concerned me was `MainActivityHelper.checkFolder()` was called twice, with only the second call would keep the SAF prompt dialog on the screen - so feel free to improve this.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/filesystem/FileUtil.java
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/FileUtil.java
@@ -35,6 +35,7 @@ import android.provider.MediaStore;
 import android.support.annotation.NonNull;
 import android.support.v4.provider.DocumentFile;
 import android.util.Log;
+import android.widget.Toast;
 
 import com.amaze.filemanager.R;
 import com.amaze.filemanager.activities.MainActivity;
@@ -196,7 +197,9 @@ public abstract class FileUtil {
         AppConfig.runInBackground(new AppConfig.CustomAsyncCallbacks() {
 
             @Override
-            public <E> E doInBackground() {
+            public List<String> doInBackground() {
+
+                List<String> retval = new ArrayList<>();
 
                 for (Uri uri : uris) {
 
@@ -249,6 +252,7 @@ public abstract class FileUtil {
                                 }
 
                                 bufferedOutputStream = new BufferedOutputStream(contentResolver.openOutputStream(targetDocumentFile.getUri()));
+                                retval.add(targetFile.getPath());
                                 break;
                             case SMB:
                                 SmbFile targetSmbFile = new SmbFile(finalFilePath);
@@ -258,6 +262,7 @@ public abstract class FileUtil {
                                 } else {
                                     OutputStream outputStream = targetSmbFile.getOutputStream();
                                     bufferedOutputStream = new BufferedOutputStream(outputStream);
+                                    retval.add(mainActivity.mainActivityHelper.parseSmbPath(targetSmbFile.getPath()));
                                 }
                                 break;
                             case SFTP:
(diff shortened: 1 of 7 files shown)


EmmanuelMess:
If I try to save an entire page the FAB disappears, not sure if this happens on master. When sharing an image, after testing the entire page share, the FAB is not there; after Amaze restart it reappears and I can share image normally. After reopening amaze from recents, the download FAB is still there, normal after restart.

TranceLove:
OK, will look into it.

VishalNehra:
@EmmanuelMess I don't think this PR should address these issues you mentioned. 
That being said, I can't test this or any other PR as my device is still a paper weight. Will try to get it working by today.

TranceLove:
The symptom reported by @EmmanuelMess seems reproducible on my Oneplus One running Slim7 (7.1.2), when the intent was sent to a running instance of the app.

TranceLove:
Added 69ca334 to make sure the FAB menu is visible when ACTION_SEND intent was sent when Amaze is running (at the background).

However a quick test on my Oneplus One running Slim7 (7.1.2) just before I pushed this had an regression that the saved file shared from Firefox was saved as `null`.

Please refrain from merging until `do not merge` label is removed.

EmmanuelMess:
@VishalNehra Almost all of those bugs could well be in master, but I was to lazy to check.

TranceLove:
Problem should be in `master` too since I also hit the same problem when I was on other branch earlier today. The two lines in 69ca334 however should be safe to become a PR itself if it's necessary to fix in `release-3.3.0`.

EmmanuelMess:
Saving from Reddit Boost I get "Cannot overwrite file" and I end up with 0 byte image.

TranceLove:
> Saving from Reddit Boost I get "Cannot overwrite file" and I end up with 0 byte image.

Maybe caused by FileUtil.getDocumentFile() that created a file stub at the first place. Either drop it before open the OutputStream, or allow overwrite the 0 byte file. 0ffdc07 is using the latter option.

EmmanuelMess:
The only issue now would be to add a Toast or Notification telling the user the file was downloaded.
Also, this fixes a current bug, why milestone 3.4.0?

TranceLove:
> Also, this fixes a current bug, why milestone 3.4.0?

Can be regarded as newly supported feature too. Won't mind push it up to v3.3.0 if @VishalNehra also approved and everyone agree to.

(OK, after I work on the Toast first.)

app/src/main/java/com/amaze/filemanager/filesystem/FileUtil.java
```diff
@@ -334,11 +349,19 @@ public static final void writeUriToStorage(@NonNull final MainActivity mainActiv
...
                     }
                 }
-                return null;
+                return (retval.size() > 0) ? retval : null;
             }
 
             @Override
             public Void onPostExecute(Object result) {
+                if(result !=  null) {
+                    List<String> paths = (List<String>) result;
```
EmmanuelMess: Why are you casting `result`, can't it be of type `List<String>`.

app/src/main/java/com/amaze/filemanager/filesystem/FileUtil.java
```diff
@@ -334,11 +349,19 @@ public static final void writeUriToStorage(@NonNull final MainActivity mainActiv
...
+                    if (paths.size() == 1) {
```
EmmanuelMess: Why are you returning a `List` and only using the first element?

app/src/main/java/com/amaze/filemanager/filesystem/FileUtil.java
TranceLove: Implementation at `AppConfig.runInBackground()` is a bit messy that I couldn't find a way to add the type erasures back - for the time being.

app/src/main/java/com/amaze/filemanager/filesystem/FileUtil.java
TranceLove: It is possible that more than one file can be saved per `FileUtil.writeUriToStorage()` since it accepts `List<Uri>`.

If more than one `Uri` can be saved, I'll want to make the Toast pops up as
 - if one `Uri` is saved, display the exact path of the location of the file saved
 - if more than one `Uri` are saved, simply notify user how many files are saved instead of popping up multiple Toasts

VishalNehra:
If this is working well we can merge it for milestone v3.30

EmmanuelMess:
> If this is working well we can merge it for milestone v3.30

I'll move milestone.

PR #1226, opened 2018-05-11
UtilsHandler method refractorization

EmmanuelMess:
* Refactor UtilsHandler to centralize saving to database via OperationData
* Removed UtilsHandler.add*() in favor of saveToDb()
* Fixed bookmark name and path getting switched
* Replaced UtilsHandler.delete*() with removeFromDatabase(OperationData)
* Removed HandlerUtils.clear*Table() and made public clearTable(Operation) 

PS: later we can put some of these methods in Operations enum.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/database/UtilsHandler.java
+++ b/app/src/main/java/com/amaze/filemanager/database/UtilsHandler.java
@@ -151,6 +151,30 @@ public class UtilsHandler extends SQLiteOpenHelper {
         });
     }
 
+    public void removeFromDatabase(OperationData operationData) {
+        AppConfig.runInBackground(() -> {
+            switch (operationData.type) {
+                case HIDDEN:
+                case HISTORY:
+                case LIST:
+                case GRID:
+                    removePath(operationData.type, operationData.path);
+                    break;
+                case BOOKMARKS:
+                    removeBookmarksPath(operationData.name, operationData.path);
+                    break;
+                case SMB:
+                    removeSmbPath(operationData.name, operationData.path);
+                    break;
+                case SFTP:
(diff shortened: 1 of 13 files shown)


app/src/main/java/com/amaze/filemanager/database/UtilsHandler.java
```diff
@@ -126,6 +128,53 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
...
+                case LIST:
+                case GRID:
+                    setPath(operationData.type, operationData.path);
+                    break;
+                case BOOKMARKS:
+                case SMB:
+                    setPath(operationData.type, operationData.name, operationData.path);
+                    break;
+                case SFTP:
+                    addSsh(operationData.name, operationData.path, operationData.hostKey,
```
TranceLove: `addSsh()` with private key included would crash app on creating new SCP/SFTP connection because the async save-to-database operation didn't give enough time for LoadFilesListTask to login with key authentication ([SftpConnectDialog line 331](https://github.com/EmmanuelMess/AmazeFileManager/blob/a6fb7df4ca2af66aa8a55d8f1783354d432d243a/app/src/main/java/com/amaze/filemanager/ui/dialogs/SftpConnectDialog.java#L331)).

Verified on Oneplus One running Slim7 (7.1.2).

Must confess it was a flaw that I didn't resolve when I worked on #863. I'll see if I can work around this... most likely with an optional callback?

EmmanuelMess: >  because the async save-to-database operation didn't give enough time for LoadFilesListTask to login with key authentication

This is by design? If so it should be fixed by you ASAP.

PR #1233, opened 2018-05-16
Allow create file with filename ".txt"

### LINKED ISSUE #1231: Can't create ".txt" file
Can't create ".txt" file with dialog, you can create "t.txt" and "txt" though.

### LINKED ISSUE #1235: Creating files/folders with names starting with dot don't show "hidden file" warning
Create a file named ".a.txt", if the "show hidden files" configuration isn't on the file will dissapear without explanation. We should show a warning when file or folder names start with a dot.

TranceLove:
Fixes #1231 and fixes #1235.

Trigger WarnableTextInputValidator.afterTextChanged() if prefill is not empty so warning message can be displayed too.


TranceLove:
> It works, but the warning makes no sense, one could argue that the extension is ".txt" and the name is "" but maybe the warning should read something along the lines of "you are creating a hidden file" (will create issue about this).

Let's see if a792fce can help.

TranceLove:
Two ticks, conflicts resolved and finished CI build with green. Merging.

PR #1245, opened 2018-05-21
Little help on enabling testing on MainActivity

TranceLove:
Tweaks to enable writing test cases against MainActivity (or classes that calls it).

* Added shadows-httpclient to prevent ClassNotFound error
* MainActivity.getStorageDirectories(): Check for existence of `/storage/sdcard0` before actually add to list of storage directories, to prevent test crash because of MainActivity failing to create tabs

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/activities/MainActivity.java
+++ b/app/src/main/java/com/amaze/filemanager/activities/MainActivity.java
@@ -576,10 +576,12 @@ public class MainActivity extends ThemedActivity implements OnRequestPermissions
             if (TextUtils.isEmpty(rawExternalStorage)) {
                 // EXTERNAL_STORAGE undefined; falling back to default.
                 // Check for actual existence of the directory before adding to list
-                if(new File(DEFAULT_FALLBACK_STORAGE_PATH).exists())
+                if(new File(DEFAULT_FALLBACK_STORAGE_PATH).exists()) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/amaze/filemanager/activities/MainActivity.java
```diff
@@ -573,7 +575,11 @@ private void initializeInteractiveShell() {
             // Device has physical external storage; use plain paths.
             if (TextUtils.isEmpty(rawExternalStorage)) {
                 // EXTERNAL_STORAGE undefined; falling back to default.
-                rv.add("/storage/sdcard0");
+                // Check for actual existence of the directory before adding to list
+                if(new File(DEFAULT_FALLBACK_STORAGE_PATH).exists())
```
EmmanuelMess: Please use convention:
```
if(<cond>) {
    ...
}
```

TranceLove:
Two ticks, merge.

EmmanuelMess:
Do I add this to release 3.3.0 branch?

TranceLove:
> Do I add this to release 3.3.0 branch?

I think it shouldn't be a problem, though this doesn't fix anything for 3.3.0 but a mere helper for writing test cases.

EmmanuelMess:
@TranceLove 
> * MainActivity.getStorageDirectories(): Check for existence of /storage/sdcard0 before actually add to list of storage directories, to prevent test crash because of MainActivity failing to create tabs

PR #1252, opened 2018-05-25
Ask for permission before installing apk on API >= M

### LINKED ISSUE #1227: Can't install apps on Oreo
Since we don't explicitly ask for install permissions using the [new API](https://developer.android.com/studio/publish/#publishing-unknown) Amaze fails to install any apk file.

EmmanuelMess:
* Ask for permission before installing apk on API >= M
* Created PermissionsActivity for MainActivity

Fixes #1227.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/activities/superclasses/PermissionsActivity.java
+++ b/app/src/main/java/com/amaze/filemanager/activities/superclasses/PermissionsActivity.java
@@ -1,0 +1,113 @@
+package com.amaze.filemanager.activities.superclasses;
+
+import android.Manifest;
+import android.content.pm.PackageManager;
+import android.os.Build;
+import android.os.Bundle;
+import android.support.annotation.NonNull;
+import android.support.annotation.RequiresApi;
+import android.support.v4.app.ActivityCompat;
+import android.widget.Toast;
+
+import com.afollestad.materialdialogs.DialogAction;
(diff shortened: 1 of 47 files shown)


app/src/main/java/com/amaze/filemanager/activities/superclasses/PermissionsActivity.java
```diff
@@ -0,0 +1,113 @@
+package com.amaze.filemanager.activities.superclasses;
```
TranceLove: It sounds a bit awkward to me that a superclass is placed below the subclass. However, it makes sense if it is heading for 3.3.0 since the blast radius will be too big if `MainActivity` should be placed below `PermissionsActivity`.

EmmanuelMess: What do you mean "below"? Physically below in some visual representation or hierarchically below?

TranceLove: Hierarchically below.

EmmanuelMess: `MainActivity` extends `PermissionsActivity`, `MainActivity` is hierarchically below `PermissionsActivity`.

PR #1228, opened 2018-05-12
Fix glide iconloading

VishalNehra:
Fixes image loading for cloud 
Also fixes GIF loading
Fixes #893 
Fixes #550

PR #1257, opened 2018-05-29
Add support for hash code of SFTP files in file lists

TranceLove:
Addendum to #1228, add support of calculating hash code for `RemoteResourceInfo` the representation of files stored on SSH server which [sshj](https://github.com/hierynomus/sshj) uses.

Hash code calculation method is derived from [here](https://stackoverflow.com/a/113600/768633).

Unit test included.


VishalNehra:
Will be closing this as there's some problem with glide not identifying encoder correctly if we use hashes as keys for remote connection. 
Will create a new issue for the same.

TranceLove:
Understood. Closing.

PR #1258, opened 2018-05-30
Fix possibility of archive entry being extracted to outside extract destination

TranceLove:
Details about the vulnerability: [here](https://snyk.io/research/zip-slip-vulnerability); fix for 3.2.1 is #1260 which had been released to Google Play Store as version 3.2.2.

Also brought the test suite from e0d40191e in #1222 for behavioural verification.

IOException will be thrown on erroneous archive entry, causing app to pop up an "Error" Toast.

Safe to cherry-pick into `release-3.3.0`.

Our sincere thanks to folks at [snyk.io](https://snyk.io) who tipped us about the vulnerability.

PR #1260, opened 2018-05-30
Fix possibility of archive entry extracted outside specified destination for 3.2.1

TranceLove:
Backported from #1258.

Copied from [comment](https://github.com/TeamAmaze/AmazeFileManager/pull/1258#issuecomment-393011994):

> Since 3.2.1 codebase doesn't in favour of writing unit tests, verification was done against Pixel 2 emulators running 5.1.1, 6.0, 7.1.1, 8.1, plus Oneplus One running Slim7 (7.1.2), Galaxy S2 running SlimLP (5.1.1) and GPD XD running LegacyROM (4.4.4).

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/services/ExtractService.java
+++ b/app/src/main/java/com/amaze/filemanager/services/ExtractService.java
@@ -231,13 +231,17 @@ public class ExtractService extends Service {
         private void unzipEntry(ZipFile zipFile, ZipEntry entry, String outputDir)
                 throws Exception {
 
-            if (entry.isDirectory()) {
-                // zip entry is a directory, return after creating new directory
-                createDir(new File(outputDir, entry.getName()));
-                return;
+            final File outputFile = new File(outputDir, fixEntryName(entry.getName()));
+
+            if (!outputFile.getCanonicalPath().startsWith(outputDir)){
+                throw new IOException("Incorrect ZipEntry path!");
             }
 
-            final File outputFile = new File(outputDir, entry.getName());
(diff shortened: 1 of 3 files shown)


EmmanuelMess:
This is a fix, but after trying to open evil.txt, can't open good.txt.

EmmanuelMess:
Ok, nevermind, there's a folders I mistaked for the "back" button, in the non win file. It just keeps going "up" "/" -> "/../" -> "/../../".

EmmanuelMess:
@VishalNehra this should have been released yesterday.

VishalNehra:
I think we can finalize v3.3.0 soon enough so this won't need to be released with the older one.

EmmanuelMess:
I think that is a really bad idea, our app is vulnerable to a known vulnerability right now, not after 3.3.0 is ready.

TranceLove:
> I think that is a really bad idea, our app is vulnerable to a known vulnerability right now, not after 3.3.0 is ready.

+1. I think we're still miles away from 3.3.0 for the time being. It will be a good idea to patch current users instead of let them wait and worry about the vulnerability.

VishalNehra:
In that case please go ahead and build the release and push on Play Console.

EmmanuelMess:
> In that case please go ahead and build the release and push on Play Console.

You have to authenticate.

You also made me realize that we may not be able to release to FDroid unless the master branch has a new tag with the fix.

EmmanuelMess:
> In that case please go ahead and build the release and push on Play Console.

I'll take this as approval of this PR.

Issue #528, opened 2017-04-17
Per-Folder Sort Settings


jimbo1qaz:
Currently I cannot set the sorting order per-folder. For example, I want Download to be newest-first, folders last, but /sdcard to be sorted by name, folders first. Is it possible to implement such a feature?

vishal0071:
It's absolutely possible, but doing so will probably lead to slow UX. Like we would need to save these preferences in database, map them with path. Now searching each entry in database for a specific path is pretty heavy operation. List loading will be heavily effected even for those paths on which there is no sort option selected.

jimbo1qaz:
Well this feature is present on Windows. I personally consider it a fundamental basic feature, and am surprised at how no Android file manager has it. Is it possible to configure a default folder sort, with a handful of overrides (eg downloads folder)? Windows stores in per-folder desktop.ini, would a central database be appropriate for Amaze?

EmmanuelMess:
Load every folder that has ordering on app start into a trie, check routes against that trie, then load the preference from a preloaded map, I don't think it really is so time consuming. You could improve by not checking subdirectories of a dir that doesn't exist (substring check in trie).

MaanooAk:
I'm willing to try implement this feature (when I will have some time) and make it so it can be disabled from the settings in order to avoid the overhead.
@vishal0071 (or anyone else) If you have planning implementing it yourself or you are opposed to it just let me know

vishal0071:
@MaanooAk I had not planned yet. Go ahead (:

jerryn70:
+1 for this. This is a must need feature for me also.
@MaanooAk is there any progress in your work ?

MaanooAk:
@jerryn70 no (cough up in other project)
I did some tests on storing sorting per folder, nothing really good came out of it. I will post here if I'll start again

tamalbuk:
@jimbo1qaz Solid Explorer has per folder sorting options.

PR #1262, opened 2018-05-31
Add sort option: Only this folder

ningshu:
Fixes [#528](https://github.com/TeamAmaze/AmazeFileManager/issues/528)

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/ui/dialogs/GeneralDialogCreation.java
+++ b/app/src/main/java/com/amaze/filemanager/ui/dialogs/GeneralDialogCreation.java
@@ -97,6 +97,7 @@ import java.io.File;
 import java.io.IOException;
 import java.security.GeneralSecurityException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -691,7 +692,7 @@ public class GeneralDialogCreation {
         textInputLayoutEncryptSaveAs.setHint(intentParcelable.isDirectory() ?
                 c.getString(R.string.encrypt_folder_save_as) :
                 c.getString(R.string.encrypt_file_save_as));
-        
+
         passwordEditText.post(() -> {
             InputMethodManager imm = (InputMethodManager) main.getSystemService(Context.INPUT_METHOD_SERVICE);
             imm.showSoftInput(passwordEditText, InputMethodManager.SHOW_IMPLICIT);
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/amaze/filemanager/ui/dialogs/GeneralDialogCreation.java
```diff
@@ -922,32 +926,61 @@ public static void showCompressDialog(final MainActivity m, final ArrayList<Hybr
...
+        final String path = m.getCurrentPath();
         int accentColor = m.getMainActivity().getAccent();
         String[] sort = m.getResources().getStringArray(R.array.sortby);
-        int current = Integer.parseInt(sharedPref.getString("sortby", "0"));
+        int current = SortHandler.getSortType(m.getContext(), path);
         MaterialDialog.Builder a = new MaterialDialog.Builder(m.getActivity());
         a.theme(appTheme.getMaterialDialogTheme());
         a.items(sort).itemsCallbackSingleChoice(current > 3 ? current - 4 : current,
                 (dialog, view, which, text) -> true);
+        final Set<String> onlyThisFloders = sharedPref.getStringSet(PREFERENCE_SORTBY_ONLY_THIS, new HashSet<>());
```
EmmanuelMess: > <p>Note that you <em>must not</em> modify the set instance returned
> by this call.  The consistency of the stored data is not guaranteed
> if you do, nor is your ability to modify the instance at all.

-- [source](https://developer.android.com/reference/android/content/SharedPreferences.html#getStringSet(java.lang.String,%20java.util.Set%3Cjava.lang.String%3E))

I like your use of Set for efficiently finding if Strings exist.

ningshu:
Error usage on api [getStringSet](https://developer.android.com/reference/android/content/SharedPreferences.html#getStringSet(java.lang.String,%20java.util.Set%3Cjava.lang.String%3E))  has been fixed.

VishalNehra:
@TranceLove please test this PR and give your views.
I tested on my Nexus 6p it's never successful no matter what I do. It always fails with logs saying the table doesn't exist. I know it's not possible since all queries written are correct. This might be my device (ROM) specific. 
I also tested on N5x emulator running P with success.

TranceLove:
A possible cure to the symptom: https://stackoverflow.com/a/6791968/768633

VishalNehra:
Maybe adding it in constructor instead for each instance would be better.
Nice catch! I couldn't find the problem after spending much time.

TranceLove:
But a question before going to conclusion: why `UtilsHandler` never had such problem? We asked it to create records early enough?

Lazy init kills I suppose...

VishalNehra:
Not sure... I've done this in multiple projects without fail.. not sure why this problem all of a sudden.

EmmanuelMess:
The call to `onUpgrade()` is not performed unless `getWritableDatabase();` is called, happed to me in SimpleAccounting. @ningshu please explain the line with a comment so that we don't delete it later.

VishalNehra:
@EmmanuelMess please add commits to this PR, I think we can push this with v3.3.0 so let's not wait over these things.

ningshu:
The problem of missing ‘sort’ table has been fixed.

TranceLove:
Two ticks, merging.

PR #1265, opened 2018-06-03
This should prevent false hits when only the size of the requested image changes

EmmanuelMess:
This should prevent false cache hits when only the size of the requested image changes.


VishalNehra:
Will be closing this as there's some problem with glide not identifying encoder correctly if we use hashes as keys for remote connection.
Will create a new issue for the same.

PR #1266, opened 2018-06-05
Fix intent-filter for cloudrail to identify entrypoint for google drive

VishalNehra:
There was no issue created for this. But it fixes the redirect problem on current master with google drive connection. We currently get no redirect back to amaze so the connection is never created.
I don't think I've added SHA-1 fingerprint for your android studio debug key so you shouldn't even be able to add a drive connection but rest assured, I tested this and it's safe to merge.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -89,13 +89,13 @@
                 <category android:name="android.intent.category.DEFAULT" />
                 <category android:name="android.intent.category.BROWSABLE" />
                 <data android:scheme="com.amaze.filemanager" />
-                <data android:mimeType="resource/folder" />
             </intent-filter>
 
             <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
                 <data android:mimeType="resource/folder" />
                 <category android:name="android.intent.category.DEFAULT" />
+                <data android:scheme="com.amaze.filemanager" />
             </intent-filter>
 
             <intent-filter android:label="@string/intent_save_as">
--- a/app/src/main/java/com/amaze/filemanager/activities/MainActivity.java
+++ b/app/src/main/java/com/amaze/filemanager/activities/MainActivity.java
@@ -244,7 +244,7 @@ public class MainActivity extends PermissionsActivity implements SmbConnectionLi
     public static final String ARGS_INTENT_ACTION_VIEW_MIME_FOLDER = "resource/folder";
 
     private static final String CLOUD_AUTHENTICATOR_GDRIVE = "android.intent.category.BROWSABLE";
-    private static final String CLOUD_AUTHENTICATOR_REDIRECT_URI = "com.amaze.filemanager:/oauth2redirect";
+    private static final String CLOUD_AUTHENTICATOR_REDIRECT_URI = "com.amaze.filemanager:/auth";
 
     // the current visible tab, either 0 or 1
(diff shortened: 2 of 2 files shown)


TranceLove:
If I can't test the code actually on device I cannot comment on the code nor give +1.

If adding our debug keys is inconvenient to you, I think at least you send us an apk with changed code applied, with all proguard/minify/etc options disabled, so I can verify the code is there with a decompiler.

app/src/main/AndroidManifest.xml
```diff
@@ -89,13 +89,13 @@
                 <category android:name="android.intent.category.DEFAULT" />
                 <category android:name="android.intent.category.BROWSABLE" />
                 <data android:scheme="com.amaze.filemanager" />
-                <data android:mimeType="resource/folder" />
```
EmmanuelMess: Wasn't this used?

VishalNehra: This is being used in another intent-filter now.

VishalNehra:
If you want to test then you need to give me the Signing-certificate fingerprint of your environment.
Below is the summary of it in Google's words 

Signing-certificate fingerprint
---
Add your package name and SHA-1 signing-certificate fingerprint to restrict usage to your Android apps [Learn more](https://support.google.com/googleapi/answer/6158849?hl=en_US#installedapplications&android)
Get the package name from your AndroidManifest.xml file. Then use the following command to get the fingerprint:
````
 keytool -exportcert -keystore path-to-debug-or-production-keystore -list -v
````
After I add these in Google Cloud console then you'll be able to connect to Google Drive using the debug compiles from your Android Studio.

**Comment by VishalNehra** [comment]
Guys please respond fast.

**Comment by TranceLove** [comment]
Couldn't get it to work on my Oneplus One running Slim7(7.1.2). Had explicitly tell Android Studio to use my debug key to sign the app. Even I'm on a different machine, the key is still the same - had verified their SHA512 checksum, they are the same as one I sent you earlier.

screenshot from 2018-06-14 10-46-20

screenshot from 2018-06-14 10-46-33

When I tried to open a Google Drive connection, it did fired up CloudRail, then asked me the Google account to use at my web browser (defaulted to Firefox), and prompted me to grant permission to Amaze FM.

Then in logcat I did saw
```
06-14 10:40:53.782 906-3859/system_process I/ActivityManager: START u0 {act=android.intent.action.VIEW dat=com.amaze.filemanager:/auth?code=4/<some code> cmp=com.amaze.filemanager/.activities.MainActivity (has extras)} from uid 10103 on display 0
```
but the Google Drive shortcut didn't create in Drawer, nor my Google Drive list was brought up in Amaze.

Did I miss anything here?

VishalNehra:
Yes, you need to use Google Chrome to authenticate the Drive connection.

TranceLove:
Verified working on Pixel 2 emulator running 7.1 Galaxy S2 running SlimLP (5.1.1) and Google Chrome installed, Google Drive bookmark successfully created at Drawer and is able to access Google Drive.

Only a question: is it the expected behaviour that after Chrome redirect user back to Amaze, only the bookmark is created but won't load the file list on user's Google Drive automatically?

Will vote +1 if the above answer is yes.

VishalNehra:
@EmmanuelMess please check your email and provide me your SHA keys from debug keystore so that you can test this PR.

Issue #1159, opened 2018-04-11
"Open as..." dialog after 'creating' a bookmark on API 15


EmmanuelMess:
Testing with what seems to be my everything-that-could-possibly-break-breaks phone, when a bookmark is selected in the drawer, the app supposedly 'creates' a new folder ("Bookmark not found creating new folder" toast) and shows the "open as..." dialog, as if what was created were a file.

TranceLove:
Reproducible with Nexus One emulator running ICS, ARM or x86.

Should be hitting the same problem as #1160, that is a consequence of system denied creating a folder and bookmark in an nonexistent user storage space.

TranceLove:
An update to the issue: it does related to the broken Android emulator implementation, as mentioned [here](https://stackoverflow.com/a/45967319/768633). Basically, downgrade the SDK tools launching the emulator instead of strictly from within Android Studio would fix the storage issue.

So try these to launch the ICS emulator:
1. Download Android SDK tools r25.2.5 ([Windows](http://dl-ssl.google.com/android/repository/tools_r25.2.5-windows.zip), [Linux](http://dl-ssl.google.com/android/repository/tools_r25.2.5-linux.zip), [OSX](http://dl-ssl.google.com/android/repository/tools_r25.2.5-macosx.zip))
2. Extract under `$ANDROID_SDK_ROOT`. Of course you'll want to use a different name for the folder, like `tools-r25`, to prevent conflict/overwriting
3. Inside `$ANDROID_SDK_ROOT/tools-r25`, run the emulator using command line

`./emulator -avd Nexus_One_API_15`

You should see the emulator correctly mount and prepare the SD card partition, and create file would work (using current `master`).

I'll mark this as invalid and close. Reopen this should any problems arise even with the above workaround, or with a physical device.

PR #1268, opened 2018-06-07
Null safety on EditText.getTypeFace() for API 15

### LINKED ISSUE #1160: Crash after opening a text file on API 15
Crash after opening a text file on API 15, `getExternalCacheDir()` is null. 
```gradle
com.amaze.filemanager E/AndroidRuntime: FATAL EXCEPTION: main
    java.lang.NullPointerException
        at com.amaze.filemanager.activities.TextEditorActivity.lambda$load$4$TextEditorActivity(TextEditorActivity.java:296)
        at com.amaze.filemanager.activities.TextEditorActivity$$Lambda$3.onAsyncTaskFinished(Unknown Source)
        at com.amaze.filemanager.asynchronous.asynctasks.ReadFileTask.onPostExecute(ReadFileTask.java:125)
        at com.amaze.filemanager.asynchronous.asynctasks.ReadFileTask.onPostExecute(ReadFileTask.java:27)
        at android.os.AsyncTask.finish(AsyncTask.java:602)
        at android.os.AsyncTask.access$600(AsyncTask.java:156)
        at android.os.AsyncTask$InternalHandler.handleMessage(AsyncTask.java:615)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:137)
        at android.app.ActivityThread.main(ActivityThread.java:4424)
        at java.lang.reflect.Method.invokeNative(Native Method)
        at java.lang.reflect.Method.invoke(Method.java:511)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:784)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:551)
        at dalvik.system.NativeStart.main(Native Method)
```

TranceLove:
Fixes #1160. If `EditText.getTypeFace()` is null the default typeface (`TypeFace.DEFAULT`) will be inserted into the `EditText`. Also reversed equality checks (`mInputTypefaceMono.equals(mInput.getTypeface()`) for null safety.

This should be where the crash came from; however on testing with emulator running API 15, there had been frequent cases opening text file ended up crashes Android itself (black screen -> boot animation -> lock screen). Debugging app suggested the crash was inside `Looper.loop()`, which seems to be out of my knowledge...

The only message in the logcat I can see which makes sense:

`DexOpt: unable to opt direct call 0x00d0 at 0x8f in Lcom/amaze/filemanager/activities/TextEditorActivity;.onCreate`

But in debug session the whole `TextEditorActivity.onCreate()` ran smooth without any problem.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -88,7 +88,7 @@ dependencies {
     testImplementation "org.apache.sshd:sshd-core:1.7.0"
 
     androidTestImplementation 'com.android.support.test.espresso:espresso-core:3.0.0'
-    androidTestImplementation 'com.android.support.test:runner:1.0.2'
+    debugImplementation 'com.android.support.test:runner:1.0.2'
     androidTestImplementation 'com.android.support.test:rules:1.0.2'
     androidTestImplementation 'com.android.support:support-annotations:27.0.2'
 
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -29,6 +29,7 @@ android {
         multiDexEnabled true
 
         vectorDrawables.useSupportLibrary = true
+        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
     }
 
     signingConfigs {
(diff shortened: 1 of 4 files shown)


EmmanuelMess:
I think it is known that emulator do not always replicate what real devices do. Can't we make Robolectric tests that use API 15 and just test if text can be displayed?

TranceLove:
Their minimum supported API is 4.1.2 (16). See

https://github.com/robolectric/robolectric/blob/2fbd20743bf82055c14abdd1ebeb5caad5cb9d31/robolectric/src/main/java/org/robolectric/internal/SdkConfig.java#L20-L31

TranceLove:
Since Robolectric test is a no-go here, I'll had to revert to use Espresso (#1274) to write the test - see 472a35d. It is reading the file within the emulator, so no need to worry about the storage-not-found problem (#1159).

Test case ran successfully with Nexus One emulator running 4.0.3 (API 15).

A drawback regarding this test was that, I only launch TextEditorActivity standalone with a `ACTION_VIEW` intent, but not from within Amaze FM; but since this PR is only about TextEditorActivity, I think it should be fine... though tests describing the complete flow can be written with extra time anyway.

PR #1270, opened 2018-06-08
fix missing url error

shellhub:


TranceLove:
May I know the exact missing URL error you're fixing?

shellhub:
you will see this error in latest version android studio @TranceLove

TranceLove:
OK, [this is what is happening when opening AndroidManifest.xml with latest version of Android Studio](https://developer.android.com/studio/write/app-link-indexing): support of opening a HTTP(S) link that will end up open the app directly.

I won't say it's an error, at most it should be a warning.

I can't speak for others within the team, but I think we won't be using this, as I haven't found a suitable use case yet, so I will vote +1 here.

However, please be reminded that not everyone of us will always upgrade Android Studio, either we are running low on bandwidth, on the road, or simply being lazy. You will help us a lot by saying this at the first place, so your work can be reviewed sooner.

shellhub:
I think also it is warning instead of error, but android studio (maybe old version) say it is error, so we should add `tools:ignore="AppLinkUrlError"` to intent-filter make code without any error!
screen shot 2018-06-09 at 10 50 05 am

Issue #1244, opened 2018-05-20
Trying to connect to SMB fails


EmmanuelMess:
AVD Pixel 2 with API 27

```
? E/TrustyKeymaster: calling trusty_keymaster_call insize 199 msg size 203
    Received 4 byte response
    Error: Response of size 4 contained error code -33
com.amaze.filemanager W/System.err: java.security.UnrecoverableKeyException: Failed to obtain information about key
com.amaze.filemanager W/System.err:     at android.security.keystore.AndroidKeyStoreProvider.loadAndroidKeyStoreSecretKeyFromKeystore(AndroidKeyStoreProvider.java:282)
        at android.security.keystore.AndroidKeyStoreSpi.engineGetKey(AndroidKeyStoreSpi.java:98)
        at java.security.KeyStore.getKey(KeyStore.java:1062)
        at com.amaze.filemanager.utils.files.CryptUtil.getSecretKey(CryptUtil.java:386)
        at com.amaze.filemanager.utils.files.CryptUtil.aesEncryptPassword(CryptUtil.java:276)
        at com.amaze.filemanager.utils.files.CryptUtil.encryptPassword(CryptUtil.java:478)
        at com.amaze.filemanager.utils.SmbUtil.getSmbEncryptedPath(SmbUtil.java:56)
        at com.amaze.filemanager.ui.dialogs.SmbConnectDialog.lambda$onCreateDialog$2$SmbConnectDialog(SmbConnectDialog.java:268)
        at com.amaze.filemanager.ui.dialogs.SmbConnectDialog$$Lambda$2.onClick(Unknown Source:62)
        at com.afollestad.materialdialogs.MaterialDialog.onClick(MaterialDialog.java:410)
        at android.view.View.performClick(View.java:6294)
        at android.view.View$PerformClick.run(View.java:24770)
        at android.os.Handler.handleCallback(Handler.java:790)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:164)
com.amaze.filemanager W/System.err:     at android.app.ActivityThread.main(ActivityThread.java:6494)
        at java.lang.reflect.Method.invoke(Native Method)
? W/audio_hw_generic: Not supplying enough data to HAL, expected position 5063939 , only wrote 5063760
com.amaze.filemanager W/System.err:     at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:438)
[...]

PR #1274, opened 2018-06-09
Add Espresso test framework to source tree

TranceLove:
Facilitate writing test cases that requires Android device (physical/emulator) to run, as in #1244. The test case written for #1244 was included for reference.

A note about this line in app/build.gradle:
> `testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"`

It didn't affect Robolectric-based tests, and we only run tests from Android Studio, it should be fine.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -88,7 +88,7 @@ dependencies {
     testImplementation "org.apache.sshd:sshd-core:1.7.0"
 
     androidTestImplementation 'com.android.support.test.espresso:espresso-core:3.0.0'
-    androidTestImplementation 'com.android.support.test:runner:1.0.2'
+    debugImplementation 'com.android.support.test:runner:1.0.2'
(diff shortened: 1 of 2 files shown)


EmmanuelMess:
Having any other name for "test" folder for Robolectric caused issues, "androidTest" is the default for Espresso?

TranceLove:
> Having any other name for "test" folder for Robolectric caused issues, "androidTest" is the default for Espresso?

Or you'll risk Android Studio not launching the test in your device/emulator. Same goes to Kotlin sources in Android Studio, they must be put in `/src/{main,test}/java` but not `/src/{main,test}/kotlin` which works perfectly well in IDEA.

IIRC I hit the wall when I do the test case back, but I'll try once more to be sure.

TranceLove:
Closing this PR because changes here had been merged to #1268.

PR #1276, opened 2018-06-10
Bump version number to v3.2.2

EmmanuelMess:
And with this I'll release, so voice mistakes here.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -19,8 +19,8 @@ android {
         applicationId "com.amaze.filemanager"
         minSdkVersion 14
         targetSdkVersion 25
-        versionCode 63
-        versionName "3.2.1"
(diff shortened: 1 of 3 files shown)


EmmanuelMess:
@VishalNehra you have been releasing for the past 4 years, any advice?

VishalNehra:
Release in batches, but since there's a minor fix which is well tested, there shouldn't be a problem to release for 100% of target users. Also, any change in listing (new version/description change) takes around 1 hour to reflect on Play Store.
Other than that, there's not much to tell, it's a simple process. Just remember to keep keystore private and safe.

EmmanuelMess:
Done.

EmmanuelMess:
Now, what about FDroid?

PR #1278, opened 2018-06-10
Update compile SDK version + depending libs to version 27

TranceLove:
Fixes problem of proguard failed to run on making release builds, when this message was seen in console output

`Warning: com.bumptech.glide.load.resource.bitmap.VideoDecoder: can't find referenced method 'android.graphics.Bitmap getScaledFrameAtTime(long,int,int,int)' in library class android.media.MediaMetadataRetriever`

Does not affect 3.2.x branches, but >= 3.3.0 since proguard is enabled for their release builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ android:
     - platform-tools
     - tools
     - build-tools-27.0.3
-    - android-26
+    - android-27
     - extra-google-m2repository
 
 script: "./gradlew assembledebug"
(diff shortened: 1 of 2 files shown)


VishalNehra:
So you say we won't be able to compile a release build after merging this PR or we're unable to do that on current master?

TranceLove:
We need this PR to compile a release build with current `master`/`release-3.3.0` branch.

VishalNehra:
Ahh, ok my bad. Was half asleep at that time :P

PR #1284, opened 2018-06-13
Fix processviewer notification

### LINKED ISSUE #1286: On copy "Processing" dialog stays indefinitely when there's not enough space
When copying "Processing" dialog stays indefinitely when there's not enough space.

VishalNehra:
1. Fixes opening of ProcessViewerFragment when not inside app instance.
2. Add ability to expand notification to view more info without going to ProcessViewerFragment.
With this I'm done with my share of fixing major bugs for v3.3.0. Let's quickly get it out.
3. Fixes #1286


TranceLove:
Verified working with Oneplus One running Slim7 (7.1.2), Galaxy S2 running SlimLP (5.1.1), and Pixel 2 emulators running 7.1 and 5.1.1 testing copy files over SSH. However, a glitch about the notification: the progress bar overlapped with the texts, with both English/non-English device locales and English/non-English filenames.

(Pixel 2 emulator running 7.1)
screenshot_1528948189

On my Galaxy S2 the progress bar was even not visible.

screenshot_2018-06-14-12-05-01

Additionally, it would be nice if I can see an icon indicating process is in progress.

EmmanuelMess:
Maybe remove the transfer speed, because most users won't care, and time remaining gives a rough estimate.

VishalNehra:
See if all your issues are now fixed @TranceLove

TranceLove:
Looks like the same to me.

Pixel 2 emulator running 7.1/Oneplus One running Slim7 (7.1.2)
screenshot_1529027968

Pixel 2 emulator running  5.1. Font colour did changed though
screenshot_2018-06-15-11-11-17

And on testing with Galaxy S2 running SlimLP (5.1.1), I even got a crash on copy, either over SSH or SMB.
```
06-15 09:42:11.958 7217-7546/com.amaze.filemanager E/AndroidRuntime: FATAL EXCEPTION: service_progress_watcher
    Process: com.amaze.filemanager, PID: 7217
    java.lang.ArithmeticException: divide by zero
        at com.amaze.filemanager.asynchronous.services.AbstractProgressiveService.publishResults(AbstractProgressiveService.java:146)
        at com.amaze.filemanager.asynchronous.services.CopyService$DoInBackground.lambda$doInBackground$0$CopyService$DoInBackground(CopyService.java:245)
        at com.amaze.filemanager.asynchronous.services.CopyService$DoInBackground$$Lambda$0.onProgressed(Unknown Source)
        at com.amaze.filemanager.utils.ProgressHandler.addWrittenLength(ProgressHandler.java:68)
        at com.amaze.filemanager.utils.ServiceWatcherUtil$1.run(ServiceWatcherUtil.java:125)
        at android.os.Handler.handleCallback(Handler.java:739)
        at android.os.Looper.loop(Looper.java:135)
```

Culprit
https://github.com/TeamAmaze/AmazeFileManager/blob/4b713b6b1c1a0358ee1b9f9a9d8a1f6d128c3464/app/src/main/java/com/amaze/filemanager/asynchronous/services/AbstractProgressiveService.java#L146

I had to add a little hack to ensure `speed` is never 0 to get it through. Not sure if it's my device's problem as it didn't happen on emulator running 5.1, 7.1, nor Oneplus One running Slim7 (7.1.2).
```
if(speed == 0)
    speed = 1;
String remainingTime = Utils.formatTimer(Math.round((totalSize-writtenSize)/speed));
```

app/src/main/java/com/amaze/filemanager/fragments/ProcessViewerFragment.java
```diff
@@ -209,7 +221,7 @@ public void processResults(final DatapointParcelable dataPackage, int serviceTyp
 
             Spanned timerSpan = Html.fromHtml(getResources().getString(R.string.service_timer)
                     + ": <font color='" + accentColor + "'><i>"
-                    + formatTimer(++time)
+                    + Utils.formatTimer(++time)
```
EmmanuelMess: I don't understand, why are you doing `++time`, then formating, are you supposing that this function will be called exactly every second? Why aren't you setting a start time and then recalculating the elapsed time?

VishalNehra: Yes, this function will be called every second. Setting a start time and calculating elapsed time would be waste of resources.

VishalNehra:
I've fixed the progress bar visibility bug and white text color bug with samsung devices.
I've also fixed #1286 while I was at it. 
Only problem here is that the accent color can't be applied to progress bar due to limitations in API. The methods are there but are hidden in SDK 26. Let's hope future releases allow us to change the progress bar tint too.

TranceLove:
Tested with Oneplus One running Slim7 (7.1.2) and Galaxy S2 running SlimLP (5.1.1). Layout fix OK, divide by zero bug fixed, symptom in #1286 is also fixed.

However, on copying files over SMB/SSH, when I press the stop button at the notification, the notification is gone, but I saw the copy progress didn't stop - there is still network activity.

VishalNehra:
@TranceLove that's not related to this PR, you can create new issue for that. Basically I don't cancel unless the current file is finished copying. If it's say 100 Mbs then it'll only cancel after copying the 100 mbs.

TranceLove:
OK. Then I should have no more questions to ask for now. +1.

VishalNehra:
I already had plans to fix that, maybe will do it now for the better. Will work on it in another PR.

app/src/main/java/com/amaze/filemanager/fragments/ProcessViewerFragment.java
EmmanuelMess: PRing explanation.

PR #1287, opened 2018-06-14
 Added type parametrization for AppConfig.runInParallel() 

EmmanuelMess:
* Added type parametrization for AppConfig.runInParallel() 
* Changed class from interface to abstract class
* Parameters are now passed to constructor
* onPreExcecute and onPostExcecute are optional


EmmanuelMess:
Rebased, merging.

PR #1291, opened 2018-06-16
Swap parameter order of saving SMB connection data to database

### LINKED ISSUE #1289: SMB connection entries are corrupted
I was testing something and noticed after adding a new SMB connection, it loads fine for that time, but after restarting Amaze the entry in drawer is corrupted and after pressing on it, nothing happens. Even pressing on edit icon doesn't give any response.
Find the screenshot attached below. The server name was Vishal and it automatically changed to something else.
screenshot_amaze_20180615-024013

TranceLove:
Fixes #1289.

I think it's save to blame @EmmanuelMess for causing this - see 61c27cde ;)

(OK, also us for not noticing this.)


EmmanuelMess:
Nope, I noticed it, I just left it in for someone else to fix... and forgot to make an issue, sorry.

Issue #1301, opened 2018-06-20
Socket for secure FTP connection is cut and IP not visible


EmmanuelMess:
Phone: Moto E Gen 2 4G
<img src="https://user-images.githubusercontent.com/10991116/41665112-98de96f2-747d-11e8-8353-4450f8f08582.png" height="250px"/>

TranceLove:
You say you're running on 4G... but Amaze should have disallowed you to even start the FTP server at all.

If you are running on 4G with Wifi AP enabled, at my case it showed me an IP that is not the AP's IP (192.168.43.1). This should be fixed, but I'm not sure this is related to your problem.

(e.g. they gave you an IPv6 that we cannot display?)

EmmanuelMess:
The phone model is called that, not running on 4G.

PR #1307, opened 2018-06-21
FTPService URL display fixes

### LINKED ISSUE #1300: Secure FTP IP starts with "sftp://" instead of "ftps://"
Secure FTP IP starts with "sftp://" instead of "ftps://", the first one is:
> FTP over SSH is the practice of tunneling a normal FTP session over a Secure Shell connection.

And "ftps://" stands for:

> FTPS (also known as FTPES, FTP-SSL, and FTP Secure) is an extension to the commonly used File Transfer Protocol (FTP) that adds support for the Transport Layer Security (TLS) and the Secure Sockets Layer (SSL) cryptographic protocols.

I think we are showing to the user the wrong protocol as being used. But then again, I don't know much of networking protocols.

TranceLove:
* `getLocalInetAddress()` should always return 192.168.43.1 (AP's hard-coded IP address) when wifi AP is on (related to #1301)
 * URL prefix for FTP server running at implicit SSL mode changed to `ftps://`. Fixes #1300 
 * Null safety in `isEnabledWifiHotspot()`, for prevent crashing FTPService in Robolectric tests

### DIFF
--- a/app/src/androidTest/java/com/amaze/filemanager/asynchronous/services/ftp/FTPServiceStaticMethodsTest.java
+++ b/app/src/androidTest/java/com/amaze/filemanager/asynchronous/services/ftp/FTPServiceStaticMethodsTest.java
@@ -1,0 +1,47 @@
+package com.amaze.filemanager.asynchronous.services.ftp;
+
+import android.content.Context;
+import android.support.test.InstrumentationRegistry;
+import android.support.test.filters.SmallTest;
+import android.support.test.runner.AndroidJUnit4;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import static org.junit.Assert.*;
+
+/**
+ * This test is separated from FTPServiceEspressoTest since it does not actually requires the FTP
+ * service itself.
+ *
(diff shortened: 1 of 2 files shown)


TranceLove:
@EmmanuelMess May I ask for your help in debugging #1301? Please see if you can run `FTPServiceStaticMethodsTest.testGetLocalInetAddressMustNotBeEmpty()` on your Moto successfully, and tell us what you can find.

EmmanuelMess:
`FTPServiceStaticMethodsTest.testGetLocalInetAddressMustNotBeEmpty()` test is passed, furthermore `FTPServerFragment.spannedStatusSecure` is correctly set up. It is probably just that my phone is too narrow. One solution is to put the IP in a second line:
Status: Active.
IP: ftps://192.168.0.10:2211

TranceLove:
Good to hear that. I'll see what I can do here then.

EmmanuelMess:
@TranceLove you should create another PR for that.

TranceLove:
Won't mind a new PR if this one is going into `release-3.3.0` branch.

PR #1313, opened 2018-06-24
FTPServiceEspressoTest: added file upload/download test

TranceLove:
To ensure binary/text file upload/download is working, for FTP server operate with/without implicit SSL.

### DIFF
--- a/app/src/androidTest/java/com/amaze/filemanager/asynchronous/services/ftp/FtpServiceEspressoTest.java
+++ b/app/src/androidTest/java/com/amaze/filemanager/asynchronous/services/ftp/FtpServiceEspressoTest.java
@@ -27,6 +27,7 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.FileWriter;
+import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.Writer;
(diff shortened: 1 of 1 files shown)


app/src/androidTest/java/com/amaze/filemanager/asynchronous/services/ftp/FTPServiceEspressoTest.java
```diff
@@ -83,6 +101,89 @@ private void loginAndVerifyWith(FTPClient ftpClient) throws Exception
             fail("Download folder not found on device. Either storage is not available, or something is really wrong with FTPService. Check logcat.");
     }
 
+    private void testUploadWith(FTPClient ftpClient) throws Exception
```
EmmanuelMess: I think it is better to throw something specific and let the app crash if something completely unexpected happens.

TranceLove: Though I used an idiot way for assertion (`fail()` instead of `assertTrue()` which is corrected now), I think the assertion needs to be here, and failures accessing the filesystem will crash the service and the test anyway.

EmmanuelMess: Sorry, I'm not following you, I meant to say that instead of `throws Exception` you should put a specific subclass, like `throws IOException` for example.

EmmanuelMess:
Check PR

PR #1334, opened 2018-07-02
Replaced processing state of services notification with purpose built processing notification

EmmanuelMess:
Replaced processing state of services notification with purpose built processing notification. So that the notification doesn't show "unknown" fields.

@VishalNehra do check it out and tell me what you think

Don't merge yet, haven't really tested.

Known bugs:
* Flickering not fixed, the flickering seems to be due to a creation of the same Service quite a few times.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/asynchronous/services/CopyService.java
+++ b/app/src/main/java/com/amaze/filemanager/asynchronous/services/CopyService.java
@@ -48,6 +48,7 @@ import com.amaze.filemanager.filesystem.HybridFileParcelable;
 import com.amaze.filemanager.filesystem.Operations;
 import com.amaze.filemanager.filesystem.RootHelper;
 import com.amaze.filemanager.ui.notifications.NotificationConstants;
+import com.amaze.filemanager.ui.notifications.ProcessingNotificationBuilder;
 import com.amaze.filemanager.utils.DatapointParcelable;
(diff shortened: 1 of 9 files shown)


EmmanuelMess:
You can now test.

VishalNehra:
I really don't think this is a good idea. Too much inconsistent notification when it gets halted in between from time to time (on slow connections), the bigView changes to smallView of the processing notification.
Plus, the layout itself isn't much pleasing...
screenshot_amaze_20180708-025003

EmmanuelMess:
I need more constructive ideas...

app/src/main/java/com/amaze/filemanager/asynchronous/services/CopyService.java
```diff
@@ -198,6 +200,26 @@ protected void setPercentProgress(float progress) {
...
     }
 
+    @Override
+    protected ProcessingNotificationBuilder getProcessingNotificationBuilder() {
+        String filenameText;
+
+        if(fileAmount == 1) {
+            filenameText = firstFilename;
+        } else if(fileAmount == 2) {
+            filenameText = c.getString(R.string.processing_notif_two_filenames, firstFilename, secondFilename);
```
EmmanuelMess: Self note: see of using `TextUtils.listEllipsize`.

PR #1335, opened 2018-07-03
SSH/SFTP: Fix symlinks to files/directories treated as files using stat()

### LINKED ISSUE #1330: Bug: Directories are treated as files SFTP
With the new SFTP feature, when connected to a server, some directories are incorrectly treated as files, prompting them to be downloaded when clicked instead of opened.

Version: v3.3.0RC1

This is the root of my server. Nothing here should be files, it's all directories.
screenshot_amaze_20180702-074135

TranceLove:
Fixes #1330. Extra `stat()` will be called if the file/directory is actually a symbolic link.

Test case included, but be reminded that it may not run on non-POSIX OSes.

(And the way I coded the tests were actually a convenience but disregarded the fact that Android's user space storage never understands symlinks ;) )


Limero:
Great work. Thank you very much!

Issue #1333, opened 2018-07-02
Parsing problem with SFTP server path from database


VishalNehra:
[A review](https://forum.xda-developers.com/showpost.php?p=76963297&postcount=654) from one of XDA's user, scanno
His logs:

```
07-02 12:47:24.682 W/Uri (27485): Error parsing port string. 
07-02 12:47:24.682 W/Uri (27485): java.lang.NumberFormatException: For input string: "***" 
07-02 12:47:24.682 W/Uri (27485): at java.lang.Integer.parseInt(Integer.java:608) 
07-02 12:47:24.682 W/Uri (27485): at java.lang.Integer.parseInt(Integer.java:643) 
07-02 12:47:24.682 W/Uri (27485): at android.net.Uri$AbstractHierarchicalUri.parsePort( Uri.java:1136) 07-02 12:47:24.682 W/Uri (27485): at android.net.Uri$AbstractHierarchicalUri.getPort(Ur i.java:1115) 
07-02 12:47:24.682 W/Uri (27485): at com.amaze.filemanager.filesystem.ssh.SshConnection Pool.create(SshConnectionPool.java:180) 
07-02 12:47:24.682 W/Uri (27485): at com.amaze.filemanager.filesystem.ssh.SshConnection Pool.create(SshConnectionPool.java:173) 
07-02 12:47:24.682 W/Uri (27485): at com.amaze.filemanager.filesystem.ssh.SshConnection Pool.getConnection(SshConnectionPool.java:98) 07-02 12:47:24.682 W/Uri (27485): at com.amaze.filemanager.filesystem.ssh.SshClientUtil s.execute(SshClientUtils.java:72)
 07-02 12:47:24.682 W/Uri (27485): at com.amaze.filemanager.filesystem.ssh.SshClientUtil s.execute(SshClientUtils.java:127) 
07-02 12:47:24.682 W/Uri (27485): at com.amaze.filemanager.filesystem.HybridFile.forEac hChildrenFile(HybridFile.java:732) 
07-02 12:47:24.682 W/Uri (27485): at com.amaze.filemanager.asynchronous.asynctasks.Load FilesListTask.doInBackground(LoadFilesListTask.jav a:131) 
07-02 12:47:24.682 W/Uri (27485): at com.amaze.filemanager.asynchronous.asynctasks.Load FilesListTask.doInBackground(LoadFilesListTask.jav a:62) 
[...]

PR #1336, opened 2018-07-03
Little code cleanups at SftpConnectDialog and SshClientUtils

TranceLove:
* Organise imports
* Remove redundant log statements
* Use String.format(Locale.ENGLISH) to format strings per recommended


EmmanuelMess:
Isn't this just hardcoding the English locale? Shouldn't we put the strings as a resource that's translatable instead?

TranceLove:
They are actually URIs which have no translations, just did that per IDE suggested.

Seriously I won't mind closing this PR since about half an hour later [I got my clue](https://forum.xda-developers.com/showpost.php?p=76974221&postcount=659) about #1333 ;)

(But the first two changes will be done anyway.)

PR #1349, opened 2018-07-06
Bypass space check when API supports moving through renaming paths

### LINKED ISSUE #513: Move files without copying them
For now if I move a file it is copied and then the original is deleted. It has tons of drawbacks such as flash memory wear and inability to "move" when flash if full. Move instead.

### LINKED ISSUE #895: Checking free space when moving file in the same partition
I want to move a file to other folder and I can't because I get an message "Insuficient space", the filesize is bigger than the free space.

This issue maybe can be related with #513

Note I am trying to move a file in the same partition.

Sorry my poor english.

### LINKED ISSUE #1345: Bug: "Insufficient space" error can occur when moving local files
The function for calculating space needed for a move does not take in to account the space available when the move is done.

For example, if we have a device with 1GB of available storage space, then it's not possible to move any file/directory bigger than 1GB around the local file system.

Versions affected: v3.2.2, v3.3.0RC1 and most likely, previous versions.

VishalNehra:
Fixes #1345 and fixes #895 and fixes #513 (?)
Had plans for long to fix this problem. Finally it's done.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/MoveFiles.java
+++ b/app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/MoveFiles.java
@@ -22,12 +22,15 @@ package com.amaze.filemanager.asynchronous.asynctasks;
 import android.content.Context;
 import android.content.Intent;
 import android.os.AsyncTask;
+import android.widget.Toast;
 
+import com.amaze.filemanager.R;
 import com.amaze.filemanager.activities.MainActivity;
 import com.amaze.filemanager.activities.superclasses.ThemedActivity;
 import com.amaze.filemanager.database.CryptHandler;
 import com.amaze.filemanager.database.models.EncryptedEntry;
 import com.amaze.filemanager.exceptions.ShellNotRunningException;
+import com.amaze.filemanager.filesystem.HybridFile;
 import com.amaze.filemanager.filesystem.HybridFileParcelable;
 import com.amaze.filemanager.fragments.MainFragment;
 import com.amaze.filemanager.utils.application.AppConfig;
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/MoveFiles.java
```diff
@@ -188,6 +206,13 @@ public void onPostExecute(Boolean movedCorrectly) {
             });
 
         } else {
+
+            if (destinationSize < totalBytes) {
+                // destination don't have enough space; return
+                publishProgress(context.getResources().getString(R.string.in_safe));
```
EmmanuelMess: This is a hack, no *progress* is being published. I would have said "please use onPostExcecute()", but this is onPostExcecute(), please simply move the Toast creation here (and remember to remove collateral changes).

app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/MoveFiles.java
```diff
@@ -68,12 +75,23 @@ public MoveFiles(ArrayList<ArrayList<HybridFileParcelable>> files, MainFragment
...
+        Toast.makeText(context, message[0], Toast.LENGTH_LONG).show();
+    }
+
     @Override
     protected Boolean doInBackground(ArrayList<String>... strings) {
         paths = strings[0];
 
         if (files.size() == 0) return true;
 
+        for (ArrayList<HybridFileParcelable> filesCurrent : files) {
```
EmmanuelMess: I would move this (new code) to right after the switch statement (after all other checks have been made), because of asymptotic cost.

VishalNehra: Moving them after switch can leave some variables uninitialized in some cases.

EmmanuelMess: `totalBytes` and `destinationSize` both seem to only be used after the `doInBackground()`.

VishalNehra: Yes, but for some cases in switch, we return out of ````doInBackground```` which can leave them uninitialised.

VishalNehra:
Why are we still pending here? @EmmanuelMess I did fix everything you requested.

EmmanuelMess:
Sorry for the lateness

PR #1356, opened 2018-07-09
Two small and unrelated changes to StreamSource and CloudStreamSource

EmmanuelMess:
* Removed buffer size from StreamSource and CloudStreamSource because it is not used (anymore?).
* Throw IllegalArgumentException in {StreamSource, CloudStreamSource}.moveTo(position) for positions outside of file (smaller than 0 or bigger than the length)


VishalNehra:
Please throw the exception in method declaration, which is causing travis to fail the build.

PR #1358, opened 2018-07-10
PemToKeyPairTask refactor: popup passphrase dialog at onProgressUpdate instead

### LINKED ISSUE #1354: Cannot use passphrase-protected private keys created by PuTTYGen to connect to SSH server
Steps to reproduce symptom:

1. Create a new key using puttygen with RSA, 2048 bit keys
2. Add a passphrase and save the key (`.ppk`)
3. Copy the `authorized_keys` line to SSH server
4. Copy the key saved in 2) to device
5. Launch app and setup a connection using the key
6. Enter the passphrase when prompted
7. App cannot connect to SSH server even it seems you entered the correct passphrase

Back to the days of #863 it was working, it is possible it is a regression.

TranceLove:
Fixes #1354. This prevents PemToKeyPairTask from never getting the decrypted key at postExecute loop.

Because of `AsyncTask`'s nature of `doInBackground()` is performed only once, inevitably I need to use a while loop to keep the task wait for a correct passphrase (if there is one), hence making the passphrase popup at `onProgressUpdate()`.

So sorry that I couldn't make an Espresso or Robolectric test that is working, either on device or in Android Studio, so please use the following way to verify (if you need to):

1. Generate a new key in puttygen, with passphrase (don't worry, puttygen works nice with wine too)
2. Add the key to authroized_keys on SSH server
3. Copy the key to device
4. Launch Amaze FM
5. Setup connection with the key
6. Passphrase dialog will popup, prompt for passphrase
7. Try enter incorrect passphrase, dialog will indicate incorrect passphrase and still block the connection setup dialog
8. Enter the correct passphrase to connect to the SSH server, as well as saving the connection settings
9. The next time we launch Amaze again, when we tap on the connection saved at 8), dialog will popup prompt for key passphrase. Enter the correct passphrase to connect to the SSH server

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/ssh/PemToKeyPairTask.java
+++ b/app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/ssh/PemToKeyPairTask.java
@@ -102,6 +102,8 @@ public class PemToKeyPairTask extends AsyncTask<Void, IOException, KeyPair>
         while(true){
             if(paused) {
                 continue;
+            } else if (isCancelled()) {
+                return null;
             } else {
                 for (PemToKeyPairConverter converter : converters) {
                     KeyPair keyPair = converter.convert(new String(pemFile));
@@ -154,6 +156,7 @@ public class PemToKeyPairTask extends AsyncTask<Void, IOException, KeyPair>
                     .onNegative(((dialog, which) -> {
                         dialog.dismiss();
                         toastOnParseError(result);
+                        cancel(true);
                     }));
 
             MaterialDialog dialog = builder.show();
--- a/app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/ssh/PemToKeyPairTask.java
+++ b/app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/ssh/PemToKeyPairTask.java
@@ -25,7 +25,6 @@ import android.net.Uri;
 import android.os.AsyncTask;
 import android.support.annotation.NonNull;
 import android.text.InputType;
-import android.util.Log;
 import android.view.View;
 import android.widget.EditText;
 import android.widget.Toast;
@@ -33,9 +32,9 @@ import android.widget.Toast;
 import com.afollestad.materialdialogs.DialogAction;
 import com.afollestad.materialdialogs.MaterialDialog;
 import com.amaze.filemanager.R;
+import com.amaze.filemanager.asynchronous.asynctasks.AsyncTaskResult;
 import com.amaze.filemanager.ui.views.WarnableTextInputLayout;
 import com.amaze.filemanager.ui.views.WarnableTextInputValidator;
-import com.amaze.filemanager.asynchronous.asynctasks.AsyncTaskResult;
 import com.amaze.filemanager.utils.application.AppConfig;
 import com.hierynomus.sshj.userauth.keyprovider.OpenSSHKeyV1KeyFile;
 
@@ -54,8 +53,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.StringReader;
 import java.security.KeyPair;
-import java.security.Provider;
-import java.security.Security;
 
 /**
  * {@link AsyncTask} to convert given {@link InputStream} into {@link KeyPair} which is requird by
@@ -68,7 +65,7 @@ import java.security.Security;
  * @see com.amaze.filemanager.filesystem.ssh.SshConnectionPool#create(Uri)
  * @see net.schmizz.sshj.SSHClient#authPublickey(String, KeyProvider...)
  */
-public class PemToKeyPairTask extends AsyncTask<Void, Void, AsyncTaskResult<KeyPair>>
+public class PemToKeyPairTask extends AsyncTask<Void, AsyncTaskResult<KeyPair>, AsyncTaskResult<KeyPair>>
 {
     private static final String TAG = "PemToKeyPairTask";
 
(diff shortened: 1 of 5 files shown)


TranceLove:
> Instead of having an entire thread do nothing while the user (finds and then) types the correct password seems a bit wasteful. Instead, having the AsyncTask end in an 'errored' state and restarting it when the user tries to check a new password would seem better. Is there something I'm not seeing?

This is certainly a better option - as now I realise AsyncTask should run only once and should use a new one instead of holding when there is no reason to hold (e.g. long running download/copy tasks), but this would require a bigger scale overhaul, because SshConnectionPool also depends on it. So treat this fix as a band-aid for the time being.

app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/ssh/PemToKeyPairTask.java
```diff
@@ -79,80 +76,61 @@
...
+    protected void onProgressUpdate(AsyncTaskResult<KeyPair>... values) {
         super.onProgressUpdate(values);
-    }
-
-    @Override
-    protected void onPostExecute(AsyncTaskResult<KeyPair> result) {
-        if(result.exception != null) {
-            if(dialog == null) {
-
+        if(values.length > 0){
```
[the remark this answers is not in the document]
TranceLove: Varargs is an array and does not have `isEmpty()`.

app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/ssh/PemToKeyPairTask.java
```diff
@@ -68,142 +65,132 @@
...
-            for(PemToKeyPairConverter converter : converters) {
-                KeyPair keyPair = converter.convert(new String(pemFile));
-                if(keyPair != null) {
-                    retval = new AsyncTaskResult<KeyPair>(keyPair);
-                    break;
+    protected KeyPair doInBackground(Void... voids) {
+        while(true){
+            if(isCancelled()) {
+                return null;
+            } else {
```
EmmanuelMess: No need for the `else`, as you would have returned.

app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/ssh/PemToKeyPairTask.java
```diff
@@ -68,142 +65,132 @@
...
+                if(paused) {
+                    continue;
+                } else {
```
EmmanuelMess: No need for the `else`, as you would have continued.

app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/ssh/PemToKeyPairTask.java
```diff
@@ -68,142 +65,132 @@
...
-                });
-            } else {
-                if(errorMessage != null) {
-                    WarnableTextInputLayout wilTextfield = (WarnableTextInputLayout)dialog.findViewById(R.id.singleedittext_warnabletextinputlayout);
-                    EditText textfield = (EditText)dialog.findViewById(R.id.singleedittext_input);
-                    wilTextfield.setError(errorMessage);
-                    textfield.selectAll();
+        if(values.length < 1) {
+            return;
+        } else {
```
EmmanuelMess: No need for `else` you would have returned.

app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/ssh/PemToKeyPairTask.java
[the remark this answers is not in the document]
EmmanuelMess: My mistake.

VishalNehra:
@TranceLove still failing to get it working after restarting Amaze. Like before, first run seems fine.
Log:
````
07-12 02:20:24.319 23840-25145/com.amaze.filemanager E/SshClientUtils: Error executing template method
    java.lang.RuntimeException: Unable to execute template
        at com.amaze.filemanager.filesystem.ssh.SshClientUtils.execute(SshClientUtils.java:76)
        at com.amaze.filemanager.filesystem.ssh.SshClientUtils.execute(SshClientUtils.java:127)
        at com.amaze.filemanager.filesystem.HybridFile.forEachChildrenFile(HybridFile.java:733)
        at com.amaze.filemanager.asynchronous.asynctasks.LoadFilesListTask.doInBackground(LoadFilesListTask.java:131)
        at com.amaze.filemanager.asynchronous.asynctasks.LoadFilesListTask.doInBackground(LoadFilesListTask.java:62)
        at android.os.AsyncTask$2.call(AsyncTask.java:333)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
        at java.lang.Thread.run(Thread.java:764)
````

TranceLove:
> @TranceLove still failing to get it working after restarting Amaze. Like before, first run seems fine.

Sorry but I couldn't reproduce the mentioned problem... on Oneplus One running Slim7 (7.1.2), Samsung Galaxy S2 running SlimLP (5.1.1), Pixel 2 emulator running 6.0. I did used a passphrase-protected key generated for my own (3128 bits RSA), but I also tested your key as well.

VishalNehra:
Will check again then.

VishalNehra:
My test still fails. What I found out is, from manually looking at databases, you never save password for the user in SFTP connection URL field. While in SMB, the encrypted password does get stored after username in Base64 form in url only. 
Please check below screenshots. 
screenshot_amaze_20180719-020627
screenshot_amaze_20180719-020539

TranceLove:
Here's mine... Oneplus One running Slim7 (7.1.2)

screenshot_20180719-111058

If you are using key authentication, no password will be saved.

And I suppose we had agreed that [key passphrases are never saved to database](https://github.com/TeamAmaze/AmazeFileManager/pull/863#issuecomment-354404166)?

VishalNehra:
@TranceLove not talking about saving key passphrases. As you can see in my screenshot, password for my username is not being saved. 
BTW I'm using a hyphen (-) in my password, if that helps.

TranceLove:
@EmmanuelMess please review again to make sure all requested changes had been applied.

PR #1361, opened 2018-07-11
Avoid duplicated code and make sure exception is always printed in PemToKeyPairTask

EmmanuelMess:


EmmanuelMess:
@TranceLove and have two methods called `convert()`?

TranceLove:
Oops, overlooked it, sorry.

PR #1362, opened 2018-07-11
AbstractProgressiveService.publishResults() and ProgressListener.onProgressed() simplified

EmmanuelMess:
A lot of parameters were actually not needed.

Needed for #1306.

Issue #1306, opened 2018-06-20
Compressing large file expanded notification flickers


EmmanuelMess:
If you compress 3GB file, the notification flickers before the actual compression starts, and "undefined" is shown for the size and time. I think we should show a "processing compression" notification before actually starting compression.

![](https://user-images.githubusercontent.com/10991116/42038736-156391a6-7ac2-11e8-9616-eaca6bcef5bd.gif)

VishalNehra:
It does show 'processing', I don't see the problem.
So it lasts only few seconds before the progress actually starts?

EmmanuelMess:
It depends on how large is the folder, in the gif it's 37s.

VishalNehra:
As I see it, the contents of notification is not a problem. 
Time remaining and transfer rate are indeed unknown at the time since transfer hasn't started yet.
So our main focus should be on the flickering.

EmmanuelMess:
>  Time remaining and transfer rate are indeed unknown at the time since transfer hasn't started yet.

But don't simply say 'unknown', remove the fields!

VishalNehra:
A big notification without things in it? Just empty? Won't look good imo.

EmmanuelMess:
> A big notification without things in it? Just empty? Won't look good imo.

Doesn't have to be big, I'll work on it ASAP.

PR #1363, opened 2018-07-11
ProgressHandler's fields are now atomically updated without synchronizing the class

EmmanuelMess:
* Moved ProgressHandler.speedRaw to local variable
* ProgressHandler's fields are now atomically updated without synchronizing the class

Required by #1362.
Needed for #1306.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/utils/ProgressHandler.java
+++ b/app/src/main/java/com/amaze/filemanager/utils/ProgressHandler.java
@@ -15,36 +15,38 @@ public class ProgressHandler {
 
     /**
      * total number of bytes to be processed
+     * Volatile because non volatile long r/w are not atomic (see Java Language Specification 17.7)
      */
-    private long totalSize = 0L;
+    private volatile long totalSize = 0L;
 
     /**
      * total bytes written in process so far
+     * Volatile because non volatile long r/w are not atomic (see Java Language Specification 17.7)
      */
-    private long writtenSize = 0L;
+    private volatile long writtenSize = 0L;
     /**
      * total number of source files to be processed
      */
-    private int sourceFiles = 0;
+    private volatile int sourceFiles = 0;
 
     /**
      * number of source files processed so far
      */
-    private int sourceFilesProcessed = 0;
+    private volatile int sourceFilesProcessed = 0;
     /**
      * file name currently being processed
      */
-    private String fileName;
+    private volatile String fileName;
 
     /**
      * boolean manages the lifecycle of service and whether it should be canceled
      */
-    private boolean isCancelled = false;
+    private volatile boolean isCancelled = false;
 
     /**
      * callback interface to interact with process viewer fragment and notification
      */
-    private ProgressListener progressListener;
+    private volatile ProgressListener progressListener;
 
     /**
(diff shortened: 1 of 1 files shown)


EmmanuelMess:
Another solution is to make the whole class immutable, and create a new object for each change, what do you think?

VishalNehra:
Immutable for this class is not ideal. It's updated multiple times every second. We'll be left with thousands of dangling pointers for GC to be collected at the end of each operation. 
Also, please use ````Atomic```` data types if you plan to remove synchronized from methods returning booleans/ints.

VishalNehra:
They were atomic until you removed synchronized from method signature.

EmmanuelMess:
> They were atomic until you removed synchronized from method signature.

[They ARE atomic.](https://stackoverflow.com/a/4756578/3124150)

EmmanuelMess:
> Immutable for this class is not ideal. It's updated  multiple times every second. We'll be left with thousands of dangling  pointers for GC to be collected at the end of each operation.

If I reduced the update rate, would immutability be a better idea?

VishalNehra:
This is not true... The assignment is not atomic, while an assignment operation is performed, it is modified in the memory, while assignment to memory if any other thread accesses it, older value might be returned. If we use Atomic classes, the whole assignment to memory part will be atomic hence only the most recent value will be returned by another thread.
We need either synchronised or Atomic/volatile variables.

VishalNehra:
Reduce the rate? What do you mean? Current code avoids any redundant code in operation so that it is the fastest possible. Anything more you write will effect the performance of the operation.
Why do you intend to fix something that's not broken in the first place?

EmmanuelMess:
> Anything more you write will effect the performance of the operation.

Thought this only communicated information back to the notification, my mistake.

EmmanuelMess:
> Reduce the rate? What do you mean? Current code avoids any redundant code in operation so that it is the fastest possible. Anything more you write will effect the performance of the operation.
Why do you intend to fix something that's not broken in the first place?

I'm not, I need to create more smaller PRs, and I've gotten ahead of myself suggesting the rate reduction. Sorry for the inconvenience.

EmmanuelMess:
[Fixed](https://docs.oracle.com/javase/tutorial/essential/concurrency/atomic.html).

> Reads and writes are atomic for all variables declared volatile (including long and double variables).

VishalNehra:
>  I'm not, I need to create more smaller PRs, and I've gotten ahead of myself suggesting the rate reduction. 

Please work on separate branch then, and create all smaller PRs to that branch. It is not recommended to push smaller test PRs in master.

EmmanuelMess:
> Please work on separate branch then, and create all smaller PRs to that branch. It is not recommended to push smaller test PRs in master.

This is by no means a "test PR" it is  part of a bigger change, but functional and useful by itself.

VishalNehra:
Yes, it's useful by itself that's why I approved. But if something is part of a bigger change, you should probably work on that bigger thing in a separate branch instead of creating PRs for master. That'll leave master unstable.

EmmanuelMess:
> That'll leave master unstable.
I just divide the bigger change into chunks that are stable, just as this is stable.

PR #1377, opened 2018-07-28
Prevent RootHelper.getFiles() returns duplicated entries

### LINKED ISSUE #1097: Root explorer file list has duplicated entries
After enabled root explorer, device root directory does not immediately visible, nor superuser permission dialog (SuperSU in my case) won't show up. Restarted app and go to `/system`, entries are doubled.

Go to storage that doesn't need root (`/storage/self/primary` for my Oneplus One + Slim7) directory/file listings are running normally.

Both entry of the same directory/file can open the corresponding directory/file however.

Verified on Oneplus One running Slim7 (7.1.2) and Samsung Galaxy S2 (GT-I9100) running Slimsaber (5.1.1), both rooted, running current master.

TranceLove:
Fixes #1097.

The fix was actually at the last 2 commits(7b53f05 and 2d33d99), while the first 2 were done while investigating the problem; I think they'd be useful anyway, so I kept them in the PR.

Robolectric test included, with Shadow emulating SuperSU's `Shell.Interactive`. I coded on an Ubuntu, so YMMV if you're running on a Mac.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/filesystem/RootHelper.java
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/RootHelper.java
@@ -29,6 +29,8 @@ import com.amaze.filemanager.utils.files.FileUtils;
 
 import java.io.File;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
 
 import eu.chainfire.libsuperuser.Shell;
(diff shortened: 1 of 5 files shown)


EmmanuelMess:
Why change the return type of `RootHelper.runShellCommand()`?

TranceLove:
> Why change the return type of RootHelper.runShellCommand()?

Oops, overlooked something. But I still think changing the return type to `List<String>` from is a better idea, unless you do need manipulating the underlying `ArrayList` directly using its API.

EmmanuelMess:
> But I still think changing the return type to List<String> is a better idea, unless you do need manipulating the underlying ArrayList directly using its API.

Remember someone going to modify the List will, now, probably copy it (`new ArrayList(runShellCommand())) instead of using it directly, I think not returning an ArrayList works best when you don't want the user to modify the List, for example for private fields in objects.

app/src/main/java/com/amaze/filemanager/filesystem/RootHelper.java
```diff
@@ -44,17 +44,13 @@
      * @return a list of results. Null only if the command passed is a blocking call or no output is
      * there for the command passed
      */
-    public static ArrayList<String> runShellCommand(String cmd) throws ShellNotRunningException {
+    public static List<String> runShellCommand(String cmd) throws ShellNotRunningException {
```
TranceLove: > Remember someone going to modify the List will, now, probably copy it (`new ArrayList(runShellCommand())) instead of using it directly, I think not returning an ArrayList works best when you don't want the user to modify the List, for example for private fields in objects.

So it's already here now...

EmmanuelMess: I don't understand what you are trying to say.

TranceLove: With `runShellCommand()` currently returns `List<String>`, please let me know if there is any changes I need to get a +1.

EmmanuelMess: See PR.

PR #1378, opened 2018-07-29
Change 'Storage' to 'Internal Storage' and updated icon accordingly

### LINKED ISSUE #1310: Change "Storage" to "Internal Storage"
Change "Storage" to "Internal Storage" to make it clear for the user, changing the icon might also help (like it does with USB OTG).

Mitch9378:
Change 'Storage' to 'Internal Storage' and updated icon accordingly. 
Fixes #1310


app/src/main/java/com/amaze/filemanager/ui/views/drawer/Drawer.java
```diff
@@ -1,3 +1,25 @@
+/*
+ * Drawer.java
+ *
+ * Copyright (C) 2014-2018 Mitch West<mitch9378dev@gmail.com> and Contributors.
```
TranceLove: > I am not sure if I handled this license correctly. Please let me know if I need to change anything.

Recommended to put other developer's name before yours.

For an example, see this:

https://github.com/TeamAmaze/AmazeFileManager/blob/b3197cf7f10a50574b7ddc478f7713fec996841f/app/src/main/java/com/amaze/filemanager/activities/MainActivity.java#L4-L5

Mitch9378: Thanks, I wasn't really sure who to put there but I will fix it.

EmmanuelMess:
What should we do to all the translations? How would a translator know that he has to update it? (Deleting all non-English translations is  a solution, although a bit drastic, and tedious).

EmmanuelMess:
Also, I don't think storage icon is the best for internal storage:
baseline_storage_black_18dp
In my opinion a phone would represent the thing better:
baseline_phone_android_black_18dp

Mitch9378:
@EmmanuelMess  That is a good point about the translations I was wondering that myself.  I don't know how else to handle the translation issue other than how you suggested. I am willing to remove all the other translations. I think the phone icon is a reasonable adjustment I will go ahead and make that change.

Mitch9378:
The bad translations are removed and the android phone icon is now being used instead of the storage icon.

EmmanuelMess:
Is there any reason you have used bitmaps now (.png)? Vector are smaller and less files (*.xml in this case)

Mitch9378:
@EmmanuelMess  No particular reason but as you stated it is more efficient. Updated and apologies.

TranceLove:
Verified working on Oneplus One running Slim7 (7.1.2) and Galaxy S2 running SlimLP (5.1.1). Adding requested license plate update, I'd vote a +1.

An exception case for the record: when I take this changeset into Nexus S emulator running 7.1.1, both internal storage and SD card were shown with the internal storage icon. This does not happen with Galaxy S2 running SlimLP (5.1.1), that external SD card is shown the SD card icon.

Because in reality there is no Nexus S running Android 7.1.1, so I'll just put this for the record; it should not affect this PR's status of Cleared For Merge.

> What should we do to all the translations? How would a translator know that he has to update it? (Deleting all non-English translations is a solution, although a bit drastic, and tedious).

This is indeed radical... so make sure you don't make this into 3.3.0 release.

PR #1389, opened 2018-08-06
Added small explanation for OTG device selection in SAF

### LINKED ISSUE #1279: OTG selecting a USB results in Amaze completely losing its Styling.
The phone Sony Xperia Z5 running 7.1.1

Local Storage
SDCard
OTG  (USB 32Gb plug in) 

Both the Local Storage and the SDCard operate fine in terms of navigation and the correct Amaze style, colours ans layout. 

However selecting OTG from the slide out menu, displays what looks like the default Android explorer. Just black and white, squares for folder and files.
A message does flash up for a moment saying "Select the Root of OTG device"  No idea what it is trying to say here or do. 

Bottom line, Amaze does not operate correctly on OTG. But surely this would have been discovered a long while ago ?  
Am I doing something wrong ?

EmmanuelMess:
Added small explanation for OTG device selection in SAF

Fixes #1279.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/ui/dialogs/GeneralDialogCreation.java
+++ b/app/src/main/java/com/amaze/filemanager/ui/dialogs/GeneralDialogCreation.java
@@ -1147,4 +1147,9 @@ public class GeneralDialogCreation {
         a.show();
     }
 
+    public static MaterialDialog showOtgSafExplanationDialog(ThemedActivity themedActivity) {
+        return GeneralDialogCreation.showBasicDialog(themedActivity, R.string.saf_otg_explanation,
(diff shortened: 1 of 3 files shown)


VishalNehra:
Do we really need a seperate OTGDialog class just for one function?

EmmanuelMess:
> Do we really need a seperate OTGDialog class just for one function?

I think that dividing things this way produces less collisions between PRs in the future. In my opinion more files and less lines per file is better. Let me know if you think I went a bit over here.

VishalNehra:
Yes. I don't think OTG will have much dialoges to have a separate class for it.

PR #1390, opened 2018-08-06
Reverted default storage icon to sd icon

### LINKED ISSUE #1388: External storage has phone icon
External storage has phone icon, instead of sd icon. My mistake when testing.

EmmanuelMess:
Fixes #1388.


TranceLove:
So it indeed happened...

Verified working with Sony Z1 Compact running CM13 (6.0). I'd vote a +1.

PR #1391, opened 2018-08-07
Bill Pugh Singleton Implementation

shellhub:


shellhub:
Prior to Java 5, java memory model had a lot of issues and above approaches
used to fail in certain scenarios where too many threads try to get the
instance of the Singleton class simultaneously

TranceLove:
For everyone's reference:

https://dzone.com/articles/singleton-bill-pugh-solution-or-enum

But talking about coding idiom/practices I think I'll just let @EmmanuelMess to decide... ;)

shellhub:
Prior to Java 5, java memory model had a lot of issues and above approaches used to fail in certain scenarios where too many threads try to get the instance of the Singleton class simultaneously. So Bill Pugh came up with a different approach to create the Singleton class using a inner static helper class.

@TranceLove @EmmanuelMess

app/src/main/java/com/amaze/filemanager/utils/DataUtils.java
```diff
@@ -73,13 +73,16 @@
 
     private DataChangeListener dataChangeListener;
 
-    private static DataUtils sDataUtils;
+    private DataUtils(){
+
+    }
+
+    private static class SingletonHelper{
```
EmmanuelMess: This should be `DataUtilsHolder`.

EmmanuelMess:
> **Bill Pugh singleton**: implementation using the static inner helper class. Static inner classes are not loaded into memory until their getInstance() methods are called
> [...]
> A Bill Pugh Singleton is based on the “initialization on demand holder” idiom. This idiom uses inner classes and does not use any synchronization construct. It uses static blocks, but in a different way — and suggests using static inner classes.

This seems OK now, but the holder should be `DataUtilsHolder`, @VishalNehra what do you think?

shellhub:
yeah, rename `DataUtils.SingletonHelper -> DataUtils.DataUtilsHolder` is awesome @EmmanuelMess

PR #1412, opened 2018-08-19
Removed and renamed methods in AppConfig

EmmanuelMess:
* Removed RequestQueue redundant methods from AppConfig:
  * `addToRequestQueue(Request<T> req, String tag)`
  * `addToRequestQueue(Request<T> req)`
  * `cancelPendingRequests(Object tag)`
  * `getImageLoader()`
* MainActivity reference in AppConfig is now a WeakReference
*  Renamed AppConfig.getActivityContext() to getMainActivityContext() to reflect real context

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/activities/MainActivity.java
+++ b/app/src/main/java/com/amaze/filemanager/activities/MainActivity.java
@@ -24,7 +24,6 @@ package com.amaze.filemanager.activities;
 
 import android.app.Activity;
 import android.app.ActivityManager;
-import android.app.Application;
 import android.content.BroadcastReceiver;
 import android.content.ContentResolver;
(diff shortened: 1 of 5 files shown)


EmmanuelMess:
> verified no regression

You sure? your code makes extensive use of the AppConfig Context, and now it isn't referenced strongly.

TranceLove:
> You sure? your code makes extensive use of the AppConfig Context, and now it isn't referenced strongly.

So you have a better idea? Actually I also ran the tests (Espresso and Robolectric).

SSH related tasks are mostly short-lived. Most probable scenario I can think of is the app left idle for long and weak references are being reclaimed.

EmmanuelMess:
> So you have a better idea?

Nope, just wanted to make sure.

PR #1414, opened 2018-08-19
Optimized and better naming for RootHelper.runShellCommand() functions and added one that simply runs command

EmmanuelMess:


EmmanuelMess:
@VishalNehra please check this pr.

PR #1415, opened 2018-08-20
Removed unused PropertiesSheet

EmmanuelMess:


EmmanuelMess:
> Voting +1, unless there is good argument keeping this (will continue develop?)

Maybe branch and remove this commit from the new branch?

TranceLove:
@VishalNehra this sounds like a new feature which will take quite some time to complete. I'll then vote +1 for moving to separate branch since it's not ready yet.

(Although, 3.4.0 is still far away from here)

EmmanuelMess:
> Maybe someone else wanna try before actually removing all this?

Anyone can try: in another branch. You can create a patch and add it to the corresponding issue if you want.

PR #1418, opened 2018-08-20
Replace never-set CloudStreamSource.mime with constant null

EmmanuelMess:


EmmanuelMess:
I am thinking now that this could lead to someone thinking that `getMimeType()` actually returns a mime type, maybe better to remove the method and hardcode null where it was used?

TranceLove:
One of the way out. When some third-party API actually requires the MIME type of the stream we can add it back.

EmmanuelMess:
> I am thinking now that this could lead to someone thinking that `getMimeType()` actually returns a mime type, maybe better to remove the method and hardcode null where it was used?

Done.

PR #1420, opened 2018-08-24
Replaced switch with abstract  getTitle() in AbstractProgressiveService

EmmanuelMess:


app/src/main/java/com/amaze/filemanager/asynchronous/services/EncryptService.java
```diff
@@ -143,6 +144,12 @@ protected void setPercentProgress(float progress) {
         this.progressPercent = progress;
     }
 
+    @Override
+    @StringRes
+    protected int getTitle(boolean move) {
```
TranceLove: Recommend to rename `move` to `decrypting` to be more meaningful.

EmmanuelMess: This doesn't test if "decripting", it just tests if moving (because copy and move operations have the same service). I would think that move here is always false, because `DecryptService` exists. Will check.

EmmanuelMess: Done

PR #1430, opened 2018-08-30
FlexibleLayout for HiddenAdapter

EmmanuelMess:
This is a pilot test for FlexibleAdapter lib

* HiddenAdapter to FlexibleAdapter
* Completely separated hidden and history list
* History and hidden items migrated to ConstraintLayout
* Made more small aesthetic changes


TranceLove:
Tested on Oneplus 2 running Slim7 (7.1.2). If users don't mind an extra line between the items...

Before:
screenshot_20181226-150800

After:
screenshot_20181226-150851

EmmanuelMess:
>  If users don't mind an extra line between the items...

I was just trying to replicate previous functionality, I will add the line ASAP.

TranceLove:
Personally I'd prefer no lines between the items in the list, i.e. the "before" look... @VishalNehra ?

VishalNehra:
I think there should be adequate margins between items in the ' before' look. Something from material dimensions. Otherwise, there shouldn't be any margin between the title and it's path like in the 'before' look.

EmmanuelMess:
I would follow the two line list item [guideline from Material Design](https://material.io/design/components/lists.html#specs) (first example):



is that okay?

TranceLove:
> I would follow the two line list item guideline from Material Design (first example)

I prefer the second example. The spacing between the items in the first example is taking too much space IMO.

EmmanuelMess:
Removed the line, redesigned and used marquee for path:
screenshot_1546274236

VishalNehra:
Seems good. I'll have to try on my device.

VishalNehra:
Title are very very small on my device. Maybe it was better to follow the first example. Also, a selectable background with ripple effect would help with a better UX.

EmmanuelMess:
> Maybe it was better to follow the first example. 

will move to first example ASAP

PR #1435, opened 2018-09-18
can't auto refresh list

shellhub:
I must pull down to refresh to refresh.  
* Request storage permission dialog appear firstly. I `DENY`
* Grant permission dialog apper. I `GRANT`
* Request storage permission dialog appear again. I `ALLOW`

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/activities/superclasses/PermissionsActivity.java
+++ b/app/src/main/java/com/amaze/filemanager/activities/superclasses/PermissionsActivity.java
@@ -28,18 +28,18 @@ public class PermissionsActivity extends ThemedActivity
         if (requestCode == STORAGE_PERMISSION) {
             if (isGranted(grantResults)) {
                 permissionCallbacks[STORAGE_PERMISSION].onPermissionGranted();
+                permissionCallbacks[STORAGE_PERMISSION] = null;
             } else {
                 Toast.makeText(this, R.string.grantfailed, Toast.LENGTH_SHORT).show();
                 requestStoragePermission(permissionCallbacks[STORAGE_PERMISSION]);
             }
 
-            permissionCallbacks[STORAGE_PERMISSION] = null;
         } else if (requestCode == INSTALL_APK_PERMISSION) {
             if (isGranted(grantResults)) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/amaze/filemanager/activities/superclasses/PermissionsActivity.java
```diff
@@ -102,4 +102,11 @@ private boolean isGranted(int[] grantResults) {
         void onPermissionGranted();
     }
 
+    @Override
+    protected void onDestroy() {
+        super.onDestroy();
+        for (int i = 0; i < permissionCallbacks.length; i++) {
+            permissionCallbacks[i] = null; // clear to let GC do its work
```
EmmanuelMess: This is redundant, the GC will clear automatically when necessary.

shellhub: From `java.util.Arraylist.remove` source code . 
``` java
    /**
     * Removes the element at the specified position in this list.
     * Shifts any subsequent elements to the left (subtracts one from their
     * indices).
     *
     * @param index the index of the element to be removed
     * @return the element that was removed from the list
     * @throws IndexOutOfBoundsException {@inheritDoc}
     */
    public E remove(int index) {
        if (index >= size)
            throw new IndexOutOfBoundsException(outOfBoundsMsg(index));

        modCount++;
        E oldValue = (E) elementData[index];

        int numMoved = size - index - 1;
        if (numMoved > 0)
            System.arraycopy(elementData, index+1, elementData, index,
                             numMoved);
        elementData[--size] = null; // clear to let GC do its work

        return oldValue;
    }
```
I think `elementData[--size] = null; // clear to let GC do its work` statement is needed

shellhub:
@EmmanuelMess You are right. we don't need this code. GC will do for us.
``` java

    @Override
    protected void onDestroy() {
        super.onDestroy();
        for (int i = 0; i < permissionCallbacks.length; i++) {
            permissionCallbacks[i] = null; // clear to let GC do its work
        }
    }
```
Thanks @JakeWharton relpy
> In that case, no, setting to null is not needed. The activity is dead and is itself going to be collected. Settings things inside of it to null doesn't change anything.

PR #1437, opened 2018-09-22
make COLUMN_PATH UNIQUE

shellhub:
add `UNIQUE` to avoid duplicate files 
screenshot_1537583011

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/activities/superclasses/PermissionsActivity.java
+++ b/app/src/main/java/com/amaze/filemanager/activities/superclasses/PermissionsActivity.java
@@ -28,18 +28,18 @@ public class PermissionsActivity extends ThemedActivity
         if (requestCode == STORAGE_PERMISSION) {
             if (isGranted(grantResults)) {
                 permissionCallbacks[STORAGE_PERMISSION].onPermissionGranted();
+                permissionCallbacks[STORAGE_PERMISSION] = null;
             } else {
                 Toast.makeText(this, R.string.grantfailed, Toast.LENGTH_SHORT).show();
                 requestStoragePermission(permissionCallbacks[STORAGE_PERMISSION]);
             }
 
-            permissionCallbacks[STORAGE_PERMISSION] = null;
         } else if (requestCode == INSTALL_APK_PERMISSION) {
             if (isGranted(grantResults)) {
                 permissionCallbacks[INSTALL_APK_PERMISSION].onPermissionGranted();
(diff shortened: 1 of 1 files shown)


shellhub:
@EmmanuelMess @TranceLove I update code. please review

shellhub:
how to code?

EmmanuelMess:
Ask Google.

shellhub:
I don't know what's your meaning! I write the code following [android official docs](https://developer.android.com/training/data-storage/sqlite#java) . we need drop table in `onUpgrade` method and  execute `onCreate`

TranceLove:
End users won't like their data get lost whenever an update happens - no matter it's a website with millions of users or a little file manager app in your Android phone. And I think sqlite, like the other common RDBMS engines in the rest of the world, supports adding `UNIQUE` constraints in a `ALTER TABLE` SQL statement.

For a little love, this is where your `ALTER TABLE` statement should go to.
https://github.com/TeamAmaze/AmazeFileManager/blob/5c8ae0b395056b3262ea19b824713fc3192a8130/app/src/main/java/com/amaze/filemanager/database/UtilsHandler.java#L112

EmmanuelMess:
> I don't know what's your meaning! I write the code following [android official docs](https://developer.android.com/training/data-storage/sqlite#java) . we need drop table in `onUpgrade` method and execute `onCreate`

Google docs say:

> // This database is only a cache for online data, so its upgrade policy is
> // to simply to discard the data and start over

shellhub:
@EmmanuelMess @TranceLove Thank for your review.  `SQLite` don't support adding constraint after table created.  we will have to save the existing data to a temporary table, drop the old table, create the new table with the `UNIQUE` constraint, then copy the data back in from the temporary table.  But this will write a lots code. can you give any suggestion?

> https://stackoverflow.com/questions/1884818/how-do-i-add-a-foreign-key-to-an-existing-sqlite-table

EmmanuelMess:
> But this will write a lots code. can you give any suggestion?

Not much code, SQLite has a way to create temporary tables relatively easily. See [here](https://github.com/EmmanuelMess/Simple-Accounting/blob/f3b25ed6ef567ab28b8ab3d265a37675f38937ab/SimpleAccounting/app/src/main/java/com/emmanuelmess/simpleaccounting/db/TableGeneral.java#L54).

VishalNehra:
I agree, we can't afford to drop tables just because we update the table structure. There are important passwords for network connection/encryption being saved and not just other preferences.

Issue #1337, opened 2018-07-03
FTP Server don't start


VishalNehra:
FTP server is not getting started. I can reproduce on my device, also [reported by a beta tester](https://plus.google.com/+VishalNehra/posts/2KzfRZ3GHqM) on G+
Will post logs later.

sid32:
Also, would be nice if we could write to device over FTP. Solid Explorer lets you do that.

wired2051:
I also can't start FTP Server anymore - and it definitely worked. 
Device: Huawei P9 lite, Android 7.0, EMUI 5.0.1

EmmanuelMess:
@VishalNehra, or anyone experiencing the issue, please post logs.

wired2051:
> please post logs

I'd be glad to but tell me how.

EmmanuelMess:
[This](https://play.google.com/store/apps/details?id=com.dp.logcatapp) logcat reader would seem good enough, please post logs from Amaze as text here.

VishalNehra:
I think this can be closed for now.

sid32:
> I think this can be closed for now.

FTP works in latest builds, but those of us who use Tasker/Macrodroid need you to publish updated Intents instructions for us to start the FTP with Intents in Oreo.

wired2051:
For me it still doesn't work. I use Amaze 3.2.2 (latest version for Android 7 in F-Droid). My phone is a Huawei P9 lite working with Android 7.0, EMUI 5.0.4 (non-rooted).

I don't know what Tasker/Macrodroid means.

VishalNehra:
@sid32 I think intents are still the same. @TranceLove can you please confirm if you changed the intents?

VishalNehra:
@wired2051 the latest builds are not on fdroid. You can try them by signing up on play store beta program.

sid32:
Im using the same intents as before and while the FTP now works in Orero, my intents do not start it. I think I might need to add the Package and Class intent properties, but I don't know what they are.

PR #1439, opened 2018-09-24
Explicitly specify package name in FTP server related intents

TranceLove:
This fixes FTP server cannot start on Android 8 due to implicit intent broadcasts are prohibited from Android 8 (as mentioned [here](https://github.com/TeamAmaze/AmazeFileManager/issues/1337#issuecomment-423780217)).

Additionally, also fixed a display glitch that status message at `FtpServerFragment` is not in the activity's accent colour when FTP server is started from the tile (which is then becomes plain black).

Tested with non-rooted LG Nexus 5x running stock firmware 8.1.0.

Partially related to #1337, but may not be fixing it.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/asynchronous/services/ftp/FtpService.java
+++ b/app/src/main/java/com/amaze/filemanager/asynchronous/services/ftp/FtpService.java
@@ -213,9 +213,9 @@ public class FtpService extends Service implements Runnable {
         try {
             server = serverFactory.createServer();
             server.start();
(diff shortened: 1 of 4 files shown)


VishalNehra:
Working fine running on n6p oreo
Not sure about nougat versions as I couldn't check.

TranceLove:
Two ticks, merging.

EmmanuelMess:
'tis a fix, cherrypicking to release-3.3

TranceLove:
For reference, with current master including this changeset (5c8ae0b) FTP server can be started/stopped from either FtpServerFragment and FTP server Tile on rooted Oneplus 2 running Slim7 (7.1.2).

PR #1440, opened 2018-09-24
Check if screen is locked before processing tile's click event (#1253)

### LINKED ISSUE #1253: Forbid FTP startup when device is locked
I can currently enable FTP from the lock screen with the notification shortcut. I think that is a potential security risk. It should ask for the pin/pattern like when you try to turn off geolocation.

Also, it'd be great if the button was greyed out when the server is off, like other buttons do.

screenshot_20180527-121401

TranceLove:
Fixes #1253.

Check if screen is locked before processing FTP server tile's click event.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/asynchronous/services/ftp/FtpTileService.java
+++ b/app/src/main/java/com/amaze/filemanager/asynchronous/services/ftp/FtpTileService.java
@@ -46,19 +46,19 @@ public class FtpTileService extends TileService {
 
     @Override
     public void onClick() {
-        super.onClick();
-
-        if (getQsTile().getState() == Tile.STATE_ACTIVE) {
-            getApplicationContext().sendBroadcast(new Intent(FtpService.ACTION_STOP_FTPSERVER).setPackage(getPackageName()));
-        } else {
-            if (FtpService.isConnectedToWifi(getApplicationContext())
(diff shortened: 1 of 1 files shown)


EmmanuelMess:
This seems untestable until #1439 is merged.

VishalNehra:
Unable to start FTP server from tile after applying this patch manually to test.

VishalNehra:
Unable to start FTP server from tile (even when device unlocked) after applying this patch manually to test on top of #1439 on my /n6p running oreo.

TranceLove:
Care to try again with bc4e491? Works with Oneplus 2 running Slim7 (7.1.2) and LG Nexus 5X running stock firmware (8.1.0).

And sorry to say Robolectric unit tests are not possible since they don't have an implementation of the underlying `IQSService`.

VishalNehra:
Working fine now. Although I suggest use of ````unlockAndRun```` instead of not doing anything on lockscreen. Otherwise it leaves an impression that there's something wrong with the service.

TranceLove:
>  I suggest use of unlockAndRun instead of not doing anything on lockscreen. 

See 441692d. However, if device uses swipe to unlock screen, FTP server tile is still working. Tested with LG Nexus 5X running stock firmware (8.1.0).

VishalNehra:
> However, if device uses swipe to unlock screen, FTP server tile is still working.

As far as it don't have any pin/pattern it's fine.

PR #1452, opened 2018-10-01
Show less history.

tumuyan:
Remove duplicate path. Remove "/" and "/*" path. Retain "/*/*" path.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/utils/DataUtils.java
+++ b/app/src/main/java/com/amaze/filemanager/utils/DataUtils.java
@@ -291,7 +291,12 @@ public class DataUtils {
     }
 
     public void addHistoryFile(final String i) {
-        history.push(i);
+        String d=i.replaceFirst("/+[^/]*","");
+        if(d.length()>1){
+            history.remove(i);
+            history.push(i);
+        }
+
         if (dataChangeListener != null) {
             dataChangeListener.onHistoryAdded(i);
         }


app/src/main/java/com/amaze/filemanager/utils/DataUtils.java
```diff
@@ -291,7 +291,12 @@ public void setHistory(LinkedList<String> s) {
     }
 
     public void addHistoryFile(final String i) {
-        history.push(i);
+        String d=i.replaceFirst("/+[^/]*","");
+        if(d.length()>1){
```
EmmanuelMess: It is more readable to write `!d.isEmpty()`

app/src/main/java/com/amaze/filemanager/utils/DataUtils.java
```diff
@@ -291,7 +291,12 @@ public void setHistory(LinkedList<String> s) {
     }
 
     public void addHistoryFile(final String i) {
-        history.push(i);
+        String d=i.replaceFirst("/+[^/]*","");
```
EmmanuelMess: Why are you doing this?

tumuyan: I never find a file in  / or / */  directory. 
also, opening from / directory list is very fast. 
I think it is a waste of screen and attention.

EmmanuelMess: I don't think removing valid folders from history is a good idea. I think it as a browser history, you don't remove link just because they are easily accessible...

PR #1461, opened 2018-10-10
Background Colors Changed to White (#FFFFFF)

ThiruPraveen18:


VishalNehra:
We're handling background colors programatically. This is not necessary and will create inconsistent results.

PR #1463, opened 2018-10-11
Fix empty icon view when image icon is broken

### LINKED ISSUE #1338: Broken images don't show an icon untill they are selected
Images that are broken don't show any icon until they are selected. I believe [this](https://github.com/TeamAmaze/AmazeFileManager/commit/0a63ee2f91321cdaf23b3835e15d07a99bc7ace3#diff-40ce0369d0ba588ea75b0ddf1fc8559cL169) was a workaround for that bug.

VishalNehra:
Fixes #1338

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/adapters/RecyclerAdapter.java
+++ b/app/src/main/java/com/amaze/filemanager/adapters/RecyclerAdapter.java
@@ -142,7 +142,7 @@ public class RecyclerAdapter extends RecyclerView.Adapter<RecyclerView.ViewHolde
 
             Animation iconAnimation = AnimationUtils.loadAnimation(context, R.anim.check_out);
             if (imageView != null) {
-                imageView.setAnimation(iconAnimation);
+                imageView.startAnimation(iconAnimation);
             } else {
                 // TODO: we don't have the check icon object probably because of config change
             }
@@ -152,7 +152,7 @@ public class RecyclerAdapter extends RecyclerView.Adapter<RecyclerView.ViewHolde
 
             Animation iconAnimation = AnimationUtils.loadAnimation(context, R.anim.check_in);
             if (imageView != null) {
-                imageView.setAnimation(iconAnimation);
+                imageView.startAnimation(iconAnimation);
             } else {
                 // TODO: we don't have the check icon object probably because of config change
             }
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/amaze/filemanager/adapters/RecyclerAdapter.java
```diff
@@ -540,9 +540,9 @@ public void onBindViewHolder(final RecyclerView.ViewHolder vholder, int p) {
                 });
 
                 // resetting icons visibility
-                holder.genericIcon.setVisibility(View.GONE);
-                holder.pictureIcon.setVisibility(View.GONE);
-                holder.apkIcon.setVisibility(View.GONE);
+                holder.genericIcon.setVisibility(View.VISIBLE);
+                holder.pictureIcon.setVisibility(View.INVISIBLE);
+                holder.apkIcon.setVisibility(View.INVISIBLE);
```
EmmanuelMess: These seem to be in the same `FrameLayout` (in list at least) which seems that would make problematic having views invisible but still taking space.

app/src/main/java/com/amaze/filemanager/adapters/RecyclerAdapter.java
```diff
@@ -834,6 +834,7 @@ private void showThumbnailWithBackground(ItemViewHolder viewHolder, IconDataParc
             @Override
             public boolean onLoadFailed(@Nullable GlideException e, Object model, Target target, boolean isFirstResource) {
                 new Handler(msg -> {
+                    viewHolder.genericIcon.setVisibility(View.VISIBLE);
```
EmmanuelMess: This seems redundant, it is done in line 828.

EmmanuelMess:
PS:
I was ill and working on projects for school for most of this week, so I wasn't able to look at any GitHub notification...and they accumulated quite a bit. I'm able to work now, so I apologize for the lateness.

app/src/main/java/com/amaze/filemanager/adapters/RecyclerAdapter.java
VishalNehra: Nope, doesn't work without this. Can try.

app/src/main/java/com/amaze/filemanager/adapters/RecyclerAdapter.java
VishalNehra: This is the best solution I came up with. We can always debug the whole layout but I didn't have the time.

EmmanuelMess:
I am seeing mixed results, with a folder with ~288 images and one 0B jpg, the jpg appears transparent even with this PR, but other 0B jpgs work.

If @TranceLove can replicate

> In my case, invalid PNG, JPEG and MP4 files are shown with broken icon but not GIF file, which is #1338's symptom.

with lastest master but not with this PR I'll accept.

VishalNehra:
@EmmanuelMess he already approved.

EmmanuelMess:
@VishalNehra Yes, but I want to make sure that the problem is still replicable in lastest master.

TranceLove:
Checked with current master bf01a9e with Oneplus 2 running Slim7 (7.1.2), using the same test method above.

The first time I get into the folder

screenshot_20181017-084756

after selecting

screenshot_20181017-084811

screenshot_20181017-084818

VishalNehra:
So, we good? We merge?

Issue #1457, opened 2018-10-04
The option to show the time of folders\files


AnonJump:
I would like to see in the program an option that, when turned on, will display the time (hours and minutes) for the folder / file change. For example, similar is implemented in the Root Explorer.

Now in Amaze File Manager:
"Folder name"
Month, year

In the possible future:
"Folder name"
Month, year; time:minutes

mk410gd:
@EmmanuelMess Can I take this issue?

EmmanuelMess:
First I think we need to deal with the fact that there are two dates in the life of the folder, the creation time and the modification time (not taking into account metadata change dates here).

I think the user is more likely to think that the time shown is the creation date, how would the user know it's the modification time that's being shown?

EmmanuelMess:
Related #1028

mk410gd:
@EmmanuelMess I think that actual "Show last modified" in the Setttings could be divided into "Show last modified Date" and "Show last modified Time". "Show last modified Time" would be default set to OFF and the user could change it at any time.
So if we applied it, the user would know he set the modified time display.

AnonJump:
I wonder if it is worth adding _seconds_ to the option (hours, time, seconds)?

EmmanuelMess:
@mk410gd we can't rely on the user reading that specific settings message for to know that he is reading the modification time.

EmmanuelMess:
@AnonJump when would adding seconds be useful?

EmmanuelMess:
> @EmmanuelMess I think the date and time shown for the files / folders, is obvious to most users. Modification time is used to apply everywhere and in similar applications.

I still think that we should explicitly declare that it is the modification date, and not creation date. But I am unsure if the extra "cloggedness" would really be necessary.

@TranceLove @VishalNehra what do you say?

TranceLove:
> I still think that we should explicitly declare that it is the modification date, and not creation date. But I am unsure if the extra "cloggedness" would really be necessary.

With my computing experience with Windows and Linux (GNOME, Cinnamon) I'm used to read the datetime beside the file in file managers that it is the file's last modified time. Adding the limited screen sizes of the majority of Android devices, I don't think it's a good idea to explicitly state the datetime is the file's last modification time.

Fireant666:
Of course users will know & understand that it's referring to the modification date. What a crock of shit to suggest otherwise.

PR #1466, opened 2018-10-12
Add option to show the time modification of folders\files

mk410gd:
In Settings  has been added option to show the time modification of folders\files in the list of items (issue #1457)

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/adapters/data/LayoutElementParcelable.java
+++ b/app/src/main/java/com/amaze/filemanager/adapters/data/LayoutElementParcelable.java
@@ -46,6 +46,7 @@ public class LayoutElementParcelable implements Parcelable {
     public final boolean isDirectory;
     public final long date, longSize;
     public final String date1;
+    public final String date2;
     public final boolean header;
 
     //same as hfile.modes but different than openmode in Main.java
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/amaze/filemanager/adapters/data/LayoutElementParcelable.java
```diff
@@ -141,10 +141,10 @@ public LayoutElementParcelable(Parcel im) {
         symlink = im.readString();
         int j = im.readInt();
         date = im.readLong();
+        dateModification = im.readString();
```
EmmanuelMess: This line is in the wrong position, I think, the `Parcel` must be read in the same order it is written.

app/src/main/res/xml/preferences.xml
```diff
@@ -84,7 +84,7 @@
             android:key="showHidden"
             android:title="@string/hidden"/>
         <com.amaze.filemanager.ui.views.preference.CheckBox
-            android:defaultValue="true"
+            android:defaultValue="false"
```
EmmanuelMess: Why did you change the default value?

mk410gd: my fault, default value was changed

mk410gd:
@EmmanuelMess added the same functionality to the compressed files

PR #1469, opened 2018-10-13
Fix renaming files on OTG storage.

### LINKED ISSUE #1419: Cannot rename file on OTG storage
Good Day,

When I attempt to rename a file located on a OTG connected drive, the app crashes.  However if I rename the same file on the local storage, then simply copy it over to the OTG connected storage with the changed name it works fine. (So the file's fine)

Android 6.0.1 using Amaze 3.2.2 (same issue on previous versions)

Cheers and much thanks,

tryton-vanmeer:
Operations.rename would check if the new file name is valid. OTGUtil.getDocumentFile
would result in a null pointer becuase the new file doesn't exist in the document tree.
In the case of OTG, HybridFile.getName now calls the new OTGUtil.getName method.

Fixes #1419

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/filesystem/EditableFileAbstraction.java
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/EditableFileAbstraction.java
@@ -68,7 +68,7 @@ public class EditableFileAbstraction {
                 path = Utils.sanitizeInput(path);
                 this.hybridFileParcelable = new HybridFileParcelable(path);
 
-                String tempN = hybridFileParcelable.getName(context);
+                String tempN = hybridFileParcelable.getName();
                 if(tempN == null) tempN = uri.getLastPathSegment();
                 this.name = tempN;
 
--- a/app/src/main/java/com/amaze/filemanager/filesystem/HybridFile.java
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/HybridFile.java
@@ -335,7 +335,7 @@ public class HybridFile {
             case ROOT:
                 return new File(path).getName();
             case OTG:
-                return OTGUtil.getDocumentFile(path, context, false).getName();
+                return OTGUtil.getName(path);
             default:
                 StringBuilder builder = new StringBuilder(path);
                 name = builder.substring(builder.lastIndexOf("/")+1, builder.length());
--- a/app/src/main/java/com/amaze/filemanager/filesystem/HybridFile.java
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/HybridFile.java
@@ -300,9 +300,6 @@ public class HybridFile {
         return path;
     }
 
-    /**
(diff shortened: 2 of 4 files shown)


EmmanuelMess:
This seems to needlessly reinvent the wheel, `getName()` already exists. It seems better to fix the fact that `HybridFile.getName()` is getting called.

tryton-vanmeer:
Fair point. 

I now also realized I overlooked the depreciated `getName()` method in `HybridFile`. Since the name of the OTG file can be obtained the same way as the `case FILE` or `default` block, would it be better to remove `getName(Context context)` and refactor to use the old `getName()` method (with an OTG case added of course).

If that's a more satisfactory solution, I can update this PR with those changes in place.

EmmanuelMess:
It seems that we have 2 problems:
1. The rename can change the whole path, for example "otg://a.txt" to "otg://m/f.yml" and we have to check that the new path is correct (not just the filename).
2. The current check (the `Operations.isFileNameValid(String)` function) doesn't do an OTG (or any OpenMode) specific check, just checks for some characters in the String passed.

To me it seems that there are two viable solutions:
A. Pass the filename to `Operations.rename()` (or just do `new File(path).getName()` as you proposed) and check the filename with `Operations.isFileNameValid(String)`.
B. Rewrite `Operations.isFileNameValid()` to check the whole path in a specific way for each `OpenMode` (checks will probably overlap a lot because most openmodes are using the same filesystem).

Solution A seems very hacky because of problem 1. But solution B will take some time.

VishalNehra:
> The rename can change the whole path, for example "otg://a.txt" to "otg://m/f.yml" and we have to check that the new path is correct (not just the filename).

I don't think our interface allows renaming of whole path.
As for the PR, I haven't checked the cause of crash, but I think there is already such generic function ````(getName#path)```` to extract name of file, so no need for a declaring it again.

EmmanuelMess:
> > The rename can change the whole path, for example "otg://a.txt" to "otg://m/f.yml" and we have to check that the new path is correct (not just the filename).
> 
> I don't think our interface allows renaming of whole path.

Move by renaming was something that was added a while ago, wasn't it added to OTG too?

> As for the PR, I haven't checked the cause of crash, but I think there is already such generic function `(getName#path)` to extract name of file, so no need for a declaring it again.

NPE when calling `getName` was the cause of the crash.

VishalNehra:
> Move by renaming was something that was added a while ago, wasn't it added to OTG too?

It was, but not supported for OTG devices, it's just for local file system and other which actually are supported by their respective APIs
Hence Amaze doesn't support it either. 
See ````MoveFiles#156````

VishalNehra:
I don't have an OTG cable. Guys if possible, please sort this out before next release.

EmmanuelMess:
@tryton-vanmeer you need help, how's your schedule?

EmmanuelMess:
This seems to have been fixed already.

PR #1486, opened 2018-10-28
Some fixes

EmmanuelMess:
* Fix WeakReference in CloudLoaderAsyncTask
* MainActivity.mCloudCursorData -> cloudCursorData
* Fix hack comments
* Create and use new AppConfig.toast() that takes StringRes


EmmanuelMess:
This doesn't need two checks as it goes into another PR.

Issue #1485, opened 2018-10-24
Improve handling of new-style OpenSSH key files


hvdijk:
When using the new SFTP functionality (looking very nice, by the way!), when attempting to load a private key in the new-ish OpenSSH format (the type that begins with `-----BEGIN OPENSSH PRIVATE KEY-----`), the key will not load. A prompt is shown asking for a passphrase, but no matter what passphrase is entered, it will not be accepted.

This new private key format has been made the default as of OpenSSH 7.8, and as I created a new key for use on my phone, I inadvertently created one in this new format. I can understand that it may not be feasible to implement support for this format in a short time, but would it be possible to give a more useful error message?

VishalNehra:
Exactly the problem I had. @TranceLove can you please give this a look.

TranceLove:
According to https://github.com/hierynomus/sshj/issues/276, support is already there.

I'm also submitting a new PR with upgraded sshj and Espresso test to verify the behaviour.

EmmanuelMess:
This should be fixed, test in next release candidate (next sunday) or by compiling

PR #1490, opened 2018-10-30
Verification for support of "new format" OpenSSH keys

TranceLove:
For verification of #1485.

 * Upgrade sshj to 0.26.0
 * Made sshj version number a variable in build.gradle
 * Added Espresso tests for parsing "new format" OpenSSH keys
 * Fix `CreateFileOnSshdTest`

The keys were generated after reading [this](https://blog.g3rt.nl/upgrade-your-ssh-keys.html), using

`ssh-keygen -o -a 100 -t ed25519`

Please read `PemToKeyPairTaskTest2` for the actual contents of the public and private keys.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -72,6 +72,7 @@ ext {
     supportLibVersion = '27.1.1'
     robolectricVersion = '3.8'
     glideVersion = '4.8.0'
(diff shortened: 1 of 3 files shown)


EmmanuelMess:
I'll add this to 3.3 as it is just an update.

Issue #1107, opened 2018-03-13
Have "Not sorted" sorting option


EmmanuelMess:
For faster folder loads.

Issue #843, opened 2017-11-26
Use C code instead of commands


EmmanuelMess:
The idea would be to make the methods in RootUtils native and work them from C.

VishalNehra:
What good would it do? Plus, there's no super su library for C.

EmmanuelMess:
Instead of parsing commands we'd call functions, it isn't that important, but could be done for more deterministic results (what a command prints is intended to be read by a human, therefore there might exist cases where a command would return something unexpected, causing difficult to trace crashes).
Why would we need a Super su lib for C? What function could it contain?

Alexander--:
This is definitely doable. I have created an extremely primitive [file manager](https://github.com/chdir/fdutil), that uses JNI + IPC to native daemon, and it was working well under my limited testing.

There are two caveats:

1. Increased size of crossplatform apk. This can be countered by using apk splits and compressing native code with upx (upx has recently gotten support for arm/x86 library compression in addition to executable compresssion).
2. Making a lot of wrappers for every method. You will have to maintain at least one JNI wrapper for every system call: stat, chown, readdir, readlink etc. Some may be tricky to write.

A bit of knowledge of SELinux and Android security architecture might be necessary (not much).

EmmanuelMess:
Crossplataform? You mean that different architectures would need different functions or that different API versions will need different functions? How so?

Alexander--:
@EmmanuelMess API functions are the same everywhere. But each CPU architecture (x86/x64/arm/arm64/mips/mips64) needs separate `.so` library file. So you would need libamaze-x86.so, libamaze-arm.so ...). 64-bit architectures can be skipped (they accept 32-bit libraries).

You have a choice: either put a version of each library (x86 .so + arm .so + ...) in the same apk (very bad idea, huge apk size) or distribute multiple apk files: one for arm, another for x86... mips is irrelevant, but serving third apk won't hurt much when you already serve two. The "fat" apk vs split apks setup is covered well in official Android docs.

> Wrappers? You only need to reference the function from java as 'native' and have a function of the same signature in C, how much more?

A bit more, depending on your goal.

You will most certainly need a C/C++ file (example: https://github.com/chdir/fdutil/blob/master/fdlib/src/main/cpp/native-lib.cpp#L319). You can write it manually (this is actually simple to do) or use an automated tool to generate it from Java method or C headers (the quality of generated code will be meh).

[...]

Issue #737, opened 2017-08-26
Load non essential info asyncroniously


EmmanuelMess:
For example folder's file and folder count.

vishal0071:
I did make quite a lots of things load on a background thread recently.. start times are faster and more consistent now. But your example above is an exception to this. I thought it was loaded on a background task. Regardless, I don't think anything much will require additional work apart from this example, but if you do end up working on other things, make sure you don't end up causing any race condition.

VishnuSanal:
Why is this still open? Isn't this fixed with [FilePreloaderLibrary](FilePreloaderLibrary)?

EmmanuelMess:
> Why is this still open? Isn't this fixed with [FilePreloaderLibrary](FilePreloaderLibrary)?

No, FilePreloaderLibrary is not used here right now.

EmmanuelMess:
The fix for this issue would be to use the library.

Issue #707, opened 2017-08-04
Trying to load gigantic directories *before* accessing them


EmmanuelMess:
Take for instance a directory of 1500 files (could happen) or more, MainFragment will simply find **every** file in it, and then show the directory, in my Moto E2 4G that takes time. I propose loading dynamically and telling the user that the directory isn't fully loaded; take also into account that the user won't look at each (1500) file/directory inside, probably they will just search.

PR #1494, opened 2018-11-02
super huge folder

llrraa:
To deal with super huge folder which contains tons of sub-folders and files, I make the following changes:
1. Add folder database to store the size and last sizing time of the folder.
2. When properties menu of a folder is clicked, if the last sizing time is lessing than 3 days, folderSize will directly read the size from the database. Or else,  folderSize will size the folder and save the size in database.
3. When LoadFilesListTask works, it will check if the size of the folder is in database or not. If yes, it will load the size.
4. If sorted by size, folder will also do.
This may not be needed by everyone, so it will be better if someone make super huge folder as a setting, so we can use it when we need it.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/LoadFilesListTask.java
+++ b/app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/LoadFilesListTask.java
@@ -29,8 +29,10 @@ import android.text.format.Formatter;
 
 import com.amaze.filemanager.R;
 import com.amaze.filemanager.adapters.data.LayoutElementParcelable;
+import com.amaze.filemanager.database.FolderHandler;
 import com.amaze.filemanager.database.SortHandler;
 import com.amaze.filemanager.database.UtilsHandler;
+import com.amaze.filemanager.database.models.Folder;
 import com.amaze.filemanager.exceptions.CloudPluginException;
 import com.amaze.filemanager.filesystem.HybridFile;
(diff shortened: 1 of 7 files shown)


EmmanuelMess:
I think it would be better to optimize the individual processes (like loading the files, loading the metadata, size, etc.) before deciding to use a cache.
Also, see #707, #737, #843, #1107

VishalNehra:
@llrraa that's lot of work you put into.
Please have a discussion first, before you go out working on such big issues.

EmmanuelMess:
And also, [FilePreloaderLibrary](https://github.com/TeamAmaze/FilePreloaderLibrary) exists.

llrraa:
Actually I'm dealing with a instant message folder, and it contains millions of files in it. My storage is almost full, so I want to delete something in it. before that, I have to know the size of folders in it. I try other file managers in the app market, but all fail me. That's why I have to make this commit. Now it's working for me.
I send this commit just in case someone might need, and it's a thank you to you guys and Amaze.

llrraa:
By the way, if  I want to discuss with you, how can I contact you?

VishalNehra:
@llrraa yes, @EmmanuelMess is already working on it in a seperate repo. You can work with him on it.
Also, you can simply create an issue, about the things you're interested in improving and we'll have a discussion.

llrraa:
I will always be available if you need my help.

VishnuSanal:
Why is this still open?

EmmanuelMess:
This was viable a long time ago.

PR #1500, opened 2018-11-04
Fix back item in list when headers are active

### LINKED ISSUE #1499: Back button goes to ".."
Pressing back moves you here:
screenshot_1541353200

EmmanuelMess:
Fixes #1499

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/adapters/RecyclerAdapter.java
+++ b/app/src/main/java/com/amaze/filemanager/adapters/RecyclerAdapter.java
@@ -46,6 +46,7 @@ import com.bumptech.glide.load.DataSource;
 import com.bumptech.glide.load.engine.GlideException;
 import com.bumptech.glide.request.RequestListener;
 import com.bumptech.glide.request.target.Target;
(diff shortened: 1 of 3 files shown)


TranceLove:
Fix works on Oneplus 2 running Slim7 (7.1.2) and an unrooted LG Nexus 5X running stock firmware (8.1.0).

However, one question before voting +1. While the back button works as advertised, I saw the back button colour changed to grey after applying the fix. Is it expected behaviour?

screenshot_20181107-105805

EmmanuelMess:
>  While the back button works as advertised, I saw the back button colour changed to grey after applying the fix. Is it expected behaviour?
 
Yes (was it not before? can't remember).

TranceLove:
The back button was following the theme colour before applying the changeset.

Not sure if it's in Material's design, but I think it's decent when we need to use distinct colours to distinguish different actions or navigating in different directions.

May @VishalNehra confirm too if it was designed to follow the theme colour or not? But I think I can vote +1 now, since the changeset did worked, and fixing colour won't be expensive anyway.

VishalNehra:
I think we can keep the grey back item.

PR #1502, opened 2018-11-07
Get spaces of all folders

llrraa:
I add one line so as to get  spaces of all folders.
To make it accepted by users, it's better to one more menu item after "properties" named such as "parent properties". When the folder is the child of root folder, then we get  spaces of all folders with this.


EmmanuelMess:
It seems that this PR also contains PR #1494, why?

llrraa:
PR #1494 is the base. That's why it's included here. The meaningful part is just one line. If you think it's useful, you can add it to another branch.

VishalNehra:
Needs major rebase. Closing. Please write us on our telegram channel and we'll discuss about these changes.
https://t.me/AmazeFileManager
Please open new pr if it's just one line change. That would be better I think.

PR #1507, opened 2018-11-14
Fix 1443

### LINKED ISSUE #1443: Not Found on Google Play
Clicking on the link in the readme that goes to google play i get "We're sorry, the requested URL was not found on this server."
Searching for "Amaze File Manager" yields no results either.

Any reason why it was removed from the play store ?

VishalNehra:
Fixes #1443 
This PR reintroduces different flavours for fdroid and play as I had to revoke 3rd party donation links (paypal) because of some dumb Google Play policy. I tried to talk to them but there's no other solution.
So the play flavor has Google in-app billing enabled. As fdroid can't compile Google libs, I had to create a separate flavor for it.
The donation itself won't work, it'll show some error unless Amaze is re-initiated on Play Store.
Also @EmmanuelMess please pull this with v3.3 branch and release it in production asap to get Amaze up on Play Store. 
v3.3 milestone has already been reached.


VishalNehra:
Yea, you can simply Google the error. It is caused because Amaze is not currently live. I couldn't test it either.
As for PayPal link, I figured it would be better to just add one link there because if we tried to revert back to original layout (different PayPal links for us all), we would need to maintain a seperate about page layout for different flavors.
We get negligible donations through paypal anyway, and I'll just divide any and all donations I get.

PR #1513, opened 2018-11-17
quickly_find_file

tumuyan:
quickly find file:
1.Apply a 10dp padding to the scroll bar. It is more easily to use finger to operate on screen edge.
2.Add a filter in the search box to apply a quick search in current directory


EmmanuelMess:
Please open an issue explaining what you want to add and why so we can address the concept.

PR #1530, opened 2018-11-27
Ability to disable marquee

proninyaroslav:
Hi. It would be great to be able to disable marquee, for example, for me this way of displaying text isn't very informative.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/adapters/AppsAdapter.java
+++ b/app/src/main/java/com/amaze/filemanager/adapters/AppsAdapter.java
@@ -23,6 +23,7 @@ import android.app.Activity;
 import android.content.ActivityNotFoundException;
 import android.content.Context;
 import android.content.Intent;
+import android.content.SharedPreferences;
 import android.content.pm.ApplicationInfo;
 import android.graphics.Color;
 import android.net.Uri;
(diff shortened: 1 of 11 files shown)


VishalNehra:
Personally I'm not in favor to put this option. We can increase the initial time to start marquee to something more suitable, though.

proninyaroslav:
Well, I think everyone has a different perception of time. For someone it will be fast, but for someone it will be slow. Maybe make time options for marquee (slow, medium, fast)?

VishalNehra:
That's just opposite of what my concern is. I don't want settings to be all cluttered up. That's why I suggest to keep the delay times which would help every one.

proninyaroslav:
This is the same element of perception as a color theme, which can also be customized in your app. Then it makes no sense to leave ability to change a color theme if you can't customize other visual things. But this is a question of philosophy and design.

VishalNehra:
@proninyaroslav yep, definitely it's a design perspective. I also got a few queries through mail to have the ability to disable this effect. I think we can keep this preference. Just move it from 'Interface' to 'General' in preference.

proninyaroslav:
Done.

VishalNehra:
@proninyaroslav Also, please change the english title to 'Enable marquee effect for file names' instead. Then all okay from my side too (:

EmmanuelMess:
Why was this added to 3.3.1?

PR #1553, opened 2018-12-26
Fix time format and organize imports

### LINKED ISSUE #1303: Negative remaining time
See "time remaining":

Seconds is negative and I think I've seen minutes go >60.

TranceLove:
Fixes #1303.

Tested on Oneplus 2 running Slim7 (7.1.2) copying files over SSH, extract zip and encrypt file.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/utils/Utils.java
+++ b/app/src/main/java/com/amaze/filemanager/utils/Utils.java
@@ -45,6 +45,7 @@ import com.amaze.filemanager.filesystem.HybridFileParcelable;
 import java.io.File;
 import java.text.SimpleDateFormat;
 import java.util.Locale;
+import java.util.concurrent.TimeUnit;
 
 /**
  * Contains useful functions and methods (NOTHING HERE DEALS WITH FILES)
@@ -260,10 +261,13 @@ public class Utils {
 
     /**
      * Formats input to plain mm:ss format
+     *
+     * @param timerInSeconds duration in seconds
+     * @return time in mm:ss format
      */
-    public static String formatTimer(long timer) {
-        final long min = timer / 60;
-        final long sec = timer % 60;
+    public static String formatTimer(long timerInSeconds) {
+        final long min = TimeUnit.SECONDS.toMinutes(timerInSeconds);
+        final long sec = TimeUnit.SECONDS.toSeconds(timerInSeconds - TimeUnit.MINUTES.toSeconds(min));
         return String.format(Locale.US, "%02d:%02d", min, sec);
     }
 }
--- a/app/src/main/java/com/amaze/filemanager/utils/Utils.java
+++ b/app/src/main/java/com/amaze/filemanager/utils/Utils.java
@@ -13,7 +13,6 @@ import android.support.v4.content.ContextCompat;
 import android.support.v4.graphics.drawable.DrawableCompat;
 import android.util.DisplayMetrics;
 import android.view.View;
-import android.view.Window;
 import android.widget.CheckBox;
 import android.widget.Toast;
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/amaze/filemanager/utils/Utils.java
```diff
@@ -241,8 +262,8 @@ public static int nthToLastCharIndex(int elementNumber, String str, char element
      * Formats input to plain mm:ss format
      */
     public static String formatTimer(long timer) {
-        final long min = TimeUnit.SECONDS.toMinutes(timer);
-        final long sec = TimeUnit.SECONDS.toSeconds(timer - TimeUnit.MINUTES.toMillis(min));
-        return String.format("%02d:%02d", min, sec);
+        final long min = timer / 60;
```
EmmanuelMess: You seem to be doing exactly what was being done before, but in a more 'illegible' way. See ` TimeUnit.SECONDS.toMinutes(timer);`:

```
public long toMinutes(long d) { return d/(C4/C3); }// C4/C3 == 60L
```

app/src/main/java/com/amaze/filemanager/utils/Utils.java
```diff
@@ -241,8 +262,8 @@ public static int nthToLastCharIndex(int elementNumber, String str, char element
...
+        final long sec = timer % 60;
+        return String.format(Locale.US, "%02d:%02d", min, sec);
```
EmmanuelMess: Why are you hardcoding a locale? doesn't the Android System manage locales acordingly?

app/src/main/java/com/amaze/filemanager/utils/Utils.java
```diff
@@ -241,8 +262,8 @@ public static int nthToLastCharIndex(int elementNumber, String str, char element
      * Formats input to plain mm:ss format
      */
     public static String formatTimer(long timer) {
-        final long min = TimeUnit.SECONDS.toMinutes(timer);
-        final long sec = TimeUnit.SECONDS.toSeconds(timer - TimeUnit.MINUTES.toMillis(min));
```
EmmanuelMess: It seems that the timer was already in seconds. The mistake seems to be the `toMillis(min)`, it should be `toSeconds(min)`

app/src/main/java/com/amaze/filemanager/utils/Utils.java
```diff
@@ -241,8 +262,8 @@ public static int nthToLastCharIndex(int elementNumber, String str, char element
      * Formats input to plain mm:ss format
      */
     public static String formatTimer(long timer) {
```
EmmanuelMess: Nowhere does it say the timer is in seconds, I think that is what led to the original mistake.

app/src/main/java/com/amaze/filemanager/utils/Utils.java
TranceLove: Per recommendation from Android Studio's lint. As long as this is simply `mm:ss`, it should be fine.

Unless this will change to locale-specific strings (e.g. mm分ss秒 in Chinese)...

EmmanuelMess: > Per recommendation from Android Studio's lint. As long as this is simply `mm:ss`, it should be fine.
>
> Unless this will change to locale-specific strings (e.g. mm分ss秒 in Chinese)...

It should be fine, but I don't know what transformations Java does based on the locale, and would not like to introduce subtle bugs, irreplicable bugs (see section "Be wary of the default locale" [here](https://developer.android.com/reference/java/util/Locale)).

I see two solutions, using the `SimpleDateTime` class and using [this fix](https://stackoverflow.com/questions/14389349/android-get-current-locale-not-default) from StackOverflow. I would also accept that you return the code to String.format("%02d:%02d", min, sec) so that lint keeps on warning.

PR #1557, opened 2018-12-30
Removed PREFERENCE_SHOW_HIDDENFILES check for showing hidden files list

EmmanuelMess:
Removed PREFERENCE_SHOW_HIDDENFILES check for showing hidden files list, because the user would be confused as to how to show the file just hidden (view selecting and clicking hide).


TranceLove:
Just wonder how should the settings in `PreferenceActivity` do after applying this changeset?

EmmanuelMess:
> Just wonder how should the settings in `PreferenceActivity` do after applying this changeset?

It only deals with files and folders that start with a dot.

VishalNehra:
I think changes would be more than this. We have a separate table for hidden files afaik. Also, hidden adapters and other UI stuff. You're meaning to remove Amaze's hidden feature in favour of filesystem convention, if I'm not mistaken?

EmmanuelMess:
> You're meaning to remove Amaze's hidden feature in favour of filesystem convention, if I'm not mistaken?

No, just that check.

Issue #1572, opened 2019-01-19
File copy operation is paused when the phone goes in stand by mode


bugnano:
When I'm copying a large file using SFTP, the process can take more than 1 minute, so the phone goes in stand by mode after that time.
Instead of continuing the copy operation in the background (as it does not require user input...), the copy operation does not continue until I unlock the phone.
After unlocking the phone the copy operation continues, but pauses again at the next phone stand by.

EmmanuelMess:
Ping @TranceLove.

TranceLove:
Copying files over SFTP also uses `CopyService` under the hood... and I tried the same on Oneplus 2 running Slim7 (7.1.2) and Sony Xperia Z1 Compact running CM13 (6.0.1) too, using Amaze's current `master` and my Ubuntu workstation running OpenSSH server 7.6p1. File copy was not suspended and file was successfully saved to device.

So my next questions to ask are
* details about device, Android version, etc
* does "standby mode" means pressing the power button on the device to turn screen off and lock device?

bugnano:
The device is a BQ Aquaris X5 running LineageOS based on Android 7.1.2

Yes, by stand by mode I mean either pressing the power button, or waiting enough time to have the device turn the screen off and lock itself automatically.

On the server side I'm using an old Fedora 20 server with openssh-server-6.4p1

When trying the same operation with FX File manager, the copy operation is much faster, and when the device is locked, the copy operation continues normally in the background, so it is not a problem neither on the server side, nor on the mobile operating system side.

VishalNehra:
Battery optimisation is enabled by default. We should present a way for
user to disable it.
Just for sake of testing, could you try disable battery optimisation for
Amaze manually from system settings and try copying file again?

As for copying speeds, sadly it can't be improved any further.

bugnano:
Not being able to speed up the file transfer is not a big issue.
However, I tried again copying a big file through SFTP with and without the battery optimizations for Amaze.
With the battery optimizations, after about 2 minutes of stand by, the file copy operation pauses.
Without the battery optimizations, the file copy operation fails after a few seconds.

VishalNehra:
Will look into it.

VishalNehra:
@zent-co could you please check this.

zent-co:
[...]

PR #1577, opened 2019-01-25
GenericCopyUtil refactoring for improving I/O performance

TranceLove:
Following discussion at #1572 I made an attempt to improve GenericCopyUtil on I/O operations.

* Unified all file copy operations to use NIO (under the hood, using delegation)
* Consequently, removed `CustomReadableByteChannel` and increment `ServiceWatcherUtil.position` directly
* Increased stream buffer size of SFTP and SMB to 64KB. This seems to be the better buffer size that increased network I/O copy performance without risk crashing the app.

Espresso and Robolectric tests for the refactoring are available. Sorry for the `DummyFileGenerator` in `src/main`, but it's the only way to make the class available to both Espresso and Robolectric tests - it had been guarded against main app code with `@VisibleForTesting` annotation.

Tested using Oneplus 2 running Slim7 (7.1.2) and Sony Xperia Z1 Compact running CM13 (6.0.1).

Known issue: `CopyService` was seen not able to stop after copying multiple files, i.e. the notification pull down did not go away after finished copying, on Sony Xperia Z1 Compact running CM13 (6.0.1).


app/src/main/java/com/amaze/filemanager/utils/files/GenericCopyUtil.java
```diff
@@ -325,129 +334,94 @@ public void copy(HybridFileParcelable sourceFile, HybridFile targetFile) throws
...
-                ServiceWatcherUtil.position += bytes;
-                return bytes;
+        buffer.flip();
+        while(buffer.hasRemaining())
+            to.write(buffer);
 
-            } else
-                return 0;
-        }
+        from.close();
```
VishalNehra: No need to close channels as they are being done in finally block of ````startCopy````

app/src/main/java/com/amaze/filemanager/utils/files/GenericCopyUtil.java
```diff
@@ -325,129 +334,94 @@ public void copy(HybridFileParcelable sourceFile, HybridFile targetFile) throws
...
+        to.close();
+    }
 
-        @Override
-        public boolean isOpen() {
-            return byteChannel.isOpen();
+    @VisibleForTesting
+    void doCopy(@NonNull ReadableByteChannel from, @NonNull FileChannel to) throws IOException{
```
VishalNehra: Do we really need this method? AFAIK, ````FileChannel```` implements both ````WritableByteChannel```` and ````ReadableByteChannel````.

TranceLove: `transferFrom` is implemented in FileChannel.

https://developer.android.com/reference/java/nio/channels/FileChannel#transferFrom(java.nio.channels.ReadableByteChannel,%20long,%20long)

app/src/main/java/com/amaze/filemanager/utils/files/GenericCopyUtil.java
TranceLove: Unfortunately, without closing the channels `GenericCopyUtilTest.testCopyFile4()` will fail.

VishalNehra: I see.

app/src/main/java/com/amaze/filemanager/utils/files/GenericCopyUtil.java
EmmanuelMess: IMO it is better to leave as much implementation as possible to Android, especially when dealing with the file system.

VishalNehra:
Approved this, but please modfiy the changes I asked for later when you get time @TranceLove

TranceLove:
@VishalNehra you mean dropping `FileChannel`-specific changes in favour of simply using `ByteBuffer`?

VishalNehra:
@TranceLove yes

TranceLove:
Made changes per @VishalNehra requested.

For the record, I refactored `GenericCopyUtilEspressoTest` so it tests doCopy() methods separately instead of grouping them together. Ran the test twice on Oneplus 2 running Slim7 (7.1.2), the was indeed a case where `FileChannel.transferFrom()` ran faster than while loop (as used by `doCopy(ReadableByteChannel from, WritableByteChannel)`, but the other test ran showed the opposite result.

Maybe more sampling can prove @EmmanuelMess 's suggestion is correct, but anyway.

TranceLove:
Two ticks, merging.

PR #1582, opened 2019-02-02
Fixing build errors

yashwanthreddyg:
Issue was due to duplicate attr name from fabsmenu library and unescaped apostrophes

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -180,7 +180,7 @@ dependencies {
     }
     annotationProcessor "com.github.bumptech.glide:compiler:$glideVersion"
 
-    implementation 'com.github.jahirfiquitiva:FABsMenu:1.1.3'//Floating Action Buttons Menu (aka expandable FAB)
+    implementation 'me.jahirfiquitiva:FABsMenu:1.1.1'//Floating Action Buttons Menu (aka expandable FAB)
 
     //Simple library show
     implementation('com.mikepenz:aboutlibraries:6.1.1@aar') {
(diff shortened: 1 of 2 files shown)


app/src/main/res/values-is/strings.xml
```diff
@@ -334,9 +334,9 @@
                         <b>Virkjaðu skráadeilingu á Windows</b>\n
                     </li>
                     <p style='margin-left: 40px'>
-                        Opnaðu stjórnskjáinn (Control Panel), smelltu á 'Choose homegroup and sharing
```
EmmanuelMess: You seem to have deleted a few "á" here.

app/build.gradle
```diff
@@ -180,7 +180,7 @@ dependencies {
     }
     annotationProcessor "com.github.bumptech.glide:compiler:$glideVersion"
 
-    implementation 'com.github.jahirfiquitiva:FABsMenu:1.1.3'//Floating Action Buttons Menu (aka expandable FAB)
+    implementation 'me.jahirfiquitiva:FABsMenu:1.1.1'//Floating Action Buttons Menu (aka expandable FAB)
```
EmmanuelMess: Why did you downgrade from 1.1.3 to 1.1.1?

yashwanthreddyg: It was a mistake. 1.1.3 had a fabSize attribute conflict. Upgraded it to 1.1.4.

app/src/main/res/values-is/strings.xml
yashwanthreddyg: Fixed this.

PR #1587, opened 2019-02-15
Update strings.xml

sveinki:
Added new translated string.


sveinki:
Oh, this is for the Icelandic translation.

PR #1598, opened 2019-03-06
Fix crash when rename file on smb server

### LINKED ISSUE #1596: Crash on renaming SMB file
When I rename a file in an SMB share, the app crashes.
But the file will be renamed anyway.

marcinadd:
Fixes #1596 
Renaming file was on main thread but you due to strict policy network can't be on main thread. I created AsyncTask via lamba expression


TranceLove:
Two ticks, merging.

PR #1603, opened 2019-03-11
Fix respecting hidden files while using sftp

### LINKED ISSUE #1601: sftp connections do not honor "show hidden files"
hidden files in an sftp connection are always shown regardless of whether I have "show hidden files" enabled.
This seems like the same problem that was previously fixed for SMB connections:
https://github.com/TeamAmaze/AmazeFileManager/issues/534

marcinadd:
I have fixed #1601 in simple way. 
Tested on:
Samsung Galaxy S4 Android 9.0 and it works perfectly.


app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/LoadFilesListTask.java
```diff
@@ -26,6 +27,7 @@
 import android.provider.MediaStore;
 import android.support.v4.util.Pair;
 import android.text.format.Formatter;
+import android.util.Log;
```
TranceLove: Redundant import.

marcinadd: Fixed

TranceLove:
Two ticks, merging.

Issue #1569, opened 2019-01-16
Amaze should use content:// rather than file:// when viewing files


infinity0:
Otherwise the viewing attempt will fail with "Permission denied" in `adb logcat`, if the viewer app does not itself *also* have the "Storage" permission.

To work-around this temporarily, you can explicitly give the viewer app the "Storage" permission. But ideally we should not need to do this - the whole point of the `content://` uri is to allow Amaze to pass on its own Storage permission temporarily to the viewer app.

https://stackoverflow.com/questions/33013508/content-uri-vs-file-uri-what-to-use-in-the-future

PR #1611, opened 2019-03-25
fix issue:: Amaze should use content:// rather than file:// when view…

ah-dev-med:
Amaze should use content:// rather than file:// when viewing files #1569


VishalNehra:
Declining as no response from author @ah-dev-med

Issue #1600, opened 2019-03-10
Extracting RAR files fails on LineageOS 16


toolstack:
I've recently updated to LineageOS 16 (aka Pie) and extracting RAR files now crashes Amaze.  Extraction works fine on LineageOS 15.1.

Zip files extract normally.

Using Amaze V3.3.2 from fdroid.

I've tested this on two devices, OnePlus 5t and a Galaxy Tab S2, both exhibit the same issue on any RAR file I try and extract.

I pulled the following logcat (logcat -v time *:E ) from the 5t:

```
03-10 17:15:58.887 E/storaged(  923): getDiskStats failed with result NOT_SUPPORTED and size 0
03-10 17:16:07.982 E/AndroidRuntime( 7420): FATAL EXCEPTION: AsyncTask #2
03-10 17:16:07.982 E/AndroidRuntime( 7420): Process: com.amaze.filemanager, PID: 7420
03-10 17:16:07.982 E/AndroidRuntime( 7420): java.lang.RuntimeException: An error occurred while executing doInBackground()
03-10 17:16:07.982 E/AndroidRuntime( 7420): 	at android.os.AsyncTask$3.done(AsyncTask.java:354)
03-10 17:16:07.982 E/AndroidRuntime( 7420): 	at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:383)
03-10 17:16:07.982 E/AndroidRuntime( 7420): 	at java.util.concurrent.FutureTask.setException(FutureTask.java:252)
03-10 17:16:07.982 E/AndroidRuntime( 7420): 	at java.util.concurrent.FutureTask.run(FutureTask.java:271)
03-10 17:16:07.982 E/AndroidRuntime( 7420): 	at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:245)
03-10 17:16:07.982 E/AndroidRuntime( 7420): 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
03-10 17:16:07.982 E/AndroidRuntime( 7420): 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
03-10 17:16:07.982 E/AndroidRuntime( 7420): 	at java.lang.Thread.run(Thread.java:764)
[...]

PR #1614, opened 2019-04-08
Prevents crash when extracting Rar files on devices using Android 9(Pie)

archibishop:
Fixes [#1600](https://github.com/TeamAmaze/AmazeFileManager/issues/1600)

Fixes the crash that is caused on devices running Android 9. Starting Android 6.0 `Apache HTTP client` was deprecated. So to use it those lines had to be added the `android manifest` file.

Tested on:
Nokia 3.1 running Android 9 and works.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -107,7 +107,6 @@ dependencies {
     testImplementation "org.robolectric:robolectric:$robolectricVersion"//tests android interaction
     testImplementation "org.robolectric:shadows-multidex:$robolectricVersion"//tests android interaction
     testImplementation "org.robolectric:shadows-httpclient:$robolectricVersion"//tests android interaction
-    testImplementation "commons-logging:commons-logging:1.2" //required by junrar. Damn.
 
     testImplementation "org.apache.sshd:sshd-core:1.7.0"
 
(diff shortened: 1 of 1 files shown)


TranceLove:
Tested on Oneplus Two running Slim7 (7.1.2) and LG Nexus 5x running AOSPExtended (9.0), work as advertised.

While I'll vote a +1, may I suggest an alternate solution?

When I worked 9d2081ca7 it was found that junrar depends on `commons-logging`. What if we simply include `jcl-over-slf4j` instead of enabling support of Android's legacy HTTP client?

Feel free to take a look at my suggestion at e62cc93. Tested on Oneplus Two running Slim7 (7.1.2) and LG Nexus 5x running AOSPExtended (9.0).

VishalNehra:
@TranceLove so what you mean is with e62cc93 we can still have junrar working? 
Also please tell why we're getting this issue to begin with? Why is rar extraction and logging dependencies interconnected?

TranceLove:
To begin with, [`junrar` depends on commons-logging](https://mvnrepository.com/artifact/com.github.junrar/junrar/4.0.0).

Next, while googling for an answer, I found [HttpClient for Android](https://hc.apache.org/httpcomponents-client-4.3.x/android-port.html), and from the background info it said:

> Google Android 1.0 was released with a pre-BETA snapshot of Apache HttpClient. To coincide with the first Android release Apache HttpClient 4.0 APIs had to be frozen prematurely, while many of interfaces and internal structures were still not fully worked out. As Apache HttpClient 4.0 was maturing the project was expecting Google to incorporate the latest code improvements into their code tree. Unfortunately it did not happen. Version of Apache HttpClient shipped with Android has effectively become a fork. Eventually Google decided to discontinue further development of their fork while refusing to upgrade to the stock version of Apache HttpClient citing compatibility concerns as a reason for such decision.

And [HttpClient 4.0 indicated `commons-logging` is indeed a dependency](https://mvnrepository.com/artifact/org.apache.httpcomponents/httpclient/4.0), suggesting `commons-logging` will become available with the legacy HttpClient is enabled.

On the other side, legacy HttpClient started to deprecate since Android 6, and I think we should stay away from it as well.

(Lucky we don't have WebDAV support yet, however ;))

archibishop:
@TranceLove @VishalNehra  Let me make the suggested changes.

archibishop:
I have tested on Nokia 3.1 running Android 9 and works.

VishalNehra:
@TranceLove please test

EmmanuelMess:
This has 2 approvals, and should be merged.

PR #1624, opened 2019-04-24
Fix crash on start and build for new gradle

4eRTuk:


4eRTuk:
Travis CI failed due to unaccepted license.

app/src/main/java/com/amaze/filemanager/database/UtilsHandler.java
```diff
@@ -346,7 +347,9 @@ public void updateSsh(String connectionName, String oldConnectionName, String pa
                 e.printStackTrace();
 
                 // failing to decrypt the path, removing entry from database
+                Looper.prepare();
```
EmmanuelMess: Why did you add these?

4eRTuk: Because Toast message is shown in async thread and produces a crash.

EmmanuelMess: How would Looper.prepare(); be a good solution though.

4eRTuk: It's AsyncTask's architectural limitation. There is no access to onPostExecute or onProgress from getSmbList() or getSftpList() functions, which are called from [these lines](https://github.com/TeamAmaze/AmazeFileManager/blob/e8fe39eeffd0df5ab429f12a8e64ede87a48ec22/app/src/main/java/com/amaze/filemanager/activities/MainActivity.java#L370-L371).

I can replace Looper with context's runOnUiThread() block, but I do not see big difference with creation of message loop for async thread with Looper.

EmmanuelMess: I don't understand how this solution doesn't crash.

[Example implementation](https://developer.android.com/reference/android/os/Looper.html) shows that a looper thread processes messages, sent via a handle in the `Looper.prepare() ... Looper.loop()` construct, but the messages seem to be processed in the same thread as the construct, not the main thread.

Furthermore, Toast isn't sending a message to the main thread AFAIK. And `Looper.quit()` isn't called as the documentation [seems to require](https://developer.android.com/reference/android/os/Looper.html#loop()).

4eRTuk: When toast is created, it is trying to send info message trough message queue, but in AsyncTask there is no attached one and exception appears.

Looper prepares message queue and attaches it to current thread. The idea is described [here](https://prasanta-paul.blogspot.com/2013/09/android-looper-and-toast-from.html) in details.

Method ```quit``` is do nothing useful despite the docs. Google's engineers [do not recommend](https://groups.google.com/d/msg/android-platform/_jsUqXi6DKM/9I_8U_0wtk0J) to use it.

One more option is to replace this construction with more common ```Handler(Looper.getMainLooper()).post(...)``` if you are more familar with it.

VishalNehra:
Master branch is passing on CI @4eRTuk

app/src/main/java/com/amaze/filemanager/database/UtilsHandler.java
EmmanuelMess: > When toast is created, it is trying to send info message trough message queue, but in AsyncTask there is no attached one and exception appears.

As far as I understand it `Looper.prepare() ... Looper.loop()` processes messages sent to a HANDLER in the CURRENT THREAD, NOT the MAIN THREAD.

[AFAIK](https://stackoverflow.com/a/47536058/3124150) Toasts cannot be shown from outside the main thread.

> Method ```quit``` is do nothing useful despite the docs.

From your source: 

> It is really important to quit the Looper, as if we don't do so, it will block all further operations, since Looper.loop() is a blocking call.

Blocking call means that the thread will be blocked at eternum.

PR #1644, opened 2019-06-18
Allow SD Card to change drawer name

### LINKED ISSUE #1076: Can't change name of sd in navigation bar
Please make it name as "SD card" , it show code of sd card as "123-ABC"

VishalNehra:
Fixes #1076


EmmanuelMess:
Long clicking doesn't seem like an intuitive way of changing the name.

EmmanuelMess:
Something I forgot is that SD cards' primary partition already have a name and are capable of saving name metadata.

maverick74:
So... Will there be another way to rename the card?

If we can't rename it for now, it would be nice to have, at least, a more readable name e.g.: "SDCard1"

Issue #1620, opened 2019-04-15
amaze crashes when going directly from inside a zip file to the apk manager (via side panel)


TjrGithub:
steps to reproduce:

* fresh install lineageos-with-microg 15.1 as of 2019-04-10 with addon_root on a wiped Mido
* install xposed and xprivacylua, restrict nothing important
* install amaze 3.3.2 from fdroid
* open amaze
* tap on a .zip file
* choose explore
* swipe to open material-design-left-sidebar
* tap on apk explorer
* watch crash, write this bug report

actual behavior: crash

expected behavior: as if I had closed the zip file before opening material-design-left-sidebar

[logcat_04-15-2019_12-42-27-amaze-crash-when-going-directly-from-zip-explorer-to-apk-explorer.txt](https://github.com/TeamAmaze/AmazeFileManager/files/3080022/logcat_04-15-2019_12-42-27-amaze-crash-when-going-directly-from-zip-explorer-to-apk-explorer.txt)

VishalNehra:
@TranceLove could you try to reproduce this? I'm unable to..

VishalNehra:
@TjrGithub please see if you can reproduce this after setting Amaze as default file manager to open zip files.

TranceLove:
I could reproduce the symptom, but with an extra step.

1. Open Amaze
2. Open an archive within Amaze
3. Open sidebar, open apk explorer
4. No apks showed up, and the window title was not `Apk` but `4`
5. Press back
6. Amaze crashed

```
2019-06-22 18:47:36.186 3420-4764/com.amaze.filemanager.debug E/AndroidRuntime: FATAL EXCEPTION: AsyncTask #5
    Process: com.amaze.filemanager.debug, PID: 3420
    java.lang.RuntimeException: An error occurred while executing doInBackground()
        at android.os.AsyncTask$3.done(AsyncTask.java:325)
        at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:354)
        at java.util.concurrent.FutureTask.setException(FutureTask.java:223)
        at java.util.concurrent.FutureTask.run(FutureTask.java:242)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
        at java.lang.Thread.run(Thread.java:761)
     Caused by: java.lang.NullPointerException
[...]

PR #1646, opened 2019-06-28
Prevent crash when going directly from inside a zip file to the apk manager (via side panel)

archibishop:
Fixes [#1620](https://github.com/TeamAmaze/AmazeFileManager/issues/1620)

Fix crash when going directly from inside a zip file to the apk manager (via side panel) and then trying to go back.
The correct title is now shown in the window title.

Tested on:
Nokia 3.1 running Android 9 and works.


EmmanuelMess:
What was causing the `path` to be null?

archibishop:
@EmmanuelMess The issue was that a wrong `openmode` was being yet some drawer items e.t.c apk , images , videos e.t.c require a different `openmode` of `custom`. Main fragment has a default openmode of file which works well if `internal storage` is loaded not the items with different `openmode` types. 
If the right open mode is used e.g apk then user presses back on after changing selecting apk from after loading a zip file then the lines below should be executed and takes you back to the home path which is already defined.
Screen Shot 2019-06-28 at 20 36 28

However with the wrong openmode it continues it fails at this point. Since it executes the file it executes case file and then root at which returns null.
Screen Shot 2019-06-28 at 20 55 57

PR #1658, opened 2019-08-10
FtpService Updates

zent-co:
I've observed Android shutting down the FTP Server on my Samsung S8+ running Android 9 and my Nexus 5x running LineageOS (Android 8.1.0). This service was not started as a foreground service and according to [Services overview](https://developer.android.com/guide/components/services): "If the service is bound to an activity that has user focus, it's less likely to be killed; if the service is declared to run in the foreground, it's rarely killed." 

The master branch code on both of my devices could not transfer large files via FTP because the version of Android on both of these phones killed the Service before the transfer was complete. The FTP Server service now stays running until it is stopped by the user. 

This PR improves the FTP Server performance, but it doesn't address Issue #1572.. Even with LineageOS, I could not duplicate the same behavior as Issue #1572. I also tried using a wakelock with the foreground service, but this combination did not help the user who submitted the issue. I think there is something beyond our control with that user's version of LineageOS that is taking over the FTP Service when his phone goes into standby. 

Even though I couldn't figure out what is going on with #1572, I believe this is still an improvement that should be included in future releases.

(edited by @TranceLove: please add # for convenient reference to issues/PR)

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/asynchronous/services/ftp/FtpService.java
+++ b/app/src/main/java/com/amaze/filemanager/asynchronous/services/ftp/FtpService.java
@@ -146,8 +146,6 @@ public class FtpService extends Service implements Runnable {
         serverThread = new Thread(this);
         serverThread.start();
 
-        acquireWakeLock();
-
         Notification notification = buildNotification();
 
         startForeground(NotificationConstants.FTP_ID, notification);
@@ -256,8 +254,6 @@ public class FtpService extends Service implements Runnable {
             serverThread = null;
         }
         if (server != null) {
-            releaseWakeLock();
-
             server.stop();
             sendBroadcast(new Intent(FtpService.ACTION_STOPPED).setPackage(getPackageName()));
         }
@@ -393,17 +389,6 @@ public class FtpService extends Service implements Runnable {
 
     public static NotificationCompat.Builder getBuilder() {
         return builder;
-    }
-
-    private void acquireWakeLock(){
-        PowerManager powerManager = (PowerManager) getSystemService(POWER_SERVICE);
-        wakeLock = powerManager.newWakeLock(PowerManager.PARTIAL_WAKE_LOCK,
-                "FtpService::Wakelock");
-        wakeLock.acquire();
-    }
-
-    private void releaseWakeLock(){
-        wakeLock.release();
     }
 
     private Notification buildNotification(){
--- a/app/src/main/java/com/amaze/filemanager/asynchronous/services/ftp/FtpService.java
+++ b/app/src/main/java/com/amaze/filemanager/asynchronous/services/ftp/FtpService.java
@@ -24,6 +24,8 @@ package com.amaze.filemanager.asynchronous.services.ftp;
 
 /**
  * Created by yashwanthreddyg on 09-06-2016.
+ *
+ * Edited by zent-co on 30-07-2019
  */
 
 import android.app.AlarmManager;
@@ -41,6 +43,7 @@ import android.net.wifi.WifiManager;
 import android.os.Build;
 import android.os.Environment;
 import android.os.IBinder;
+import android.os.PowerManager;
 import android.os.SystemClock;
 import android.preference.PreferenceManager;
 import android.support.annotation.NonNull;
@@ -122,6 +125,8 @@ public class FtpService extends Service implements Runnable {
 
     private static NotificationCompat.Builder builder;
 
+    private PowerManager.WakeLock wakeLock;
+
     @Override
     public int onStartCommand(Intent intent, int flags, int startId) {
         isStartedByTile = intent.getBooleanExtra(TAG_STARTED_BY_TILE, false);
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/amaze/filemanager/asynchronous/services/ftp/FtpService.java
```diff
@@ -112,6 +123,10 @@
 
     private boolean isStartedByTile = false;
 
+    private static NotificationCompat.Builder builder;
+
+    private PowerManager.WakeLock wakeLock;
```
EmmanuelMess: It seems you never use this field.

app/src/main/java/com/amaze/filemanager/asynchronous/services/ftp/FtpService.java
```diff
@@ -112,6 +123,10 @@
 
     private boolean isStartedByTile = false;
 
+    private static NotificationCompat.Builder builder;
```
EmmanuelMess: Why is this static? Shouldn't it get destroyed when the service is destroyed and recreated when needed?

app/src/main/java/com/amaze/filemanager/asynchronous/services/ftp/FtpService.java
```diff
@@ -366,4 +387,53 @@ private static Boolean callIsWifiApEnabled(@NonNull WifiManager wifiManager) {
         return r;
     }
 
+    public static NotificationCompat.Builder getBuilder() {
```
EmmanuelMess: Why is this static?

app/src/main/java/com/amaze/filemanager/ui/notifications/FtpNotification.java
```diff
@@ -49,46 +48,23 @@ private void createNotification(Context context, boolean noStopButton) {
                 + address.getHostAddress() + ":"
                 + port + "/";
 
+        NotificationCompat.Builder builder = FtpService.getBuilder();
```
EmmanuelMess: How are you sure that `FtpService.buildNotification()` was called and builder is not null? IMO it should be a compile time fail if it was not called for some reason, not a runtime crash.

app/src/main/java/com/amaze/filemanager/asynchronous/services/ftp/FtpService.java
zent-co: Yes, missed that one. Thanks.

app/src/main/java/com/amaze/filemanager/ui/notifications/FtpNotification.java
zent-co: The reason I made the NotificationCompat.Builder builder object static, the getter for the builder object static, and this method call all the way they were is that I couldn't figure out how to get the notification builder from FtpService to FtpNotification. However, after reviewing this today I now see that none of this needs to be static and I can update the notification that is always started in FtpService in FtpNotification by using the FTP_CHANNEL_ID. 

I am still testing my changes, but I know that the update to the notification in FtpNotification will always happen after FtpService notification is created because:
1. The notification is created in FtpService's onStartCommand callback
2. FtpNotification is only updated with a broadcast is sent to FtpNotification after the FTP Server is started and running. 

I should have listened to my gut because making the above static and calling it in FtpNotification just didn't feel right.

app/src/main/java/com/amaze/filemanager/asynchronous/services/ftp/FtpService.java
```diff
@@ -366,4 +383,51 @@ private static Boolean callIsWifiApEnabled(@NonNull WifiManager wifiManager) {
...
+
+        int stopIcon = android.R.drawable.ic_menu_close_clear_cancel;
+        CharSequence stopText = this.getString(R.string.ftp_notif_stop_server);
+        Intent stopIntent = new Intent(FtpService.ACTION_STOP_FTPSERVER).setPackage(getApplicationContext().getPackageName());
+        PendingIntent stopPendingIntent = PendingIntent.getBroadcast(getApplicationContext(), 0,
+                stopIntent, PendingIntent.FLAG_ONE_SHOT);
+
+        builder.addAction(stopIcon, stopText, stopPendingIntent);
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN && Build.VERSION.SDK_INT < Build.VERSION_CODES.O) {
```
EmmanuelMess: All these if can be replaced by `NotificationConstants.setMetadata(getApplicationContext(), builder, NotificationConstants.TYPE_FTP);`

app/src/main/java/com/amaze/filemanager/ui/notifications/FtpNotification.java
```diff
@@ -49,46 +49,31 @@ private void createNotification(Context context, boolean noStopButton) {
                 + address.getHostAddress() + ":"
                 + port + "/";
 
+        NotificationCompat.Builder builder = new NotificationCompat.Builder(context.getApplicationContext(), NotificationConstants.CHANNEL_FTP_ID);
```
EmmanuelMess: You rebuild the entire thing here. You should try to centralize building to a single method.

app/src/main/java/com/amaze/filemanager/ui/notifications/FtpNotification.java
```diff
@@ -49,46 +49,31 @@ private void createNotification(Context context, boolean noStopButton) {
...
+        CharSequence stopText = context.getString(R.string.ftp_notif_stop_server);
+        Intent stopIntent = new Intent(FtpService.ACTION_STOP_FTPSERVER).setPackage(context.getApplicationContext().getPackageName());
+        PendingIntent stopPendingIntent = PendingIntent.getBroadcast(context.getApplicationContext(), 0,
+                stopIntent, PendingIntent.FLAG_ONE_SHOT);
+
+        builder.addAction(stopIcon, stopText, stopPendingIntent);
 
-        // Pass Notification to NotificationManager
-        notificationManager.notify(NotificationConstants.FTP_ID, notification);
+        notificationManager.notify(NotificationConstants.FTP_ID, builder.build());
```
EmmanuelMess: Channel?

app/src/main/java/com/amaze/filemanager/asynchronous/services/ftp/FtpService.java
```diff
@@ -366,4 +383,51 @@ private static Boolean callIsWifiApEnabled(@NonNull WifiManager wifiManager) {
         return r;
     }
 
+    private Notification buildNotification(){
```
EmmanuelMess: Please move to FtpNotification.

app/src/main/java/com/amaze/filemanager/ui/notifications/NotificationConstants.java
```diff
@@ -31,6 +31,8 @@
     public static final String CHANNEL_NORMAL_ID = "normalChannel";
     public static final String CHANNEL_FTP_ID = "ftpChannel";
 
+    public static final String CHANNEL_FTP_NAME = "FtpChannel";
```
EmmanuelMess: This is not localizable.

app/src/main/java/com/amaze/filemanager/ui/notifications/FtpNotification.java
zent-co: It looks like this is the only item that isn't resolved. Does it still need to be done? I can probably do it Friday or Saturday.

EmmanuelMess: I left it to you because I don't want to break everything :smile:.

zent-co: Haha! Thanks for leaving me one thing to break. 😄

EmmanuelMess:
Ok, I removed the static variables. In general do not use static variables on Android because they [live until they don't](https://stackoverflow.com/q/17900456/3124150).

zent-co:
> Code is OK, can't really replicate the issue in master, but I suppose this fixes. Before merge someone else should really try to replicate though.

I agree. The more people that can look at this the better. It is very dependent on the version of Android.

app/src/main/java/com/amaze/filemanager/ui/notifications/FtpNotification.java
```diff
@@ -19,22 +17,24 @@
...
-                createNotification(context, intent.getBooleanExtra(FtpService.TAG_STARTED_BY_TILE, false));
+                updateNotification(context);
                 break;
             case FtpService.ACTION_STOPPED:
                 removeNotification(context);
                 break;
         }
     }
 
-    private void createNotification(Context context, boolean noStopButton) {
```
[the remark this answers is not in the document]
zent-co: No specific reason besides I neglected to fully understand what that code was doing. I now see that if the service is started from the tile, then there should not be a stop button and the service should be stopped from the tile.  I can fix this.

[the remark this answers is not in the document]
zent-co: This is now fixed.

PR #1665, opened 2019-08-26
Update gradle

EmmanuelMess:
Updated gradle.

----
<a class="imgpatreon" href="https://www.patreon.com/emmanuelmess" target="_blank">
Become a patreon</a>

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -181,7 +181,7 @@ dependencies {
     }
     annotationProcessor "com.github.bumptech.glide:compiler:$glideVersion"
 
-    implementation 'com.github.jahirfiquitiva:FABsMenu:1.1.3'//Floating Action Buttons Menu (aka expandable FAB)
+    implementation 'com.github.jahirfiquitiva:FABsMenu:1.1.4'//Floating Action Buttons Menu (aka expandable FAB)
 
     //Simple library show
     implementation('com.mikepenz:aboutlibraries:6.1.1@aar') {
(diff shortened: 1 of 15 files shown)


TranceLove:
One thing I noticed though on building, is when gradle is at `:app:lintVitalFdroidRelease` or `:app:lintVitalPlayRelease` task, got this at the console
```
/home/trancelove/git/AmazeFileManager/app/src/main/res/values-ldltr/bools.xml:3: Error: The bool "is_right_to_left" in values-ldltr has no declaration in the base values folder; this can lead to crashes when the resource is queried in a configuration that does not match this qualifier [MissingDefaultResource]
    <bool name="is_right_to_left">false</bool>
          ~~~~~~~~~~~~~~~~~~~~~~~
/home/trancelove/git/AmazeFileManager/app/src/main/res/values-ldrtl/bools.xml:3: Error: The bool "is_right_to_left" in values-ldrtl has no declaration in the base values folder; this can lead to crashes when the resource is queried in a configuration that does not match this qualifier [MissingDefaultResource]
    <bool name="is_right_to_left">true</bool>
          ~~~~~~~~~~~~~~~~~~~~~~~

   Explanation for issues of type "MissingDefaultResource":
   If a resource is only defined in folders with qualifiers like -land or -en,
   and there is no default declaration in the base folder (layout or values
   etc), then the app will crash if that resource is accessed on a device
   where the device is in a configuration missing the given qualifier.

   As a special case, drawables do not have to be specified in the base
   folder; if there is a match in a density folder (such as drawable-mdpi)
   that image will be used and scaled. Note however that if you  only specify
   a drawable in a folder like drawable-en-hdpi, the app will crash in
   non-English locales.

   There may be scenarios where you have a resource, such as a -fr drawable,
   which is only referenced from some other resource with the same qualifiers
   (such as a -fr style), which itself has safe fallbacks. However, this still
   makes it possible for somebody to accidentally reference the drawable and
   crash, so it is safer to create a default dummy fallback in the base
   folder. Alternatively, you can suppress the issue by adding
   tools:ignore="MissingDefaultResource" on the element.

   (This scenario frequently happens with string translations, where you might
   delete code and the corresponding resources, but forget to delete a
   translation. There is a dedicated issue id for that scenario, with the id
   ExtraTranslation.)

2 errors, 0 warnings
```

Seems it didn't break the build, and the app didn't crash on my device (Oneplus 2 running Slim7), I think this won't hurt - just for the note.

PR #1681, opened 2019-09-05
Added code to check if decompressor is null. Throw a toast if it is.

zent-co:
I found that if the file type doesn't match a known compressed file type, the decompressor object returned by static method CompressedHelper.getCompressorInstance() is null. 

If decompressor is null, then the changePath() method throws the NPE on decompressor.changePath(). In order to prevent this, I check to make sure that decompressor is not null, and if it is I throw a toast to indicate there is an issue with the file. 

Depending on expectations, the UI response may need to be changed.


EmmanuelMess:
@zent-co if the decompressor returned null it's too late to correct the problem. The problem is that a new decompressor is getted in the first place.

VishalNehra:
Please create issues before creating relevant PRs
I've created one for this. You can edit and point fix to #1682

EmmanuelMess:
2 approvals merging.

PR #1683, opened 2019-09-12
Translations and script to translate

EmmanuelMess:
Now `tx pull` can be used to pull new translations.

----
<a class="imgpatreon" href="https://www.patreon.com/emmanuelmess" target="_blank">
Become a patreon</a>


EmmanuelMess:
@TranceLove tx pull should work again now

VishalNehra:
So if anyone adds translation to transifex, will it automatically create here? And if we directly change some file here, what would happen?
Also, main strings file changes would be reflected on transifex as well, so others can translate new strings?

EmmanuelMess:
Nevermind, transifex bot seems to be working now.

PR #1694, opened 2019-09-24
Use content uri to open files

### LINKED ISSUE #1691: PDF file opening crashes on SDK 29
Android 10 giving following error while opening pdf file. Other files seem to be opening fine.
Do note, the log is not related to Amaze
```
2019-09-23 13:07:42.442 7336-7645/? E/DisplayData: openFd: java.io.FileNotFoundException: open failed: EACCES (Permission denied)
2019-09-23 13:07:42.443 7336-7645/? E/PdfLoader: Can't load file (doesn't open)  Display Data [PDF : acrobat_8_help.pdf] +FileOpenable, uri: file:///storage/emulated/0/Download/acrobat_8_help.pdf
```

EmmanuelMess:
There already was a provider tag, but I created a new one, is that correct?

Fix #1691

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -226,6 +226,16 @@
                 android:name="android.support.FILE_PROVIDER_PATHS"
                 android:resource="@xml/provider_paths"/>
         </provider>
+
+        <provider
+            android:name="androidx.core.content.FileProvider"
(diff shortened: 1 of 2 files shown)


TranceLove:
With the previously defined custom FileProvider no visible regression was found. I think we should be able to live without it. Keeping my +1.

PR #1698, opened 2019-10-07
Removed all references to Google Plus

### LINKED ISSUE #1671: Google Plus references should be removed
Since Google Plus is shut down, references to it should be removed from About page.

bowiechen:
Fixes #1671. 

I noticed this project has references to Transifex. If it is still actively used, I'm not sure if my deletions in the local .xml files will survive a pull.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,8 +49,6 @@ We strongly recommend using apk signed by us (either Play Store version or from 
 
 Social
 ---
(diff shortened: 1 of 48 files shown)


EmmanuelMess:
> I noticed this project has references to Transifex. If it is still actively used, I'm not sure if my deletions in the local .xml files will survive a pull.

I hope they do.

PR #1719, opened 2019-10-28
Changed dismiss button of hidden dialog to be "OK"

### LINKED ISSUE #1717: Misleading prompt text in "Hidden Files" dialog
**Expected behavior**

The button to close the "Hidden Files" pane to read "OK" or something of positive phrasing.

**Actual behavior**

The button reads "Cancel".

**Steps to reproduce**
1. Tap the "Hidden Files" menu item in the top right context menu.

**System info**
- Phone model: Google Pixel 2
- Android version: 10
- Rooted: No
- Amaze File Manager version: v3.3.4

bowiechen:
Fixes #1717.


VishalNehra:
Changes seem fine, but I suggest we keep it 'Close' instead of 'Ok' 
Reason being, consider if you press on delete icon in hidden dialog it'll immediately remove the file from hidden list, regardless of whether you press on 'Ok' or not.

bowiechen:
Good point.

EmmanuelMess:
Did I add commits to this branch?

app/src/main/res/values-zh-rTW/strings.xml
```diff
@@ -602,4 +602,5 @@
   <string name="archive_password_prompt">請輸入壓縮檔的密碼。</string>
   <string name="cannot_extract_archive">不能開啟壓縮檔 \"%s\"： %s</string>
   <string name="archive_unsupported_or_corrupt">不能開啟壓縮檔 \"%s\"。可能是程式庫不支援此壓縮檔，或者壓縮檔真的壞掉了。</string>
+  <string name="close">關閉</string>
```
EmmanuelMess: You should translate this on transifex when the source file is updated, because it will probably will be overwritten.

PR #1721, opened 2019-10-28
WIP: Refresh file list on "Hidden Files" dialog dismissal

### LINKED ISSUE #1712: Unhiding hidden files does not show the file (requires a refresh)
**Expected behavior**

The file removed from the hidden files list should be immediately visible in the browser. 

**Actual behavior**

The file is not immediately visible. The file is only visible when the user manually triggers a refresh of the current file view. 

**Steps to reproduce** 

1. Select a file.
1. Tap "Hide".
1. Tap "Hidden Files" from top right context menu.
1. Tap "x" on the file to remove from hidden files list.
1. Tap "Cancel".

bowiechen:
Fixes #1712. 

The list of files is now refreshed when the dialog is dismissed. 

Currently, un-hiding a file works fine, but un-hiding a directory immediately triggers a refresh. Refresh is a costly operation so we would like to push that operation to the end of the dialog lifecycle.


786-rizwan:
Hiding a file, don't hide that file from main Gallery(File Manager)

VishalNehra:
Closing this reused code in #1980 
Thank you (:

PR #1723, opened 2019-10-29
Configure transifex and update readme

EmmanuelMess:
Configure Transifex file names for regions and update README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,9 +37,8 @@ See our [Privacy Policy](https://github.com/TeamAmaze/AmazeFileManager/wiki/Priv
 Translators
 ----
 Want to help translate Amaze to your language?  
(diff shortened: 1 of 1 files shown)


VishalNehra:
Can you leave a reference, I don;t understand the need for this one.

EmmanuelMess:
> Can you leave a reference, I don;t understand the need for this one.

We don't really need the config file, unless someone wants to use the command line tool, the update to the readme is useful though.

EmmanuelMess:
@VishalNehra Removed config file.

EmmanuelMess:
> Does this change mean that we do not encourage people to make PR for changing translations?

Changing files is already not encouraged.

> What about the changes made in #1707? Will the changes be reverted by Transifex?

Changes to en_US language are updated INTO Transifex, changes to strings in other languages are overwritten.

PR #1726, opened 2019-10-31
Use EventBus to transmit FtpReceiverActions of FTP server status

TranceLove:
This reduces overhead and prevent malicious apps outside Amaze to clear FTP server status with Intents.

Tested only on Oneplus 2 running Slim7 (7.1.2) though.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/asynchronous/services/ftp/FtpService.java
+++ b/app/src/main/java/com/amaze/filemanager/asynchronous/services/ftp/FtpService.java
@@ -75,6 +75,7 @@ import org.apache.ftpserver.ssl.ClientAuth;
 import org.apache.ftpserver.ssl.impl.DefaultSslConfiguration;
 import org.apache.ftpserver.usermanager.impl.BaseUser;
 import org.apache.ftpserver.usermanager.impl.WritePermission;
+import org.greenrobot.eventbus.EventBus;
 
 public class FtpService extends Service implements Runnable {
 
@@ -95,10 +96,13 @@ public class FtpService extends Service implements Runnable {
     private static final String WIFI_AP_ADDRESS_PREFIX = "192.168.43.";
     private static final char[] KEYSTORE_PASSWORD = "vishal007".toCharArray();
 
-    // Service will (global) broadcast when server start/stop
-    static public final String ACTION_STARTED = "com.amaze.filemanager.services.ftpservice.FTPReceiver.FTPSERVER_STARTED";
-    static public final String ACTION_STOPPED = "com.amaze.filemanager.services.ftpservice.FTPReceiver.FTPSERVER_STOPPED";
-    static public final String ACTION_FAILEDTOSTART = "com.amaze.filemanager.services.ftpservice.FTPReceiver.FTPSERVER_FAILEDTOSTART";
+    // Service will broadcast via event bus when server start/stop
+    public enum FtpReceiverActions {
+        STARTED,
+        STARTED_FROM_TILE,
+        STOPPED,
+        FAILED_TO_START
+    }
 
     // RequestStartStopReceiver listens for these actions to start/stop this server
     static public final String ACTION_START_FTPSERVER = "com.amaze.filemanager.services.ftpservice.FTPReceiver.ACTION_START_FTPSERVER";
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/amaze/filemanager/asynchronous/services/ftp/FtpService.java
```diff
@@ -96,9 +97,13 @@
...
-    static public final String ACTION_STARTED = "com.amaze.filemanager.services.ftpservice.FTPReceiver.FTPSERVER_STARTED";
-    static public final String ACTION_STOPPED = "com.amaze.filemanager.services.ftpservice.FTPReceiver.FTPSERVER_STOPPED";
-    static public final String ACTION_FAILEDTOSTART = "com.amaze.filemanager.services.ftpservice.FTPReceiver.FTPSERVER_FAILEDTOSTART";
+    public enum FtpReceiverActions
+    {
+        STARTED,
+        STARTED_FROM_TILE,
+        STOPPED,
+        FAILED_TO_START
+    }
```
bowiechen: Are we doing brackets on new lines or on the same line?

app/src/main/java/com/amaze/filemanager/fragments/FtpServerFragment.java
```diff
@@ -331,37 +339,36 @@ public void onReceive(Context context, Intent intent) {
...
-                    break;
-                case FtpService.ACTION_STOPPED:
-                    statusText.setText(spannedStatusNotRunning);
-                    url.setText("URL: ");
-                    ftpBtn.setText(getResources().getString(R.string.start_ftp).toUpperCase());
-                    break;
-            }
+    @Subscribe(threadMode = ThreadMode.MAIN_ORDERED)
+    public void onFtpReceiveActions(FtpService.FtpReceiverActions signal)
+    {
```
bowiechen: Are we keeping the brackets on a newline or the same line?

app/src/main/java/com/amaze/filemanager/asynchronous/services/ftp/FtpService.java
```diff
@@ -220,10 +225,11 @@ public void run() {
...
             server = serverFactory.createServer();
             server.start();
-
-            sendBroadcast(new Intent(FtpService.ACTION_STARTED).setPackage(getPackageName()).putExtra(TAG_STARTED_BY_TILE, isStartedByTile));
+            EventBus.getDefault().post(isStartedByTile ? FtpReceiverActions.STARTED_FROM_TILE : FtpReceiverActions.STARTED);
+            //sendBroadcast(new Intent(FtpService.ACTION_STARTED).setPackage(getPackageName()).putExtra(TAG_STARTED_BY_TILE, isStartedByTile));
         } catch (Exception e) {
-            sendBroadcast(new Intent(FtpService.ACTION_FAILEDTOSTART).setPackage(getPackageName()));
+            EventBus.getDefault().post(FtpReceiverActions.FAILED_TO_START);
+            //sendBroadcast(new Intent(FtpService.ACTION_FAILEDTOSTART).setPackage(getPackageName()));
```
bowiechen: Are the comments here necessary? If not, we can clean up the code base.

app/src/main/java/com/amaze/filemanager/asynchronous/services/ftp/FtpService.java
```diff
@@ -95,10 +96,13 @@
     private static final String WIFI_AP_ADDRESS_PREFIX = "192.168.43.";
     private static final char[] KEYSTORE_PASSWORD = "vishal007".toCharArray();
 
-    // Service will (global) broadcast when server start/stop
-    static public final String ACTION_STARTED = "com.amaze.filemanager.services.ftpservice.FTPReceiver.FTPSERVER_STARTED";
-    static public final String ACTION_STOPPED = "com.amaze.filemanager.services.ftpservice.FTPReceiver.FTPSERVER_STOPPED";
-    static public final String ACTION_FAILEDTOSTART = "com.amaze.filemanager.services.ftpservice.FTPReceiver.FTPSERVER_FAILEDTOSTART";
+    // Service will broadcast via event bus when server start/stop
+    public enum FtpReceiverActions {
```
EmmanuelMess: Is there a reason to use an enum?

TranceLove: We are emitting events within the app, we can choose the lightweight pojo path. enums are typed, pojo, and we don't need additional properties anyway, so I chose enums over static classes (per EventBus's example).

EmmanuelMess:
I don't get what components event bus is need to comunicate. Is it the notification and the service?

TranceLove:
Agreed so, but for the time being EventBus seems to be a convenient replacement instead of refactoring at a bigger scale, while keeping the current way of decoupling between components.

We may keep the whole thing intent-based, but intents are relatively heavy, and we don't want to make it beyond the app's boundary anyway (though it's possible).

Some Kotlin-based solutions without external libs can be found, but we are not using Kotlin.

PR #1728, opened 2019-11-03
Made string random text similar to random_summary

TipzTeam:
I made string random similar to random_summary because it is all about random theme, but string random have a different names.


EmmanuelMess:
You should update the string [here](https://www.transifex.com/amaze/amaze-file-manager/translate/#zh_HK/strings-xml--master), changes to files will be overwritten.

PR #1729, opened 2019-11-03
Update indonesian translation

zmni:


EmmanuelMess:
Please do it via transifex: [here](https://www.transifex.com/amaze/amaze-file-manager/language/id/).

PR #1777, opened 2019-11-10
Fixed NPE when tabs don't exist in database.

EmmanuelMess:


app/src/main/java/com/amaze/filemanager/fragments/TabFragment.java
```diff
@@ -126,25 +127,25 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
                 }
             } else {
                 if (path != null && path.length() != 0) {
```
bowiechen: Instead of doing 
```Java
path != null && path.length() != 0
```
you can use 
```Java
!TextUtils.isEmpty(path)
```
to simplify all the null and empty checks.  Read more about [`TextUtils.isEmpty(ChatSequence)`](https://developer.android.com/reference/android/text/TextUtils.html#isEmpty(java.lang.CharSequence)) on the Android Dev page.

bowiechen: Instead of doing 
```Java
path != null && path.length() != 0
```
you can use 
```Java
!TextUtils.isEmpty(path)
```
to simplify all the null and empty checks.  Read more about [`TextUtils.isEmpty(ChatSequence)`](https://developer.android.com/reference/android/text/TextUtils.html#isEmpty(java.lang.CharSequence)) on the Android Dev page.

EmmanuelMess: Yeah, but that was already in master.

PR #1780, opened 2019-11-23
setting secure connection as default

albariko97:


EmmanuelMess:
Wasn't there are reason for the checkbox? I am not so well versed in the networking side, but doesn't the secure connection need extra input?

PR #1813, opened 2020-01-08
Add mts file extension as video/mts

### LINKED ISSUE #1696: .mts video files do not display as video, how can I add to MIME-type?

TranceLove:
Fixes #1696.

Tested on Oneplus 2 running Slim7 (7.1.2) with MX Player installed, and LG Nexus 5x running AOSPExtended (9.0) with VLC installed.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/ui/icons/MimeTypes.java
+++ b/app/src/main/java/com/amaze/filemanager/ui/icons/MimeTypes.java
@@ -138,6 +138,7 @@ public final class MimeTypes {
         MIME_TYPES.put("ogv", "video/ogg");
         MIME_TYPES.put("flv", "video/x-flv");
         MIME_TYPES.put("mkv", "video/x-matroska");
+        MIME_TYPES.put("mts", "video/mp2t");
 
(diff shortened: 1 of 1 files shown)


EmmanuelMess:
Glide can't load a preview though.

TranceLove:
> Glide can't load a preview though.

My rough test suggested the same. I suppose preview thumbnails shall not work until Android supports mts natively?

VishalNehra:
Glide never supported video playback, it only supported gif playback in icon afaik.
BTW I think the mime type for mts should be `video/MP2T` according to (this)[https://tools.ietf.org/html/rfc3555#section-4.2.9]

EmmanuelMess:
The glide thing was just a note, no worries.

> BTW I think the mime type for mts should be `video/MP2T` according to [this](https://tools.ietf.org/html/rfc3555#section-4.2.9)

I don't quite follow, in what way is "video/MP2T" related to *.mts files?

VishalNehra:
MTS is MPEG2 Transport Streams, and according to above link that's the RFC standard mime for it.

TranceLove:
See 2c44382. However I'm keeping the MIME type to all lowercase because they don't need to be uppercase for any reason, and practically all MIME types are in all lowercase letters anyway.

PR #1814, opened 2020-01-08
Restrict intents send to FtpReceiver to control FTP server

TranceLove:
Per mail discussion offline it is found that any app may fire an Intent with action `com.amaze.filemanager.services.ftpservice.FTPReceiver.ACTION_START_FTPSERVER` or `com.amaze.filemanager.services.ftpservice.FTPReceiver.ACTION_STOP_FTPSERVER` to control Amaze's built-in FTP server without user acknowledgement.

Changes in changeset
- add a custom permission `com.amaze.filemanager.permission.CONTROL_FTP_SERVER` and set it at `signature` level
- `FtpService` requires `com.amaze.filemanager.permission.CONTROL_FTP_SERVER` permission
- Intents sent to `FtpReceiver` requires package name which is sender's app package name. Will check if the app is installed; also it will deny intents sent from system apps - they won't send intents to us anyway

This can be considered as a "part 1" of the bigger task on improving FTP server security; another PR will be created for re-enabling integration with external apps, by prompting user on external apps sending intents to `FtpReceiver`.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/utils/Utils.java
+++ b/app/src/main/java/com/amaze/filemanager/utils/Utils.java
@@ -30,15 +30,16 @@ import android.content.res.Configuration;
 import android.graphics.drawable.Drawable;
 import android.net.Uri;
 import android.os.Build;
-import androidx.annotation.ColorRes;
-import androidx.annotation.NonNull;
-import androidx.core.content.ContextCompat;
-import androidx.core.graphics.drawable.DrawableCompat;
 import android.text.format.DateUtils;
 import android.util.DisplayMetrics;
 import android.view.View;
 import android.widget.CheckBox;
 import android.widget.Toast;
+
+import androidx.annotation.ColorRes;
+import androidx.annotation.NonNull;
+import androidx.core.content.ContextCompat;
+import androidx.core.graphics.drawable.DrawableCompat;
 
 import com.amaze.filemanager.R;
 import com.amaze.filemanager.activities.MainActivity;
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -23,6 +23,9 @@
     xmlns:tools="http://schemas.android.com/tools"
     package="com.amaze.filemanager">
 
+    <permission android:name="com.amaze.filemanager.permission.CONTROL_FTP_SERVER"
+        android:protectionLevel="signature" />
+
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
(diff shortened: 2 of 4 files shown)


app/src/androidTest/java/com/amaze/filemanager/asynchronous/services/ftp/FtpServiceEspressoTest.java
```diff
@@ -60,6 +60,7 @@ public void testFtpService() throws IOException {
         PreferenceManager.getDefaultSharedPreferences(service).edit()
                 .putBoolean(FtpService.KEY_PREFERENCE_SECURE, false).commit();
         service.onStartCommand(new Intent(FtpService.ACTION_START_FTPSERVER)
+                .setPackage("com.amaze.filemanager")
```
EmmanuelMess: Can't malicious app set the package like its set here?

app/src/main/java/com/amaze/filemanager/utils/Utils.java
```diff
@@ -263,4 +263,32 @@ public static String formatTimer(long timerInSeconds) {
         final long sec = TimeUnit.SECONDS.toSeconds(timerInSeconds - TimeUnit.MINUTES.toSeconds(min));
         return String.format("%02d:%02d", min, sec);
     }
+
+    public static String wildcardToRegex(String wildcard) {
```
EmmanuelMess: Why is this method needed?

app/src/androidTest/java/com/amaze/filemanager/asynchronous/services/ftp/FtpServiceEspressoTest.java
TranceLove: Of course. Just for a first line of defense, hence I called PackageManager to see if the app's package is actually installed on the device.

On the second phase user will be prompted if he/she should allow an external app to switch on/off the FTP server.

app/src/main/java/com/amaze/filemanager/utils/Utils.java
TranceLove: Convenience method for package name checking, so `FtpReceiver.blacklistedApps` can use wildcards to define package names to blacklist.

app/src/androidTest/java/com/amaze/filemanager/asynchronous/services/ftp/FtpServiceEspressoTest.java
EmmanuelMess: > Of course. It's just for a first line of defense, hence I called PackageManager to see if the app's package is actually installed on the device.
> (EDIT: OK I get this - malicious app can simply declare intent is sent from Amaze itself. But is there a better solution?)

This is not a viable solution, as it only makes starting the server more difficult and is easily bypassed by malicious actors.

> On the second phase user will be prompted if he/she should allow an external app to switch on/off the FTP server.

This seems like the only viable defence right now, IMO.

TranceLove: And `packageName` actually does the opposite thing that makes the intent to fire to the source package itself... :( will work on it.

TranceLove:
OK, let's makes things easier to understand by not messing the 2 parts (declaring permissions, prompt user to grant permission) together. Will submit PR again.

PR #1816, opened 2020-01-11
Feature atomicoperations

### LINKED ISSUE #1804: Make operations atomic and undoable
Currently most operations when they fail make no attempt to revert the state of what was changed to be as it was before the operation was made. I propose making operations so that they have a "do()" "revert()" and "errorRevert()" where errors can be handled gracefully.

EmmanuelMess:
Adds an Operator that runs AbstractOperations, these are a way of encapsulating any task that might need an undo(), and a check(). It also allows easy for sequencing of recursive tasks with a requires(AbstractOperation).

It can also make the async part of the app simpler, if every task is an AbstractOperation it can be run as if they are all the same, merging all Services into one OperatorService.

Fixes #1804.


app/src/main/java/com/amaze/filemanager/filesystem/operations/CopyOperation.java
```diff
@@ -0,0 +1,266 @@
...
+	protected boolean check() {
+		return true;
+	}
+
+	@Override
+	protected void operate() throws IOException {
+		// initial start of copy, initiate the watcher
+		watcherUtil.watch(copyService);
+
+		if (FileUtil.checkFolder((targetPath), context) == 1) {
```
bowiechen: `(targetPath)` => `targetPath`

app/src/main/java/com/amaze/filemanager/filesystem/operations/CopyOperation.java
```diff
@@ -0,0 +1,266 @@
...
+					} else {
+
+						// the target open mode is where we're currently at
+						hFile = new HybridFile(mode, targetPath, sourceFiles.get(i).getName(),
+								f1.isDirectory());
+					}
+
+					if (progressHandler.getCancelled()) {
+						throw new CancellationIOException();
+					}
```
bowiechen: Throwing an exception in a `try` block is generally considered a bad practice. I think we can refactor this out.

app/src/main/java/com/amaze/filemanager/filesystem/operations/CopyOperation.java
```diff
@@ -0,0 +1,266 @@
...
+					progressHandler.setSourceFilesProcessed(++sourceProgress);
+					progressHandler.setFileName(sourceFiles.get(i).getName());
+					copyRoot(sourceFiles.get(i), hFile, move);
+
+                    /*
+                    if(checkFiles(new HybridFile(sourceFiles.get(i).getMode(), path),
+                            new HybridFile(OpenMode.ROOT,targetPath+"/"+name))) {
+                        failedFOps.add(sourceFiles.get(i));
+                    }
+                    */
```
bowiechen: Is this comment necessary?

app/src/main/java/com/amaze/filemanager/filesystem/operations/CopyOperation.java
```diff
@@ -0,0 +1,266 @@
...
+			throw new IOException("Copy is impossible!");
+		}
+
+		// making sure to delete files after copy operation is done
+		// and not if the copy was cancelled
+		if (move && !progressHandler.getCancelled()) {
+			for (HybridFileParcelable a : sourceFiles) {
+				if (!failedFOps.contains(a)) {
+					DeleteOperation delete = new DeleteOperation(context, isRootExplorer, a);
+					requires(delete);
```
bowiechen: Isn't there an `undo` method that does this?

app/src/main/java/com/amaze/filemanager/filesystem/operations/exceptions/CancellationIOException.java
```diff
@@ -0,0 +1,7 @@
+package com.amaze.filemanager.filesystem.operations.exceptions;
+
+import java.io.IOException;
+
+public class CancellationIOException extends IOException {
+
+}
```
bowiechen: Should this have a default constructor or an explicitly defined constructor like `ShellNotRunningIOException`?

app/src/main/java/com/amaze/filemanager/filesystem/operations/extract/TarExtractOperation.java
```diff
@@ -73,13 +74,14 @@ protected void extractWithFilter(@NonNull Filter filter) throws IOException {
...
-                //TAR is sequential, you need to walk all the way to the file you want
-                while (entry.hashCode() != inputStream.getNextTarEntry().hashCode());
-                extractEntry(context, inputStream, entry, outputPath);
-            }
+            crashOnCancelled();
+
+            listener.onUpdate(entry.getName());
+            //TAR is sequential, you need to walk all the way to the file you want
+            while (entry.hashCode() != inputStream.getNextTarEntry().hashCode()) ;
+            extractEntry(context, inputStream, entry, outputPath);
```
bowiechen: I'm seeing this piece of code repeated multiple times. Perhaps this can be abstracted into a helper method in another class (`ExtractOperationHelper`?) and be referenced from there. This will be more maintainable in the long run.

app/src/main/java/com/amaze/filemanager/filesystem/operations/extract/XzExtractOperation.java
```diff
@@ -66,12 +68,12 @@ protected void extractWithFilter(@NonNull Filter filter) throws IOException {
...
-                //TAR is sequential, you need to walk all the way to the file you want
-                while (entry.hashCode() != inputStream.getNextTarEntry().hashCode());
-                extractEntry(context, inputStream, entry, outputPath);
-            }
+            crashOnCancelled();
+
+            listener.onUpdate(entry.getName());
+            //TAR is sequential, you need to walk all the way to the file you want
+            while (entry.hashCode() != inputStream.getNextTarEntry().hashCode()) ;
+            extractEntry(context, inputStream, entry, outputPath);
```
bowiechen: See comment in `TarExtractOperation`.

app/src/main/java/com/amaze/filemanager/filesystem/operations/extract/LzmaExtractOperation.java
```diff
@@ -67,12 +69,12 @@ protected void extractWithFilter(@NonNull Filter filter) throws IOException {
...
-                //TAR is sequential, you need to walk all the way to the file you want
-                while (entry.hashCode() != inputStream.getNextTarEntry().hashCode());
-                extractEntry(context, inputStream, entry, outputPath);
-            }
+            crashOnCancelled();
+
+            listener.onUpdate(entry.getName());
+            //TAR is sequential, you need to walk all the way to the file you want
+            while (entry.hashCode() != inputStream.getNextTarEntry().hashCode()) ;
+            extractEntry(context, inputStream, entry, outputPath);
```
bowiechen: See comment in `TarExtractOperation`.

app/src/main/java/com/amaze/filemanager/filesystem/operations/extract/GzipExtractOperation.java
```diff
@@ -74,12 +75,12 @@ protected void extractWithFilter(@NonNull Filter filter) throws IOException {
...
-                //TAR is sequential, you need to walk all the way to the file you want
-                while (entry.hashCode() != inputStream.getNextTarEntry().hashCode());
-                extractEntry(context, inputStream, entry, outputPath);
-            }
+            crashOnCancelled();
+
+            listener.onUpdate(entry.getName());
+            //TAR is sequential, you need to walk all the way to the file you want
+            while (entry.hashCode() != inputStream.getNextTarEntry().hashCode()) ;
+            extractEntry(context, inputStream, entry, outputPath);
```
bowiechen: See comment in `TarExtractOperation`.

app/src/main/java/com/amaze/filemanager/filesystem/operations/singlefile/CreateFileOperation.java
```diff
@@ -0,0 +1,170 @@
...
+			}
+		}
+
+		errorCallBack.done(file, true);
+	}
+
+	@Override
+	protected void undo() {
+
+	}
```
bowiechen: Delete the file when undoing a file creation?

app/src/main/java/com/amaze/filemanager/filesystem/operations/singlefile/DeleteOperation.java
```diff
@@ -0,0 +1,114 @@
...
+		this.context = context;
+		this.rootMode = rootMode;
+
+		this.file = file;
+	}
+
+	@Override
+	protected boolean check() {
+		return true;
+	}
```
bowiechen: Check for the existence of the file to delete?

bowiechen:
I would also recommend keeping the PR smaller, preferably something that can be reviewed in 10-20 minutes, that way we don't run the risk of forgetting what was reviewed at the top of the files list.

app/src/main/java/com/amaze/filemanager/filesystem/operations/AbstractOperation.java
```diff
@@ -0,0 +1,67 @@
...
+
+	private Operator getOperator() {
+		return operator;
+	}
+
+	/**
+	 * If this check passes the operation will be started,
+	 * nothing here is reversable and this operation can have no side effects!
+	 */
+	protected abstract boolean check();
```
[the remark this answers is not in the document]
EmmanuelMess: It is a function that is run before the start and if it returns null it stops anything from happening. You can check anything from names to recursive directory copies here.

Keep in mind the only way to stop an operation later is throwing an exception.

app/src/main/java/com/amaze/filemanager/filesystem/operations/AbstractOperation.java
```diff
@@ -0,0 +1,67 @@
...
+
+	/**
+	 * The proper operation,
+	 * if another operation is required to be started call requires() and it will eventually start
+	 */
```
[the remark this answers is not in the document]
EmmanuelMess: "When this function is called, all prerequisite operations will execute 
before this operation executes." that is not correct, requires runs code _after the operation ends_.

app/src/main/java/com/amaze/filemanager/filesystem/operations/exceptions/CancellationIOException.java
EmmanuelMess: Should this have a default constructor, it does, the default constructor is implicit :P

app/src/main/java/com/amaze/filemanager/filesystem/operations/CopyOperation.java
EmmanuelMess: Undo reverses the entire operation (deleting the copies), this deletes the source files.

app/src/main/java/com/amaze/filemanager/filesystem/operations/extract/TarExtractOperation.java
EmmanuelMess: They are not exactly the same code, and the boilerplate code necessary is too much, as is each extraction is separated and clean in its own file.

app/src/main/java/com/amaze/filemanager/filesystem/operations/AbstractOperation.java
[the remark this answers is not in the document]
bowiechen: Nice. I would recommend saying something like 
```plain
This method checks if the prerequisite actions/operations are ready.
```

app/src/main/java/com/amaze/filemanager/filesystem/operations/exceptions/CancellationIOException.java
bowiechen: Right, I see a constructor defined for `ShellNotRunningIOException` and `ZipIOException` but not this one. Is there a good reason why?

app/src/main/java/com/amaze/filemanager/filesystem/operations/CopyOperation.java
bowiechen: Ah, I see what you're getting at. This class also doubles as `MoveOperation`, correct?

app/src/main/java/com/amaze/filemanager/filesystem/operations/AbstractOperation.java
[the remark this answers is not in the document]
bowiechen: So if a caller calls `op1.requires(op2)`, what is the correct order here?
```
a) op1 -> op2
b) op2 -> op1
```

My understanding is that a prerequisite task is checked by calling `op.check()` to ensure the Operation is ready to be executed, but I thought `requires` adds prerequisite tasks?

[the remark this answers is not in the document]
EmmanuelMess: > So if a caller calls `op1.requires(op2)`, what is the correct order here?
> 
> ```
> a) op1 -> op2
> b) op2 -> op1
> ```

a

> My understanding is that a prerequisite task is checked by calling `op.check()` to ensure the Operation is ready to be executed, and I thought `requires` adds prerequisite tasks?

I made mistakes choosing the correct words. 

`check` makes sure the current state of the filesystem is correct for the operation that is going to be performed, and check that the operation can actually be performed.
`requires` ads post-requisite tasks.

This is done this way so that operations of folders (that are normally recursive) can be done sequentially.

For example: copying:
	_Check_ stuff, i.e. that the destination is writable.
	If copying a folder: Create parent folder. -> _require_ copy of children.
	If copying a file: Create child, copy contents.

If the `require` wasn't there the operation would be recursive, as is exists it can be run sequentially by the Operator.

app/src/main/java/com/amaze/filemanager/filesystem/operations/CopyOperation.java
EmmanuelMess: > Ah, I see what you're getting at. This class also doubles as `MoveOperation`, correct?

Yes.

app/src/main/java/com/amaze/filemanager/filesystem/operations/exceptions/CancellationIOException.java
EmmanuelMess: > Right, I see a constructor defined for `ShellNotRunningIOException` and `ZipIOException` but not this one. Is there a good reason why?

Both of those Exception require an exception to be passed as argument. `ShellNotRunningIOException` needs a `ShellNotRunningException` and `ZipIOException` needs a `ZipException`.

app/src/main/java/com/amaze/filemanager/filesystem/operations/CopyOperation.java
EmmanuelMess: How?

EmmanuelMess:
@bowiechen Should I split this PR into smaller ones?

VishalNehra:
Needs major rebase. Closing.

PR #1825, opened 2020-01-30
Revert ZipService to depends on java.util.zip instead of zip4j

TranceLove:
Fixes regression of corrupted zip files after migrated to zip4j.

Additionally, sends a ACTION_MEDIA_SCANNER_SCAN_FILE intent so the generated zip file is visible when device is accessible as storage device over USB.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/asynchronous/services/ZipService.java
+++ b/app/src/main/java/com/amaze/filemanager/asynchronous/services/ZipService.java
@@ -28,37 +28,39 @@ import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.content.SharedPreferences;
+import android.net.Uri;
 import android.os.AsyncTask;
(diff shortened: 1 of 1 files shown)


TranceLove:
Yes, it had been tested on Oneplus 2 running Slim7 (7.1.2), LG Nexus 5x running AOSPExtended (9.0), and Sony Xperia T2 Ultra running stock rom (5.1.1). Amaze was able to

- create zip file with randomly selected files in Amaze
- created zip file can be visible to file manager (Linux Mint Tracia Cinnamon with Nemo file manager)
- created zip file can be opened properly

VishalNehra:
Does this not require any test case changes?

TranceLove:
> Does this not require any test case changes?

No test case related to ZipService as I aware of.

I did saw tests (Robolectric ones) are failing at master though. Will submit another PR for that.

PR #1832, opened 2020-02-08
Ignore autogenerated strings.xml

EmmanuelMess:
strings-iw and strings-in are generated on compilation, so ignoring them in github is a good idea IMO.


VishalNehra:
How are they getting generated on compilation?

TranceLove:
> How are they getting generated on compilation?

See #1803

PR #1860, opened 2020-03-09
Use package name as CONTROL_FTP_SERVER permission & FileProvider prefix

TranceLove:
Hardcoding `CONTROL_FTP_SERVER` permission name and `FileProvider` name prefix prevents release and debug/developer build co-existence. This commit changes them to use app package name as prefix.


TranceLove:
Two ticks, merging.

PR #1864, opened 2020-03-26
Use getStorageVolumes()

paulo-raca:
Use getStorageVolumes() instead of environment variables on Android >= N

There are a few advantages:
- Better APIs
- Nicer names to every volume
- USB storage "just works", no need to fiddle with the extra classes and permissions

There are, however, a few disadvantages:
- Cannot reliably distinguish SD cards and USB devices: It relies on device name or path containing "USB" to use the USB icon
- The StorageVolume.mPath directory is a hidden API and requires reflection to be accessed (Clumsy, but works reliably)

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/activities/MainActivity.java
+++ b/app/src/main/java/com/amaze/filemanager/activities/MainActivity.java
@@ -632,7 +632,7 @@ public class MainActivity extends PermissionsActivity implements SmbConnectionLi
             StorageManager sm = getSystemService(StorageManager.class);
             for (StorageVolume volume : sm.getStorageVolumes()) {
                 Log.i("MainActivity", "Volume: " + volume + "");
-                if (!volume.getState().equals(Environment.MEDIA_MOUNTED)) {
+                if (!volume.getState().equalsIgnoreCase(Environment.MEDIA_MOUNTED) && !volume.getState().equalsIgnoreCase(Environment.MEDIA_MOUNTED_READ_ONLY)) {
                     continue;
                 }
                 File path = Utils.getVolumeDirectory (volume);
--- a/app/src/main/java/com/amaze/filemanager/activities/MainActivity.java
+++ b/app/src/main/java/com/amaze/filemanager/activities/MainActivity.java
@@ -652,7 +652,12 @@ public class MainActivity extends PermissionsActivity implements SmbConnectionLi
                 icon = R.drawable.ic_phone_android_white_24dp;
             } else {
                 // There is no reliable way to distinguish USB and SD external storage
-                icon = R.drawable.ic_sd_storage_white_24dp;
+                // However it is often enough to check for "USB" String
+                if (name.toUpperCase().contains("USB") || path.getPath().toUpperCase().contains("USB")) {
+                    icon = R.drawable.ic_usb_white_24dp;
+                } else {
+                    icon = R.drawable.ic_sd_storage_white_24dp;
+                }
             }
             volumes.add(new StorageDirectoryParcelable(path.getPath(), name, icon));
         }
--- a/app/src/main/java/com/amaze/filemanager/activities/MainActivity.java
+++ b/app/src/main/java/com/amaze/filemanager/activities/MainActivity.java
@@ -22,6 +22,7 @@
 
 package com.amaze.filemanager.activities;
 
+import android.annotation.TargetApi;
 import android.app.Activity;
 import android.app.ActivityManager;
 import android.content.BroadcastReceiver;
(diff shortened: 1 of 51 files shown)


TranceLove:
Before going further, we may want to fix the problem of differentiate between storage types in the PR.

Didn't look into the code yet, but I did saw LineageOS's file manager can differentiate between storage and SD card.

Screenshot_20200327-174602_

(Fairphone 3 running LineageOS 16.0 GSI)

EmmanuelMess:
> Cannot distinguish SD cards and USB devices, so the icons won't match

Why?

EmmanuelMess:
Will separate the getStorageDirectories into two functions

EmmanuelMess:
> Nicer names to every volume

Where do those come from?

paulo-raca:
>> Cannot distinguish SD cards and USB devices, so the icons won't match

> Why?

The StorageVolume class has limited metadata on the volume: `isRemovable()`, `isEmulated()`, `isPrimary()`. I'm only using `isRemovable()` to determine the icon

>> The StorageVolume.mPath directory is a hidden API and requires reflection to be accessed

> This has to be tested for every API version.

Tested and works :)

> Will separate the getStorageDirectories into two functions

Sure

>> Nicer names to every volume

> Where do those come from?

`StorageVolume.getDescription()` returns "Internal Storage" for the the built-in flash and the partition label for external devices. E.g., "Drive USB Kingston"

paulo-raca:
> Didn't look into the code yet, but I did saw LineageOS's file manager can differentiate between storage and SD card.

Apparently LineageOS looks for the USB string on the device path. It doesn't work on any of my devices, so it is probably a LineageOS tweak, but I can replicate and also test it on the device name (Which should work for the default labels on most USB Drives)

EmmanuelMess:
> Apparently LineageOS looks for the USB string on the device path. It doesn't work on any of my devices, so it is probably a LineageOS tweak, 

What do you mean "It doesn't work on any of my devices"?

paulo-raca:
I mean that I don't get "USB" string in the mount path in any device I tested. IIRC, the path is something like `/mnt/rw/uuid`

EmmanuelMess:
> I mean that I don't get "USB" string in the mount path in any device I tested. 

And in the name?

app/src/main/java/com/amaze/filemanager/activities/MainActivity.java
```diff
@@ -612,18 +616,65 @@ private void initializeInteractiveShell() {
...
+    /**
+     * @return All available storage volumes (including internal storage, SD-Cards and USB devices)
+     */
+    @TargetApi(Build.VERSION_CODES.N)
+    public synchronized ArrayList<StorageDirectoryParcelable> getStorageDirectoriesNew() {
+        // Final set of paths
+        ArrayList<StorageDirectoryParcelable> volumes = new ArrayList<>();
+        StorageManager sm = getSystemService(StorageManager.class);
+        for (StorageVolume volume : sm.getStorageVolumes()) {
+            Log.i("MainActivity", "Volume: " + volume + "");
```
EmmanuelMess: Please delete this log call. (If you allow contributor changes I can push changes directly)

paulo-raca:
The default name is usually something like "Somebrand USB Drive".

That worked for me. It isn't very reliable, but I suppose it is good enough

EmmanuelMess:
We should make a version of `StorageVolume.getDescription()` that works for older APIs, so that names are consistent.

app/src/main/java/com/amaze/filemanager/activities/MainActivity.java
paulo-raca: I won't be able to do it today, but AFAICT contributor changes are enabled.

Thanks!

paulo-raca:
> Edit: working on this.
Thanks!

EmmanuelMess:
> > Edit: working on this.
> 
> Thanks!

Made a PR I think.

EmmanuelMess:
@paulo-raca any news?

paulo-raca:
@EmmanuelMess, sorry, I hadn't noticed your PR before

app/src/main/res/values/strings.xml
```diff
@@ -650,5 +649,11 @@
     <string name="cloud_plugin_google_play_uri" translatable="false">market://details?id=com.filemanager.amazecloud</string>
     <string name="cloud_plugin_google_play_web_uri" translatable="false">https://play.google.com/store/apps/details?id=com.filemanager.amazecloud</string>
     <string name="failed_install_apk">Failed to install</string>
+
+    <!-- from https://android.googlesource.com/platform/frameworks/base/+/master/core/res/res/values/strings.xml -->
+    <!-- Storage description for internal shared storage. [CHAR LIMIT=NONE] -->
+    <string name="storage_internal">Internal shared storage</string>
```
VishalNehra, line 655: Can we keep the original "Internal Storage" instead? This will look bad on low ppi devices.

paulo-raca: This change came from @EmmanuelMess, but I agree that "Internal Storage" looks better

TranceLove:
Tested on Fairphone 3 running LineageOS 16 GSI, after applying changeset storage types can still be identified. However, it cannot return the partition label of the inserted micro SD card.

Before:
Screenshot_20200415-160849_Amaze_Debug

After:
(TBC)

Changeset had two ticks, it can be merged at will however.

EmmanuelMess:
>  However, it cannot return the partition label of the inserted micro SD card.

No, it can't.

paulo-raca:
Thank you :)

PR #1867, opened 2020-04-02
Fix parsing output from mount command to fix mounting system partitions rw

### LINKED ISSUE #1835: KitKat 4.4: Renaming a system file doesn't work (even if Supersu grant root access)
On Kitkat 4.4, renaming a system file doesn't work (even if Supersu grant root access to amaze).

It works using the FX File Explorer app (it only asks before to "mount read write").

(I could use FX File Explorer, but I prefer to use your open source app)

ps: thanks for your awesome work!

TranceLove:
Fixes #1835.

Corrects the parsing of output from `mount` command when attempt to mount corresponding system partitions read-write.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/utils/RootUtils.java
+++ b/app/src/main/java/com/amaze/filemanager/utils/RootUtils.java
@@ -1,12 +1,29 @@
+/*
+ * RootHelper.java
+ *
+ * Copyright (C) 2016-2020 Arpit Khurana <arpitkh96@gmail.com>,
+ * Vishal Nehra <vishalmeham2@gmail.com>, Emmanuel Messulam<emmanuelbendavid@gmail.com>,
+ * John Carlson <jawnnypoo@gmail.com>, Raymond Lai <airwave209gt at gmail.com> and Contributors.
+ *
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/amaze/filemanager/utils/RootUtils.java
```diff
@@ -41,27 +58,41 @@
...
         for (String line : output) {
             String[] words = line.split(" ");
 
-            if (path.contains(words[2])) {
+            // mount command output for older Androids
+            // <code>/dev/block/vda /system ext4 ro,seclabel,relatime,data=ordered 0 0</code>
+            String mountPointOutputFromShell = words[1];
+            String mountPointFileSystemTypeFromShell = words[2];
+            String mountPointArgumentFromShell = words[3];
+            if(Build.VERSION.SDK_INT > Build.VERSION_CODES.M) {
```
EmmanuelMess, line 70: Not >=?

TranceLove: Good catch. Will clarify this before merging.

TranceLove:
Two ticked, merging.

Issue #1862, opened 2020-03-10
app crashes on searching "android" in folder section


NayanAnsh:
when in internal storage section searching  the word "android" cause  the app to crash.

**To Reproduce**
1. at internal storage page search "android"

**Expected behavior**
it Should show the folders containing words "android"

**Screenshots**

**Smartphone (please complete the following information):**
I have tested  this on two devices-
 - Device: realme x2 , moto g5 plus
 - OS:  Android 9 ,android 8.1
 - Rooted: Nope
 versionName- "3.3.0RC1"

marcinadd:
This is probably the same issue as #1837.

PR #1871, opened 2020-04-14
Fix RecyclerPreloadModelProvider crashing while searching #1862

### LINKED ISSUE #1837: file search crashing
Mostly when searching a keyword, Amaze crashes before showing it.

Amaze version: 3.4.2

marcinadd:
Looks like a problem with RecyclerPreloadModelProvider.
It's difficult to debug since when you use debugger it works like a charm. Maybe kind of a race condition or something. It can be fixed by disabling RecyclerPreloadModuleProvider or with simple if-else clause.

The issue is not only related to the "android" keyword. I was able to reproduce it when there is more than one search result but it's hard to reproduce for me when I used debugger. I tried to synchronize methods but my knowledge is too little to be able to do it.

Idk maybe it is a more elegant way.

Fixes #1837


EmmanuelMess:
Maybe it has to do with lazy loading.

VishalNehra:
Not able to reproduce the issue, but keeping this check in place doesn't hurt. Will avoid `IndexOutOfBound` exception. Approving.

PR #1878, opened 2020-04-16
chore(string): update zh-rCN string resource

Bambooin:


EmmanuelMess:
Translations should be made via [Transifex](https://www.transifex.com/amaze/amaze-file-manager/dashboard/). All changes will be overwritten.

PR #1212, opened 2018-05-04
Add support for FTP server saving to external storage/SD card

TranceLove:
Fixes #1008.

Included a set of filesystem integration classes specific to Android devices, based on `DocumentFile`, so file I/O will be done with ContentResolver; also cherry-picked 2 commits in #1087 for choosing FTP server's shared path with FolderChooser.

I suspect the SlimLP (5.1.1) I use on my Galaxy S2 was short-circuited that write request to micro SD card slot didn't trigger SAF popup, even with a fresh install. While change FTP server share path worked on Nougat emulator (Pixel 2) with external SD card, please help test on true devices having SD card slot and running Android >= 5.1.1, to verify the correctness of the implementation, especially uploading file to FTP server with shared path on the external SD card.

Issue #1008, opened 2018-02-09
no ftp file transfer into sdcard


Meresmata:
Hi,
  while I am am to transfer files from android to sdcard and vice versa. I can only transfer file via ftp into the android part, but not into the sdcard!

Using a Sony Z1, Android 7 or 8 (updated in last days.)

TranceLove:
Actually it is possible, but need to make FtpServer be able to write to external storage using Android Storage Access Framework, which requires a set of integration classes.

Will look into it, but expect it happens past 3.3.0 milestone.

PR #1895, opened 2020-06-06
Use FolderChooser for FTP shared path

### LINKED ISSUE #811: Change FTP server shared path with dir chooser

VishalNehra:
Duplicate of #1087 

Fixes #811, using MaterialDialog's FolderChooser.

While working on #811, also took the time to change archive extraction and zip creation paths to use FolderChooser too.

Sorry for scattered callbacks of the folder chooser dialogs, but seems it's the way for FolderChooser to work.


app/src/main/java/com/amaze/filemanager/fragments/FtpServerFragment.java
```diff
@@ -82,6 +82,8 @@
     private MainActivity mainActivity;
 
     private TextView statusText, url, username, password, port, sharedPath;
+    public static final String TAG = "FTPServerFragment";
```
TranceLove, line 85: Perhaps `FtpServerFragment.class.getSimpleName()` is better?

TranceLove:
Code looks OK, on Fairphone 3 running LineageOS 16.0 GSI, choosing either internal or external storages.

Note however before #1008 is fixed end user will have problem with saving files to external SD card via FTP server - folder dialog won't restrict user from choosing external storage, reading is also OK, but writing is not - end user may not be expecting this.

#1212 (for #1008) had been outdated... also may want to properly return error to end users on unable to write to storage (either internal or external).

May not need to fix in this PR and leave it to a newer #1212 IMO, so I will give a tick here.

Also I may want to answer comments in #1087 by @EmmanuelMess since that PR had been closed, for the record.

> Archive preferences don't seem to do anything.

Will become useful when Amaze is capable of opening archives from `content://` URIs directly.

> There is no way of returning to the 'default mode' (extracting in compressed folder).

Agreed.

> Hidden folders show even if the option (in settings) is deactivated.

Not sure if the setting should be affecting the folder selector here too.

Archive related questions should be addressed in another PR too.

app/src/main/java/com/amaze/filemanager/fragments/FtpServerFragment.java
VishalNehra: The code in MainActivity#2044 doesn't seem to support dynamic method call :P This will be required anyways.

Issue #1896, opened 2020-06-07
Reduce code smells; refactor code


VishalNehra:
Using this issue to clock all the refactoring and reducing code smells.

PR #1897, opened 2020-06-07
Feature/1896

VishalNehra:
For #1896 Adding spotless gradle support to make code neat and clean, added step in travis as well.
Things to do, before sending PR for approval, just run 
`./gradlew spotlessJavaCheck`
in local to check for formatting errors. If it finds any then run following command, it'll automatically fix
`./gradlew :spotlessApply`

This build will probably fail, so need to run `spotlessApply` on master branch, will do that in separate PR as the initial changes would be project-wide.

### DIFF
--- a/spotless.license-java
+++ b/spotless.license-java
@@ -17,3 +17,4 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
+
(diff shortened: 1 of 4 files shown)


spotless.license-java
```diff
@@ -0,0 +1,19 @@
+/*
```
TranceLove, line 1: I used to add the filename in the license header here. Can we do the same... and is it a good idea?

VishalNehra: Doesn't seem like spotless support any such placeholder as of now. Maybe we can create a feature request for them.

Issue #1918, opened 2020-07-02
Upgrade to Robolectric 4


TranceLove:
Upgrade to Robolectric 4 which supports `androidx.test` out of the box and supports newer SDKs (up to 27 as of Robolectric 3.8).

PR #1931, opened 2020-07-15
Display snackbar when there is no network connectivity

### LINKED ISSUE #1920: FTP server cannot be used without WiFi, and there is no prompt
**Describe the bug**
When I use the data network, I can't turn on the FTP server, and there is no prompt. I can only see that the start button is gray. After trying it out, I found that this feature can only be used on WiFi networks.

**To Reproduce**
1. Install and open AmazeFileManager
2. Open menu bar
3. Scroll down to 'FTP server'
4. Click on 'FTP server'
5. Unable to click the start button

**Expected behavior**
Users should be prompted to turn on WiFi when there is no WiFi network, otherwise they don't know why they can't use FTP server

**Screenshots**
ezgif com-video-to-gif (3)

 - Device: galaxy a6s
 - OS: Android 8.1.0
 - Rooted: No
 - Version: v3.4.3

TranceLove:
#### Issue tracker   

Fixes #1920

#### Release  
Addresses `release/3.6`
  
#### Test cases
- [ ] Covered
(Sorry, AndroidX fragment testing requires Robolectric 4 to be available. See #1918)
  
#### Manual testing
- [x] Done  
  
If yes,  
- Device: Fairphone 3
- OS: LineageOS 16.0 GSI(9.0)

#### Build tasks success  

#### Related PR  
Related to PR #

#### Additional Info


app/src/main/java/com/amaze/filemanager/fragments/FtpServerFragment.java
```diff
@@ -656,4 +666,38 @@ private void setSecurePreference(boolean isSecureEnabled) {
...
+
+  private void promptUserToEnableWireless(@Nullable NetworkInfo ni) {
+    // No wifi, no data, no connection at all
+    if (ni == null || !ni.isConnected()) {
+      showSnackbar(R.string.ftp_server_prompt_connect_to_network, new Intent(ACTION_WIFI_SETTINGS));
+    } else {
+      // Data connection available, but no AP enabled
+      if (ni.getType() == ConnectivityManager.TYPE_MOBILE) {
+        showSnackbar(
+            R.string.ftp_server_prompt_open_ap, new Intent(Settings.ACTION_WIRELESS_SETTINGS));
```
VishalNehra, line 678: I think regardless of whether mobile data is enabled or not, we can always prompt user to open settings and enable wifi or start a hotspot.
Similarly, only one message should suffice 
`Connect to WiFi or enable hotspot`

TranceLove: But what if the two snackbars would take user to different places in settings? As you can see here, `Open settings` would take user to wifi settings if device has no connection at all, or mobile network settings if user has mobile data enabled but didn't enable AP.

VishalNehra: Maybe just open him to enable wifi screen like you're doing now in case there's no network.
I mean, mobile network shouldn't matter because either way you have to enable wifi or hotspot.

PR #1948, opened 2020-08-08
Fix broken symlinks on SSH server causing incomplete file listings

TranceLove:
Fix by adding a try/catch here can keep the loop to stay at HybridFile.forEachChildrenFile() for SSH servers.

#### Release  
Addresses release/3.5
  
#### Test cases
- [x] Covered
  
#### Manual testing
- [x] Done  
  
If yes,  
- Device: Fairphone 3
- OS: Lineage OS 16.0 (9.0)

#### Build tasks success  


VishalNehra:
Seems to have a few issues on codacy.
The one with `JUnit tests should include assert() or fail()` is not right I feel. Maybe we can bypass it?

TranceLove:
> Seems to have a few issues on codacy.
> The one with `JUnit tests should include assert() or fail()` is not right I feel. Maybe we can bypass it?

If all they need is an `assert`, I can just make the common verification routine to `return true` instead of being a `void`.

PR #1996, opened 2020-10-02
SSH move/back navigation fixes

TranceLove:
## PR Info
- Fix SFTP routine template in `HybridFile.forEachChildrenFile()` to return `true` instead of `null`, to workaround RxJava's restriction that callable passed in must not return null, prevent app crash when moving file from device to SSH server
- Wrap exception that may throw when closing remote outputstream on SSH server into `Log.w()` to prevent exception thrown that might crash the app
- Fix grepping path for SFTP to fix back button navigation on server, that previously user cannot go to parent folder on SSH server when pressing back button

#### Release  
Addresses release/3.5
  
#### Test cases
- [x] Covered
  
#### Manual testing
- [x] Done  
  
If yes,  
- Device: Fairphone 3
- OS: Lineage OS 16.0 (9.0)

#### Build tasks success  

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/filesystem/HybridFile.java
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/HybridFile.java
@@ -376,10 +376,14 @@ public class HybridFile {
       case ROOT:
         parentPath = getFile().getParent();
         break;
-      case OTG:
-      default:
+      case SFTP:
         StringBuilder builder = new StringBuilder(path);
         StringBuilder parentPathBuilder =
+            new StringBuilder(builder.substring(0, builder.length() - (getName(context).length())));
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/amaze/filemanager/filesystem/HybridFile.java
```diff
@@ -376,10 +376,14 @@ public String getParent(Context context) {
       case ROOT:
         parentPath = getFile().getParent();
         break;
-      case OTG:
```
EmmanuelMess, line 379: What happened to case OTG?

EmmanuelMess (changes requested):

EmmanuelMess (approved):

PR #2017, opened 2020-10-18
Fixes from codacy

EmmanuelMess:
#### Issue tracker   

Fixes #
-or-   
Addresses #

#### Release  
Addresses release/
  
#### Test cases
- [ ] Covered
  
#### Manual testing
- [ ] Done  
  
If yes,  
- Device:
- OS:

#### Build tasks success  

#### Related PR  
Related to PR #

#### Additional Info

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/TagsHelper.java
+++ b/app/src/main/java/com/amaze/filemanager/TagsHelper.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright (C) 2014-2020 Arpit Khurana <arpitkh96@gmail.com>, Vishal Nehra <vishalmeham2@gmail.com>,
+ * Emmanuel Messulam<emmanuelbendavid@gmail.com>, Raymond Lai <airwave209gt at gmail.com> and Contributors.
+ *
+ * This file is part of Amaze File Manager.
+ *
+ * Amaze File Manager is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.amaze.filemanager;
+
+public final class TagsHelper {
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
```diff
@@ -476,8 +478,9 @@ public boolean onPrepareActionMode(ActionMode mode, Menu menu) {
             return true;
           }
 
-          if (getMainActivity().mReturnIntent)
-            if (Build.VERSION.SDK_INT >= 16) showOption(R.id.openmulti, menu);
+          if (getMainActivity().mReturnIntent && Build.VERSION.SDK_INT >= 16) {
```
TranceLove, line 481: Perhaps `Build.VERSION.SDK_INT >= JELLY_BEAN` is better?

app/src/main/java/com/amaze/filemanager/ui/activities/MainActivity.java
```diff
@@ -168,6 +169,8 @@
         LoaderManager.LoaderCallbacks<Cursor>,
         FolderChooserDialog.FolderCallback {
 
+  private static final String TAG = "MainActivity";
```
TranceLove, line 172: Perhaps `MainActivity.class.getSimpleName()` is better?

TranceLove (dismissed):

app/src/main/java/com/amaze/filemanager/TagsHelper.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.amaze.filemanager;
+
+public class TagsHelper {
+  public static String getTag(Class<?> clazz) {
+    return clazz.getSimpleName().substring(0, 23);
```
TranceLove, line 25: Why you will need `substring()` when `Class.getSimpleName()` would only return the class' own name without package?

TranceLove (changes requested):

app/src/main/java/com/amaze/filemanager/TagsHelper.java
EmmanuelMess: Someone might have a class that is exceedingly long, especially in Kotlin or for nested classes, where the TAG will end up with >23 chars, witch will directly crash in production.

TranceLove: But what if the class name is shorter than 23 chars? This should be the reason why the tests ended up with `IndexOutOfBoundException`s

TranceLove (approved):

VishalNehra (approved):

PR #2173, opened 2020-12-31
ShareTask: Fix for sharing one single file

### LINKED ISSUE #2158: Using SEND_MULTIPLE Intent Action when only sending one file.
**Describe the bug**
https://github.com/TeamAmaze/AmazeFileManager/blob/master/app/src/main/java/com/amaze/filemanager/ui/dialogs/share/ShareTask.java#L63

I think it should be good to make a separate ShareTask for single file sharing, like a ShareSingleTask.

**Additional context**
I Was working on an app and I couldn't understand why my SEND intent-filter couldn't receive the file, not too  much of a problem but it still would be better to have the right intent on this particular case, my other files manager correctly use the SEND intent on single file sharing.

TranceLove:
#### Issue tracker   

Fixes #2158

#### Release  
Addresses release/3.6
  
#### Test cases
- [ ] Covered
  
#### Manual testing
- [ ] Done  
  
If yes,  
- Device: Pixel 3 emulator
- OS: Android 10

#### Build tasks success  

#### Additional Info
If one single file is being shared, uses Intent.ACTION_SEND instead of Intent.ACTION_SEND_MULTIPLE. Also put the only Uri in sharingUri list instead of the whole ArrayList.

Additionally, some variable name changes for better indicating their usage.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/ui/dialogs/share/ShareTask.java
+++ b/app/src/main/java/com/amaze/filemanager/ui/dialogs/share/ShareTask.java
@@ -43,63 +43,69 @@ public class ShareTask extends AsyncTask<String, String, Void> {
 
   private Activity contextc;
   private int fab_skin;
-  private ArrayList<Uri> arrayList;
+  private ArrayList<Uri> sharingUris;
   private ArrayList<Intent> targetShareIntents = new ArrayList<>();
-  private ArrayList<String> arrayList1 = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


EmmanuelMess:
Code convention says that the braces always are put for ifs even if they are single line.

EmmanuelMess:
Otherwise accepted.

VishalNehra (approved):

EmmanuelMess (approved):

PR #2175, opened 2021-01-01
Add option for SMB share name

### LINKED ISSUE #286: SMB Path Support
Adding a new SMB connection using the "use custom IP" option does not appear to allow specifying the server and path in the "Server IP Address" field.  When a '/' is used in the field the connection fails.  When editing an SMB connection that contains a '/', the '/' and any text after is dropped.  If "%2F" is used instead of '/' the value is retained during edits to the connection but the connection still fails.

Specifying the server and path is necessary for servers that do not advertise their shares.  Adding an SMB connection with a server and path is a feature that used to be supported by AmazeFileManager.

tnyeanderson:
This PR adds a "Share name" field to the SMB connection dialog, allows editing the connection after creation, and successfully connects to the hidden share :)

#### Issue tracker   

Fixes #286 
-or-   
Addresses #

#### Release  
Addresses hotfix/3.5.3
  
#### Test cases
- [ ] Covered
  
#### Manual testing
- [x] Done  
  
If yes,  
- Device: Pixel 2 Emulator
- OS: API 28

#### Build tasks success  

#### Related PR  
Related to PR #

#### Additional Info


tnyeanderson:
By fixing the formatting for "spotless" it appears the "one line declarations" rule is violated by codacy. Let me know if you would like me to fix or leave as-is. Thank you for the opportunity to contribute :)

tnyeanderson:
@TranceLove I see you assigned this issue to yourself. Just want to make sure you don't have to do any duplicate work so I'm mentioning you here

TranceLove:
@tnyeanderson No, never! I even didn't start yet😜 

I'll look into your code later. But meanwhile please do fix the one line declaration problem as suggested by codacy.

TranceLove (changes requested):
First of all thanks again for your work. Tried on Pixel 3 emulator running stock Android 10, it correctly connects to hidden samba share.

In addition to the codacy warning, you'll need to fix the share name textfield, which is currently set as a required field. This will prevent users from creating/updating SMB connections that does not need to explicitly specify the share's name. Please fix it.

Screenshot_1609517861

tnyeanderson:
@TranceLove I think I've removed the error properly, but I don't have any shares to test with that don't require a share name (so sorry to put that testing workload on you)

I can now create the connection without a share name, and creating it with a share name still works.

There are still "one line declaration" codacy errors, but the multiple declaration lines were already in the code before, I merely added to them... I can separate them all out on to multiple lines if you'd like, but I try not to make big changes to the formatting of a file without specific instruction.

In addition, spotless fails because there are lines where a variable value is to long, and it wants me to put the value on a separate line as shown:

```
varname =
  verylongvariablevalue
```

Would you like me to do that? I originally thought that this is what was causing the codacy error but I could be mistaken.

Sorry for the formatting questions, hope this code eases your workload a bit :)

app/src/main/java/com/amaze/filemanager/ui/dialogs/SmbConnectDialog.java
```diff
@@ -350,6 +354,9 @@ private SmbFile createSMBPath(String[] auth, boolean anonymous, boolean disableI
             .append(URLEncoder.encode(auth[2], "UTF-8"))
             .append("@");
       sb.append(yourPeerIP).append("/");
+      if (share != null || share.length() > 0) {
```
TranceLove, line 357: You may simply to use [TextUtils.isEmpty](https://developer.android.com/reference/android/text/TextUtils#isEmpty(java.lang.CharSequence)) for string empty/null check.

TranceLove (changes requested):
Checked changes, create SMB connection without share name is working now. Thanks.

One more change to go: please help prevent the redundant slash from saved into the database. I think it is to do with the null check for share name in SmbConnectDialog.

Screenshot_1609647031

For codacy warnings, I know the rules can be annoying sometimes - me for one ;), but please do follow them as others do for max readability to everyone. Thanks.

app/src/main/java/com/amaze/filemanager/ui/dialogs/SmbConnectDialog.java
tnyeanderson: Perfect, thanks for the tip (I am not an android dev by any stretch).... I will take care of those formatting issues as well tonight and hopefully we will have something acceptable by tomorrow :)

Thank you for the help and guidance!

tnyeanderson: By using `!TextUtils.isempty(share)` I believe this should fix the extra slash issue too (because of the `||` instead of `&&` in the original code, it was adding the slash even on a null/empty value). Though I can't double check this because I don't have any non-hidden shares :(

tnyeanderson:
Woohoo! Checks passing... if you could be so kind as to double check that there is no double slash with an empty share name, that would be great. Otherwise we should be good to go!

Thanks so much again for your help and guidance through this, and for giving me the opportunity to contribute. I've been using this app for years and this has been a pain point for just as long, so I'm delighted that I could help solve it :)

app/src/main/java/com/amaze/filemanager/ui/dialogs/SmbConnectDialog.java
```diff
@@ -230,7 +237,9 @@ public void afterTextChanged(Editable s) {
           chkSmbAnonymous.setChecked(true);
         }
         ipp = a.getHost();
+        sharep = a.getPath().replaceAll("/", "");
```
VishalNehra, line 240: Why replace all the slashes? Isn't it better to support a custom path instead of only root ones?
Eg. you're adding support for only `/A` or `/B` what if I want to add `/A/B/C` as Share path? You'll end up showing `ABC` as Share name to user. So I suggest just replace the first / occurrence while showing in edit text. 
That said, I think the feature text should be named `Share path` instead of `Share Name`

Would also request another change while you're at it.
Since we're adding one more new field to the dialog, due to increased number of fields it's better as UX standpoint to define which fields are optional and which are mandatory.

So request you to name `Share path (optional)` as the edit text hint, also check other fields as well. Eg. domain and name too can be optional. So you can add optional in their name as well.

VishalNehra (changes requested):

app/src/main/java/com/amaze/filemanager/ui/dialogs/SmbConnectDialog.java
tnyeanderson: Good call! That makes sense. Since I was only thinking of the share name originally, I figured there would be a slash before and a slash after because those are added to the share path URL (hence replace all slashes). By renaming the field to allow full paths (and clarify which fields are optional) I should only have to remove the first and last slash and it will give the user more flexibility... in fact, a flexibility that will benefit me as well! 

I will get these changes done today :)

tnyeanderson: Looking in the code (and through testing) it looks like the name is *not* optional, so I only labeled `Domain` and `Share path` as optional. You can now add a path and it will connect automatically to where you want.... brilliant!

Hate to bother again, but if you could just double check that an empty share name still connects properly, that would be great. Sorry that I don't have a non-hidden share and keep putting this burden on you :(

TranceLove (dismissed):
Checked again with Pixel 4 emulator running Android 10, no visible regressions found since last update.

Functionally I'd vote +1.

app/src/main/res/values/strings.xml
```diff
@@ -699,5 +699,6 @@
     <string name="legacy_listing_summary">If enabled, uses stat for listing, else uses ls</string>
     <string name="error_unsupported_v5_rar">RAR archive \"%s\" is unsupported RAR v5 archive.</string>
     <string name="ftp_prompt_restart_server">You will need to restart the FTP server for changes to take effect.</string>
+    <string name="smb_share">Share path (optional)</string>
```
VishalNehra, line 702: Wouldn't it be better to use string formatting and declare `optional` as a separate text?
`<string name="smb_share">Share path (%s)</string>
<string name="optional">optional</string>
`

VishalNehra (changes requested):

app/src/main/res/values/strings.xml
tnyeanderson: @VishalNehra  I worry this may not be as useful, because trying to reuse the domain and smb_share strings without them being optional (perhaps in the future) will result in an empty set of parenthesis. Perhaps something like this is better:

strings.xml
```
<string name="smb_share">Share path %s</string>
<string name="optional">(optional)</string>
```

smb_dialog.xml
```
android:hint="@{@string/smb_share(@string/optional)}"
```

(Please let me know if my syntax or understanding is off here. I am not an android dev and this is my first time using this type of string formatting in java xml files.... just trying to create the most extensible/reusable code here)

tnyeanderson: Also want to point out [this line](https://github.com/TeamAmaze/AmazeFileManager/blob/f7899f138504ef7a956da59a66a7a6b2d87932b5/app/src/main/java/com/amaze/filemanager/ui/dialogs/SmbConnectDialog.java#L131) which I *think* will cause a regression in the exact way I described (empty parenthesis)

VishalNehra:
Changes seem fine from my side. 
However, there seem to be a problem with existing code. Which I insist you to fix.
In case the path is not found / valid. The refreshing doesn't stop (even if we got SMBException in LoadFilesListTask).
Seem like MainFragment#955 is the issue, we're not handling null returns and not stopping the list refresh.
Can you please give a suitable prompt to user that the path is not correct? I prefer we open the same SMB dialog again (so that user don't have to reenter the details he just entered, and can just modify). May need to call MainFragment#reauthenticateSmb for this 🤔

app/src/main/res/values/strings.xml
VishalNehra: Agreed. In that case I think appending an asterisk on code side (not xml side) in front of mandatory fields would be better?

tnyeanderson:
I've added an error when the path fails to load and starts with `smb://`. Also changed the string name for the error message as it was already vague enough to not be describing what the string name entails (no mention of "decrypting") so I reused it for the generic SMB error.

I'm not sure how we can open the dialog from that context, because it needs the connection name and the path, but the scope of the function (I think) doesn't let us know what the connection name is. If you know of a way to get the connection name from here, let me know and I'll implement it.

VishalNehra:
Please rebase with 3.5.3 branch as well. There are 2 conflicts currently which will prevent the merge.

VishalNehra:
> I'm not sure how we can open the dialog from that context, because it needs the connection name and the path, but the scope of the function (I think) doesn't let us know what the connection name is. If you know of a way to get the connection name from here, let me know and I'll implement it.

Well I guess you can simply call what's been called in LoadFilesListTask#128 again in LoadFilesListTask#132 
You won't even need the changes in MainFragment#995 then 🤔

tnyeanderson:
Okay I don't want to push anything because I'm not sure if I did this right. Here's the commands I ran:

```
git remote add upstream https://github.com/TeamAmaze/AmazeFileManager
git fetch upstream hotfix/3.5.3
git rebase upstream/hotfix/3.5.3
```

Now I'm going through and resolving a bunch of conflicts.... is this how I'm supposed to do it? Then `git push`? Or is there a better way?

VishalNehra:
Yes, that's correct. You have to choose which version to keep when fixing conflicts, the upstream version or your version. 
You can use IDE's feature (right click any red file > git > resolve conflicts. In the menu you can choose which file, can choose which changes to merge.
After you're done from that menu it'll auto commit, you just perform `git push` on this branch.

tnyeanderson:
Okay so I aborted my original rebase, then started a new rebase on upstream/hotfix/3.5.3

Used Android studio to resolve all conflicts clicking "Accept theirs" on every prompt to accept the upstream changes.

Now the app won't build :(

VishalNehra:
Can you please join on telegram channel. Really difficult to find issue here 
t.me/AmazeFileManager

VishalNehra:
Too much conflicts here, created new PR 
https://github.com/TeamAmaze/AmazeFileManager/pull/2194

PR #2194, opened 2021-01-06
Add option for SMB share name

### LINKED ISSUE #286: SMB Path Support

VishalNehra:
## PR Info

This PR adds a "Share name" field to the SMB connection dialog, allows editing the connection after creation, and successfully connects to the hidden share :)

#### Issue tracker   

Fixes #286 
-or-   
Addresses #

#### Release  
Addresses release/
  
#### Test cases
- [ ] Covered
  
#### Manual testing
- [x] Done  
  
If yes,  
- Device:
- OS:

#### Build tasks success  

#### Related PR  
Related to PR #

#### Additional Info
Original PR 
https://github.com/TeamAmaze/AmazeFileManager/pull/2175


tnyeanderson:
Moving `makeRequired` to utils makes a lot of sense... and thanks for getting the conflicts and dialog figured out!

VishalNehra:
@tnyeanderson yes. Check telegram, I wrote about this there :)

app/src/main/java/com/amaze/filemanager/utils/Utils.java
```diff
@@ -399,4 +401,13 @@ public static Intent buildEmailIntent(String text) {
     emailIntent.setType("message/rfc822");
     return emailIntent;
   }
+
+  /**
+   * Marks a text input field as mandatory (appends * at end)
+   *
+   * @param field given text input field
+   */
+  public static void makeRequired(TextInputLayout field) {
```
EmmanuelMess, line 410: Please use extension function in kotlin in a new file called "TextInputLayoutExtensions" so that it is easier to find and use.

app/src/main/java/com/amaze/filemanager/utils/Utils.java
```diff
@@ -399,4 +401,13 @@ public static Intent buildEmailIntent(String text) {
     emailIntent.setType("message/rfc822");
     return emailIntent;
   }
+
+  /**
+   * Marks a text input field as mandatory (appends * at end)
```
EmmanuelMess, line 406: Does this enforce the "mandatory"? Does this add the required hint?

EmmanuelMess (changes requested):

app/src/main/java/com/amaze/filemanager/utils/Utils.java
tnyeanderson: No it doesn't, that code is [here](https://github.com/TeamAmaze/AmazeFileManager/blob/028600a08e677f0b418b010b6596187bf62bdd59/app/src/main/java/com/amaze/filemanager/ui/dialogs/SmbConnectDialog.java#L156-L191) and [here](https://github.com/TeamAmaze/AmazeFileManager/blob/028600a08e677f0b418b010b6596187bf62bdd59/app/src/main/java/com/amaze/filemanager/ui/dialogs/SmbConnectDialog.java#L287-L306)

TranceLove (approved):
Tested OK on Pixel 3 emulator running stock Android 10 for visible shares, visible shares with explicit path and hidden shares.

+1.

app/src/main/java/com/amaze/filemanager/utils/Utils.java
EmmanuelMess: As I said called "TextInputLayoutExtensions".

VishalNehra: Why? Do we plan to have more extensions for TextInputLayout? Just extensions for UI components should be enough. Shouldn't segregate more than is required.

EmmanuelMess: Extensions are quite common in Kotlin, and probably more will be needed. But I do see the point that right now it is the only one.

EmmanuelMess:
Should this PR save the path if inserted in the IP address? Saving an SMB connection with ip "10.0.2.2/sambashare" and then editing shows only the ip, but if you open it, it opens inside the "sambashare" folder.

EmmanuelMess:
Also, I cannot find the new "connection name" field. Where should it show?

tnyeanderson:
Connection name is the first field in the samba connection setup/edit window.

As for the "saving a path", it should be attached to the "Share path", not the "Server IP Address". It will get saved and open correctly, and it show up in the edit mode as well.

EmmanuelMess:
Ok, it seems something had failed last time I tested. There is an issue where username and password are not actually required, you can still save the connection without those.

VishalNehra:
This should be good to go @EmmanuelMess @TranceLove please test and confirm.

EmmanuelMess:
> This should be good to go @EmmanuelMess @TranceLove please test and confirm.

There is an issue where username and password are not actually required, you can still save the connection without those.

VishalNehra:
Yes, just checked. It's present in prod version as well. Not created because of these changes. Maybe raise a new issue for that? 🤔

EmmanuelMess:
> Yes, just checked. It's present in prod version as well. Not created because of these changes. Maybe raise a new issue for that? thinking

Was it always supposed to be required?

VishalNehra:
Not sure what caused it. It's supposed to be required if anonymous is not checked.

EmmanuelMess (approved):

TranceLove:
> > This should be good to go @EmmanuelMess @TranceLove please test and confirm.
> 
> There is an issue where username and password are not actually required, you can still save the connection without those.

This is __not__ as issue. SMB does allow anonymous login if share is made public.

EmmanuelMess:
> > > This should be good to go @EmmanuelMess @TranceLove please test and confirm.
> > 
> > 
> > There is an issue where username and password are not actually required, you can still save the connection without those.
> 
> This is **not** an issue. SMB does allow anonymous login if share is made public.

If the anonymous checkbox is not selected, why would the ui allow for not entering username and password?

TranceLove:
@EmmanuelMess Yeah true. Will raise red flag after checking.

TranceLove:
> > > > This should be good to go @EmmanuelMess @TranceLove please test and confirm.
> > > 
> > > 
> > > There is an issue where username and password are not actually required, you can still save the connection without those.
> > 
> > 
> > This is **not** an issue. SMB does allow anonymous login if share is made public.
> 
> If the anonymous checkbox is not selected, why would the ui allow for not entering username and password?

See #2262.

TranceLove (approved):

PR #2212, opened 2021-01-11
Bump leakcanary to 2.6

Unpublished:
#### Build tasks success  

#### Additional Info

Automatic dumps can now be disabled from leak ui.
Dumps are also postponed til app is backgrounded.
If still you still prefer it off by default we can just comment out the dependency to disable leak canary.


EmmanuelMess (approved):
Works

TranceLove (approved):
Works on Pixel 3 emulator running stock Android 10. +1.

PR #2246, opened 2021-01-22
Workaround data directory access for Android >= 11 if device is rooted

### LINKED ISSUE #2015: Android 11, Can't access /storage/emulated/0/android/data

 - Device: Pixel 3a
 - OS: Android 11
 - Rooted: No
 - Version: 3.5.0

TranceLove:
#### Issue tracker   

Fixes #2015

#### Release  
Addresses release/3.6
  
#### Test cases
- [ ] Covered
  
#### Manual testing
- [x] Done  
  
Devices:
 - Pixel 2 emulator running stock Android 11, rooted with the help of the handy script from https://github.com/newbit1/rootAVD
 - Also tested for regressions:
   * Fairphone 3 running LineageOS 16.0 (9.0)
   * LG Nexus 5x running /e/OS 0.13 (8.1)
   * Oneplus 2 running /e/OS 0.13 (10.0)
   * GPD XD running LegacyROM (4.4.4)

#### Build tasks success  

#### Additional Info

If user is on Android 11 or above, and device is rooted, when accessing Android data directory which would be blocked access, augment the path to /data/media/0/Android/data and list files using root.

Additionally, added filter to ls results for some devices (including AVD emulator) where toybox is used instead of busybox, to filter out entries "." and "..".

Thanks again to @danielzgtg for the workaround idea!

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/filesystem/FileUtil.java
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/FileUtil.java
@@ -20,6 +20,7 @@
 
 package com.amaze.filemanager.filesystem;
 
+import static android.os.Build.VERSION_CODES.Q;
 import static com.amaze.filemanager.filesystem.smb.CifsContexts.SMB_URI_PREFIX;
 import static com.amaze.filemanager.filesystem.ssh.SshConnectionPool.SSH_URI_PREFIX;
 
@@ -37,6 +38,7 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.nio.channels.FileChannel;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.regex.Pattern;
 
@@ -64,6 +66,7 @@ import android.content.Context;
 import android.database.Cursor;
 import android.net.Uri;
 import android.os.Build;
+import android.os.Environment;
 import android.os.ParcelFileDescriptor;
 import android.preference.PreferenceManager;
 import android.provider.BaseColumns;
@@ -91,13 +94,18 @@ public abstract class FileUtil {
   private static final Pattern FILENAME_REGEX =
       Pattern.compile("[\\\\\\/:\\*\\?\"<>\\|\\x01-\\x1F\\x7F]", Pattern.CASE_INSENSITIVE);
 
-  /**
-   * Determine the camera folder. There seems to be no Android API to work for real devices, so this
-   * is a best guess.
-   *
-   * @return the default camera folder.
-   */
-  // TODO the function?
+  private static final String EXTERNAL_STORAGE_PREFIX;
+
+  private static final List<String> EXCLUDED_DIRS =
+      Arrays.asList(
+          new File(Environment.getExternalStorageDirectory(), "Android/data").getAbsolutePath(),
+          new File(Environment.getExternalStorageDirectory(), "Android/obb").getAbsolutePath());
+
+  static {
+    String prefix = Environment.getExternalStorageDirectory().getAbsolutePath();
+    prefix = prefix.substring(0, prefix.lastIndexOf('/'));
+    EXTERNAL_STORAGE_PREFIX = prefix;
+  }
 
   /**
    * Copy a file. The target file may even be on external SD card for Kitkat.
@@ -1064,6 +1072,14 @@ public abstract class FileUtil {
     } catch (IOException e) {
       Log.e(LOG, "Could not copy dummy files.", e);
       return null;
+    }
+  }
+
+  public static @NonNull String remapPathForApi30OrAbove(@NonNull String path, boolean isRoot) {
+    if (Build.VERSION.SDK_INT > Q && isRoot && EXCLUDED_DIRS.contains(path)) {
+      return "/data/media" + path.substring(EXTERNAL_STORAGE_PREFIX.length());
+    } else {
+      return path;
     }
   }
 
--- a/app/src/main/java/com/amaze/filemanager/filesystem/root/ListFilesCommand.kt
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/root/ListFilesCommand.kt
@@ -198,7 +198,7 @@ object ListFilesCommand : IRootCommand() {
         path: String,
         isStat: Boolean
     ): HybridFileParcelable? {
-        return FileUtils.parseName(
+        FileUtils.parseName(
             if (isStat) rawFile.replace(
                 "('|`)".toRegex(),
                 ""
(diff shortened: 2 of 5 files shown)


app/src/main/java/com/amaze/filemanager/filesystem/root/ListFilesCommand.kt
```diff
@@ -50,24 +53,39 @@ object ListFilesCommand : IRootCommand() {
...
+            _path.equals(
+                    File(
+                            Environment.getExternalStorageDirectory(),
+                            "Android/data"
+                        ).absolutePath
+                ) &&
+            Build.VERSION.SDK_INT > Q
+        ) {
+            // Using root hack for Android >= 11 to access Android data directory
+            _path = "/data/media/0/Android/data"
```
danielzgtg, line 67: It would be nice if we support multiple users by passing the actual user ID

app/src/main/java/com/amaze/filemanager/filesystem/root/ListFilesCommand.kt
```diff
@@ -50,24 +53,39 @@ object ListFilesCommand : IRootCommand() {
         onFileFoundCallback: (file: HybridFileParcelable) -> Unit
     ) {
         val mode: OpenMode
-        if (root && !path.startsWith("/storage") && !path.startsWith("/sdcard")) {
+        var _path = path
+        if (root &&
+            _path.equals(
+                    File(
+                            Environment.getExternalStorageDirectory(),
+                            "Android/data"
```
danielzgtg, line 61: `Android/obb` is also restricted, and could similarly benefit from this

app/src/main/java/com/amaze/filemanager/filesystem/root/ListFilesCommand.kt
danielzgtg: Come to think of it, using `/data/media` isn't even needed if we have root, so we don't need to rewrite the path.

The reason my workaround was needed is because I needed to avoid `/sdcard` in order to trick Amaze into using the root codepath. Also, I haven't tested whether my workaround works on devices with actual SD cards

TranceLove: Seems SD card won't be visible via `/data/media` anyway.

TranceLove:
The latest force push should now be multi-user friendly, where user ID is no longer assumed to be 0 (owner). Not touching it at all anyway - just replace the storage directory prefix with `/data/media`.

Additionally,
 - check for Android > 10 before other conditions
 - remaps path only when necessary

app/src/main/java/com/amaze/filemanager/filesystem/root/ListFilesCommand.kt
```diff
@@ -138,6 +160,18 @@ object ListFilesCommand : IRootCommand() {
...
 
+    /**
+     * Remaps given path to root-accessible paths. Returns null if no need to remap.
+     */
+    private fun remap(path: String): String? {
+        return if (!EXCLUDED_DIRS.contains(path)) {
+            null
+        } else {
+            val prefix = Environment.getExternalStorageDirectory().absolutePath
+            path.replace(prefix.substring(0, prefix.lastIndexOf('/')), "/data/media")
```
danielzgtg, line 171: I think `String.replace` is inefficient and will fail in the contrived example of `/storage/emulated/0/Android/data/com.example.badapp/storage/emulated/0/myfile.txt`.

Instead we can just replace the first occurrence, or better, assume that the first occurrence is at the start because it passed the EXCLUDED_DIRS conditional. The new expression would look something like `"/data/media" + path.substring(prefix.lastIndexOf('/'))`

danielzgtg:
I have tested this on a OnePlus 7 Pro running the 1st Android 11 Beta:

- The file list in `/sdcard/Android/{data,obb}` works now
- I was able to open an app's text file somewhere inside `data` and read the contents
- I've noticed going to `/sdcard/Android/data` in two taps from `/sdcard/` has `/sdcard/Android/data` in the title, but only after navigating to a folder inside does the title become something starting with `/data/media/0`, and that navigating back goes to `/data/media/0/Android` instead of `/sdcard/Android`. However, it might not be worth fixing this
- Overall, it's quite nice how the transition is seamless and, to a casual observer, everything looks exactly the same as in Android 10!

TranceLove:
Let's use back the traditional Java approach then - by putting the path comparison and rewrite at `MainFragment`. Will miss Kotlin's convenient shorthands, but shall give a better UX as suggested - path bar will also be rewritten, and will give a Toast that user is being redirected using root.

app/src/main/java/com/amaze/filemanager/filesystem/root/ListFilesCommand.kt
```diff
@@ -54,7 +54,9 @@ object ListFilesCommand : IRootCommand() {
             // we're rooted and we're trying to load file with superuser
             // we're at the root directories, superuser is required!
             val result = executeRootCommand(path, showHidden)
-            result.first.forEach {
+            result.first.filter {
+                !it.endsWith(".") && !it.endsWith("..")
```
danielzgtg, line 58: I can't see a file I created named `hi.`

app/src/main/java/com/amaze/filemanager/filesystem/FileUtil.java
```diff
@@ -91,13 +94,18 @@
...
-   */
-  // TODO the function?
+  private static final String EXTERNAL_STORAGE_PREFIX;
+
+  private static final List<String> EXCLUDED_DIRS =
+      Arrays.asList(
+          new File(Environment.getExternalStorageDirectory(), "Android/data").getAbsolutePath(),
+          new File(Environment.getExternalStorageDirectory(), "Android/obb").getAbsolutePath());
+
+  static {
```
VishalNehra, line 104: Can we use `DataUtils.getStorages` here? 🤔

VishalNehra (changes requested):

app/src/main/java/com/amaze/filemanager/filesystem/FileUtil.java
TranceLove: If the first path returned by `DataUtils.getStorages()` is guaranteed to be the path for the equivalent of `Environment.getExternalStorageDirectory()`, yes.

VishalNehra: I really think using static block isn't ideal. It'll slow down application load times. 
Please see if you can avoid it. `DataUtils.getStorages` does have first path equivalent to external storage most of the times, maybe can be tested on few devices.

TranceLove:
Closing, in favour of non root implementation.

PR #2254, opened 2021-01-26
Range selection

### LINKED ISSUE #2070: Range selection
Apparently this has not been suggested so far, albeit very important.

## Range selection

From [Wikipedia](https://en.wikipedia.org/wiki/ES_File_Explorer#Popularization_and_functionality):

> Another early feature [of ES File Explorer] to allow the selection of many files and folders without the need to tap on each file individually was range selection, where only two items need to be highlighted to select all listed items inbetween.

This was a **vital hallmark** of ES File Explorer (the grandfather of Android file explorers, which has sadly not aged so well), and I myself used that feature **a lot**.

When using Amaze for the first time, it was the first thing I noticed a lack of.

Therefore, I strongly encourage implementing that feature.

ArthurCueio:
## PR Info

I've added a "Select between" action item when 2 or more files are selected. The wording is not set in stone,   I'm open to suggestions

#### Issue tracker   

Fixes #2070

#### Release  
Addresses release/
  
#### Test cases
- [ ] Covered
  
#### Manual testing
- [x] Done  
  
If yes,  
- Device: Motorola Moto E4
- OS: Android 7.1.1

#### Build tasks success  

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
+++ b/app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
@@ -486,8 +486,6 @@ public class MainFragment extends Fragment implements BottomBarButtonPath {
             showOption(R.id.openmulti, menu);
           }
           // tv.setText(positions.size());
-          Log.d("THING", String.valueOf(results));
-          Log.d("THING", String.valueOf(positions.size()));
           if (!results) {
             hideOption(R.id.openparent, menu);
             if (positions.size() == 1) {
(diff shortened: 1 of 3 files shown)


EmmanuelMess:
No unnecessary logs please, and necessary ones should have the simple class name as tag.

EmmanuelMess:
This works by checking if two items are selected, then a button appears in the dropdown so that if the user presses it, all items between the two preselected items get selected? That seems very unintuitive.

EmmanuelMess:
It also doesn't allow for the selection of more than one range

ArthurCueio:
> No unnecessary logs please, and necessary ones should have the simple class name as tag.

Sorry, I was messing around with the code and didn't  see those made it in, removed them and squashed the commit.

> This works by checking if two items are selected, then a button appears in the dropdown so that if the user presses it, all items between the two preselected items get selected? That seems very unintuitive.

Yes, I haven't used a feature like this in other file managers before, this is the first way that came to mind. Definitely open for suggestions here.

EmmanuelMess:
Okay, it seems this is the way ES File Manager does range selection. I personally would prefer a system more like the range selection for Google Photos. @TranceLove @VishalNehra what do you think?

VishalNehra:
Agreed here, need something like Google photos. I think there's an API which supports drag and pass us all the ui elements present below it. We can make use of that api. Maybe just support whatever adapter elements are there at one screen for now, and support for list auto-scrolling if drag comes to bottom part of the screen can be added later.

VishalNehra:
No work being done here? Closing...

PR #2283, opened 2021-02-05
Converted files to Kotlin

EmmanuelMess:
Translated files to Kotlin


VishalNehra (dismissed):

TranceLove:
OK, I think both of us don't like the codacy problem report... so please `@Suppress` them.

Otherwise it's good to go.

VishalNehra:
> OK, I think both of us don't like the codacy problem report... so please `@Suppress` them.
> 
> Otherwise it's good to go.

Should be relatively easy to fix these codacy issues.

EmmanuelMess:
@TranceLove how is @Supress used?

TranceLove:
@EmmanuelMess Not much trick here... Kotlin's own @Suppress, then comma separated list of rules that shall be skipped. Can see the tests I wrote for archive extractors😉

But for line width, that's another story... as code comment bracing the lines to skip line width.

https://github.com/pinterest/ktlint#how-do-i-suppress-an-error-for-a-lineblockfile

VishalNehra (approved):

TranceLove (approved):

PR #2399, opened 2021-04-01
preserve last modified date of copied files

### LINKED ISSUE #2375: Preserve file modification time at copy
Now after coping files through the filemanager, modification date changes. Which screws photo sorting in Gallery and potentially other applications.

Would be nice to preserve modification date after coping

dsvi:
Fixes #2375

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/filesystem/files/GenericCopyUtil.java
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/files/GenericCopyUtil.java
@@ -94,8 +94,6 @@ public class GenericCopyUtil {
    */
   private void startCopy(boolean lowOnMemory) throws IOException {
 
-    FileInputStream inputStream = null;
-    FileOutputStream outputStream = null;
     FileChannel inChannel = null;
     FileChannel outChannel = null;
     BufferedInputStream bufferedInputStream = null;
@@ -316,9 +314,15 @@ public class GenericCopyUtil {
         else if (outChannel != null) copyFile(inChannel, outChannel);
       }
       // till now file might not be written to disk yet, but to an OS buffer.
-      if (outputStream != null) outputStream.close();
-      if (outChannel != null) outChannel.close();
-      if (bufferedOutputStream != null) bufferedOutputStream.close();
+      if (bufferedOutputStream != null) {
+        bufferedOutputStream.close();
+        bufferedOutputStream = null;
+      }
+      if (outChannel != null) {
+        outChannel.close();
+        outChannel = null;
+      }
+
       if (!mTargetFile.setLastModified(mSourceFile.getDate()))
         Log.e(getClass().getSimpleName(), "Can't set last modified time for " + mTargetFile.getPath());
     } catch (IOException e) {
@@ -336,8 +340,9 @@ public class GenericCopyUtil {
 
       try {
         if (inChannel != null) inChannel.close();
-        if (inputStream != null) inputStream.close();
+        if (outChannel != null) outChannel.close();
         if (bufferedInputStream != null) bufferedInputStream.close();
+        if (bufferedOutputStream != null) bufferedOutputStream.close();
       } catch (IOException e) {
         e.printStackTrace();
         // failure in closing stream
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/amaze/filemanager/filesystem/files/GenericCopyUtil.java
```diff
@@ -315,11 +315,16 @@ else if (outChannel != null) {
         if (bufferedOutputStream != null) copyFile(inChannel, bufferedOutputStream);
         else if (outChannel != null) copyFile(inChannel, outChannel);
       }
-
+      // till now file might not be written to disk yet, but to an OS buffer.
+      if (outputStream != null) outputStream.close();
```
EmmanuelMess, line 319: This might cause the stream not to be closed if the code before throws an exception.

EmmanuelMess (changes requested):

app/src/main/java/com/amaze/filemanager/filesystem/files/GenericCopyUtil.java
dsvi: It will be closed anyways when GC collects the object. Exceptions are exceptional ;) So even though you may repeat it in finally block, i don't believe it really worth the complication. Especially since it doesn't cause much trouble even in the exceptional case.
The way it was, you were simply masked errors and reported that copy went fine, even when it wasn't. When you just do write, it doesn't mean that file was actually written. It means you have filled a buffer. Which only truly is flushed to disk at close (and not always even then). So only after checking the close, you can be confident that it actually reached the disk fine.

EmmanuelMess: This might silently leak memory for some implemetation of copy.

EmmanuelMess: @TranceLove opinion?

TranceLove: If you need the outputStream to close unconditionally (if not null of course) you may want to do this in a `finally`.

EmmanuelMess: @dsvi Please move the code in such a way that it always executes.

app/src/main/java/com/amaze/filemanager/filesystem/files/GenericCopyUtil.java
```diff
@@ -315,11 +313,22 @@ else if (outChannel != null) {
         if (bufferedOutputStream != null) copyFile(inChannel, bufferedOutputStream);
         else if (outChannel != null) copyFile(inChannel, outChannel);
       }
+      // till now file might not be written to disk yet, but to an OS buffer.
+      if (bufferedOutputStream != null) {
```
EmmanuelMess, line 317: You might never get here if an exception ocurrs.

EmmanuelMess (changes requested):

app/src/main/java/com/amaze/filemanager/filesystem/files/GenericCopyUtil.java
dsvi: I know

EmmanuelMess: This will probably not be accepted unless you fix this.

dsvi: fix what? nothing is broken there

EmmanuelMess: if the code crashes stream is not closed.

dsvi: The stream is closed no matter exception or not

TranceLove:
Should be good to go.

Please rebase and fix conflict so that the lines before your change

https://github.com/TeamAmaze/AmazeFileManager/blob/9747a5814825da4d22a5c7437ed542eaa252ab08/app/src/main/java/com/amaze/filemanager/filesystem/files/GenericCopyUtil.java#L326-L327

will read as

```java
      Objects.requireNonNull(inChannel);
      Objects.requireNonNull(outChannel);

      doCopy(inChannel, outChannel, updatePosition);
```

Deleora:
Has this feature been removed from the App?

EmmanuelMess:
> Has this feature been removed from the App?

I don't think this was ever in the app.

Deleora:
There was some work going on as I could see the pull request and everything and then the last conversation says "deleted the feature/2375" so I thought it was implemented and then removed.
Anyways, even I find this feature really good especially from Gallery perspective. So what's the status of this feature as of now?

PR #2413, opened 2021-04-08
Use UNIX timestamp when load file list using `stat` command

### LINKED ISSUE #2263: stat shall use UNIX timestamp instead of rely on *box time formatting
For those devices which still rely on https://github.com/landley/toybox instead of https://github.com/osm0sis/android-busybox-ndk.

Comparing [stat for 0.8.3](https://github.com/landley/toybox/blob/01864337c4190d19f21ee8df0beaf51f783c1940/toys/other/stat.c#L69-#L72) and [stat for 0.7.6](https://github.com/landley/toybox/blob/c0ad4545137107503d64c7ed1d9bc685ec2f8e9c/toys/other/stat.c#L68-L75), time formatting had been changed even with the same `%y` flag.

Shall use `%Y` instead, which uses UNIX time which should be stable across different versions, and even compatible with busybox.

TranceLove:
#### Issue tracker   

Fixes #2263

#### Release  
Addresses release/
  
#### Test cases
- [x] Covered
  
#### Manual testing
- [ ] Done  
  
If yes,  
- Device:
- OS:

#### Build tasks success  

#### Additional Info
With UNIX timestamp in place hope this can eventually eliminate the need of `ls`.

This changeset won't remove `ls` usage; it can be done with another PR once proved stable.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/filesystem/RootHelper.java
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/RootHelper.java
@@ -128,7 +128,7 @@ public class RootHelper {
       for (String currentLine : resultLines) {
         if (contains(currentLine.split(" "), name)) {
           try {
-            HybridFileParcelable parsedFile = FileUtils.parseName(currentLine);
+            HybridFileParcelable parsedFile = FileUtils.parseName(currentLine, true);
             if (parsedFile.getPermission().trim().startsWith("d")) return true;
             else if (parsedFile.getPermission().trim().startsWith("l")) {
               if (count > 5) return file.isDirectory();
--- a/app/src/main/java/com/amaze/filemanager/filesystem/files/FileUtils.java
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/files/FileUtils.java
@@ -783,7 +783,7 @@ public class FileUtils {
    *
    * @param line must be the line returned from 'ls' or 'stat' command
    */
-  public static HybridFileParcelable parseName(String line) {
+  public static HybridFileParcelable parseName(String line, boolean isStat) {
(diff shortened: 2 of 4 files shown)


VishalNehra (dismissed):

app/src/main/java/com/amaze/filemanager/filesystem/RootHelper.java
```diff
@@ -128,7 +128,7 @@ public static boolean isDirectory(String toTest, int count) throws ShellNotRunni
       for (String currentLine : resultLines) {
         if (contains(currentLine.split(" "), name)) {
           try {
-            HybridFileParcelable parsedFile = FileUtils.parseName(currentLine);
+            HybridFileParcelable parsedFile = FileUtils.parseName(currentLine, true);
```
EmmanuelMess, line 131: isStat is never false.

EmmanuelMess (changes requested):

app/src/main/java/com/amaze/filemanager/filesystem/RootHelper.java
TranceLove: Not always. There are two places that would cause `isStat` to become `false`:

https://github.com/TranceLove/AmazeFileManager/blob/feature/2263/app/src/main/java/com/amaze/filemanager/filesystem/root/ListFilesCommand.kt#L56-L64

https://github.com/TranceLove/AmazeFileManager/blob/feature/2263/app/src/main/java/com/amaze/filemanager/filesystem/RootHelper.java#L125-L127

EmmanuelMess (approved):

VishalNehra (approved):

PR #2427, opened 2021-04-20
Features - Drag and drop; range selection

### LINKED ISSUE #2070: Range selection

### LINKED ISSUE #2215: Drag & Drop files in multi-window
Drag & Drop files in multi-window

Raised by: vishal_nehra
Reporter: rramnel

VishalNehra:
#### Issue tracker   

Fixes #2215 
Fixes #2070 
-or-   
Addresses #

#### Release  
Addresses release/3.6
  
#### Test cases
- [ ] Covered
  
#### Manual testing
- [x] Done  
  
If yes,  
- Device:
- OS:

#### Build tasks success  

#### Related PR  
Related to PR #

#### Additional Info


VishalNehra:
Few things to test (not limited to this):
1. User can select, drag and drop on adjacent pane.
2. User can drag and drop on empty no files screen / empty space below list / on files / on directory in adjacent pane.
3. User can drag and drop on same pane, just in directory
4. User can select and drag to top / bottom part to scroll big list
5. All these things can be done with listview and gridview
6. All these actions can be performed on back item on top, will move /. copy the items to parent directory.
7. Can't move between adjacent pane in drag and select mode, adjacent pane drag is just for copy / move feature.

EmmanuelMess (changes requested):
Erratic range selection in AVD. It seems that it is not a "range" just "the ones that are under your finger when the app detects it"

EmmanuelMess:
A lot of glichyness

VishalNehra:
> A lot of glichyness, when selecting and moving the finger up or down.

That was expected as we're also showing a checkbox in front of icon, when it's animating, if it's disabled again that might be the cause of problem. Maybe We can disable this checkbox and just highlight the item or keep this as it is.

EmmanuelMess:
It seems glitchy, if I press until it is selected, and then move my finger down, it doesnt select a range.
Long pressing on the icon doesn't even select the item.
Selecting through the "Files" header runs the risk of accelerating the selection too fast past the first file.
Long pressing a selected item doesn't seem to do anything.
If you keep your finger down on the same item it stops the range selection after a second or so.

EmmanuelMess:
It is my opinion that all that cannot be selected when dragging should be removed from view to unclutter the space.
Also, both drag and range select should be usable at the same time. By making the drag feature by long pressing on the icon, and the select feature by long pressing on the body of the item.

EmmanuelMess:
> > 
> > Fixes:
> > 
> > * Quite a few of these would be fixed by having a dialog after the user drags.
>   
> Really bad UX IMO. It makes the operation slow and annoying.  The most that can be done is show a dialog to ask whether to move or copy, and provide an option to remember the operation. That'll solve both the problems. Actually three, I can put a cancel button in dialog as well that'll provide another way to cancel the operation.

Okay.

EmmanuelMess:
> > * Remove the 3 dots while dragging, and add a place to put your finger to move down into the folder.
>   
> I don't understand what you mean by putting finger down. 

Sorry I meant when the dragging starts.


> > * The preference is too difficult to find.
> 
> Agreed, there's no other way to solve this. We can only support whether user wants to drag and select or drag + drop to copy / move. Both can't be done simultaneously. Most file managers support only either, not even both. So giving both options in itself should be more than enough.

I proposed "Adding more 'views' for different categories of settings." and "By making the drag feature by long pressing on the icon, and the select feature by long pressing on the body of the item." what do you think of those fixes?

> > 
> > Fixes:
> > 
> > * Quite a few of these would be fixed by having a dialog after the user drags.
>   
> Really bad UX IMO. It makes the operation slow and annoying.  The most that can be done is show a dialog to ask whether to move or copy, and provide an option to remember the operation. That'll solve both the problems. Actually three, I can put a cancel button in dialog as well that'll provide another way to cancel the operation.

Okay.

> > * Move the folder up, so that the user sees it when dragging. Remember to deal with the top of the device
>
> This will cause a lot of problems with deciding whether user is dropping on item above or below. The framework support callbacks when the shadow enters a view. So if I shift the view a bit above wrong callbacks will be sent.
> I think just increasing icon size will create similar problem. Leaving it for enhancement.

What do you mean by "the shadow"? If there is a shadow it should be always used to indicate where the grabbed item "falls", moving the entire thing "up" above the user's finger should work, shouldn't it?

> > * There is no way to drag "up" the directory, there is no way to drag down two directories.
>  
> If you long press and come to bottom part of screen it'll drag down the directory, similarly if you drag finger to app bar while holding it'll drag up the directory. You can also drag to left and right corners to switch between tabs and keep holding the items to move.

Sorry I meant the directory tree, if A/B/c is a structure with A, B directories and c a file, there is no way to grab and drop c into A.

> > * Make a section of the screen be "move to the parent folder", maybe the first folder?
>   
> Moving to parent folder can be done using dropping to go back button. Adding anything more will add to complication and won't be worth.

You mean the "back" button on the device or the "directory up" item? The back button has the issue of being unintuitive (back button should be used to cancel operations or return to a previous view), and the "directory up" is not always visible.
My fix would be to force the ".." item to be available when dragging.

VishalNehra:
@EmmanuelMess 
Added following changes as per discussion:
- On drag / drop we will now display a dialog to user. He can cancel the operation or choose from either copy or move operation
- He can choose to remember a given operation from this dialog, so from next time it won't bother him
- The drag icon is enlarged now, the icon will have  current operation which will be performed (in case user remembered either of copy or move)
- Drag icon is moved a bit upward so this info is clearly visible.

TranceLove:
Hmm, entirely different UX than I had back in the days of ES many years ago... but whatever. They even have a selection rectangle while dragging over the screen - doesn't mean you need to follow the same.

Tried on Fairphone 3 running LineageOS 16.0 (9.0)

- sometimes drag to do range select may not work after a few cancels. Long press to select became unresponsive 
- how am I supposed to do to trigger drag and drop file/folders? I thought I could long press the item then drag to the folder I want to, either on the same tab or the next tab

> Drag feature on long press of icon might not be a good idea.

I'm +1 about this behaviour however, since this is what we would normally do to move an app icon on the launcher/home screen. I usually have Lawnchair as daily driver however. iOS and Bliss Launcher (as used by /e/) also have the same behaviour for app icons.

VishalNehra:
> * how am I supposed to do to trigger drag and drop file/folders? I thought I could long press the item then drag to the folder I want to, either on the same tab or the next tab

You can change it from amaze settings.  I've added a preference.

TranceLove:
Behaviour tested OK on Fairphone 3 running LineageOS 16.0 (9.0) and GPD XD running LegacyROM (4.4.4), on both drag-n-drop to same or other tab.

The cut/copy icons on the dialog are not visible when using Light themes, but looks OK on dark/AMOLED themes though.

Screenshot_20210516-111257_Amaze_Debug
Screenshot_20210516-111754_Amaze_Debug

TranceLove:
As daily use I found no obvious problems now, functionally I'd vote +1, except the conflict (sorry, due to #2450).

TranceLove (approved):

EmmanuelMess:
How did this pass with a single approval?

VishalNehra:
@EmmanuelMess There's a condition, with one approval and a pr open more than 30 days will get merged :)

HT-7:
>  drag to do range select

Please don't tell me I have to hold my finger towards the bottom of the screen until the file list scrolls to the last file of the range I would like to select. This is the misimplementation I dreaded since starting #2070.

On ES File Explorer, I could select file A, drag the **scroll bar** (takes one second), select file B, and tap an icon to select all inbetween, which is superior compared to having to hold the finger and wait potentially minutes for the list to scroll to the last file I wish to select.

On desktop, it is highlight file A, and hold the "Shift" key while highlightsing file B, and all items in between are selected. More intuitive to users than dragging down and staring at a long list scrolling down. I hope this is understandable.

EmmanuelMess:
> On desktop, it is highlight file A, and hold the "Shift" key while highlightsing file B, and all items in between are selected

This comparison doesn't work, we don't have a keyboard, and how many users use that feature as opposed to click and drag to select (with the axis aligned box). 

> More intuitive to users than dragging down and staring at a long list scrolling down. 

Why?

EmmanuelMess:
Please start an issue.

Issue #2318, opened 2021-02-19
Sort by size doesn't work on folders


katezach:
**Describe the bug**
When I open the app and my folders(not files) are on display, when I choose from the 3 dots (up and right corner)  sort ->sort by->size and I choose either ascending or descending, the result doesn't change.
**To Reproduce**
1. Choose Sort
2. Click on Sort by Size
3. Choose Ascending, see the results
4. Choose Descending, see the results
5. Compare.

**Expected behavior**
The folders should be sorted by size.

**Screenshots**
https://www.dropbox.com/s/nokwfz5hivyjiye/videorecord.mp4?dl=0

**Smartphone (please complete the following information):**
 - Device: Lenovo Z6 Pro
 - OS: Android 9
 - Rooted: [e.g. No]
 - Version: 11.0.405 ST

**Additional context**

PR #2455, opened 2021-05-17
issue #2318 fixed.

khashiii97:
now able to sort folders as well.

1. In ListFilesCommand.kt instead of passing size zero when the variable currentFile is a directory, we obtain the size of the folder with a simple method implemented in Utils.java and set  it as size.
2. In LoadFilesListTask, in createListParcelables method we will obtain the size of the basefile when it's a directory.
3. In FileListSorter.java when comparing based on size, if both files are a directory then they'll be compared based on their real sizes.


VishalNehra:
Please join our telegram channel for more discussion
https://t.me/AmazeFileManager

VishalNehra:
Few issues here
1. The build is failing. Please fix formatting using `./gradlew spotlessapply`
2. We never really decided to implement this as this will significantly slow down list loading time. So we need to see if we can improve the approach.
3. The implementation currently focuses on using `File` apis. This will not work on say cloud storages or any other type of filesystem.

Issue #2460, opened 2021-05-18
Crash on date parse for file listing


VishalNehra:
## Exception
* User Action: UI Error
* Request: Application crash
* Version: 3.5.3
* OS: Linux Android 10 - 29
* Device: OnePlus7TPro
* Model: HD1913
* Product: OnePlus7TPro
<details><summary><b>Crash log </b></summary><p>

java.lang.RuntimeException: An error occurred while executing doInBackground()
 at android.os.AsyncTask$4.done(AsyncTask.java:399)
 at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:383)
 at java.util.concurrent.FutureTask.setException(FutureTask.java:252)
 at java.util.concurrent.FutureTask.run(FutureTask.java:271)
 at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
 at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
 at java.lang.Thread.run(Thread.java:919)
Caused by: java.lang.NumberFormatException: For input string: "May"
 at java.lang.Long.parseLong(Long.java:594)
 at java.lang.Long.parseLong(Long.java:636)
 at com.amaze.filemanager.filesystem.files.FileUtils.parseName(FileUtils.java:756)
 at com.amaze.filemanager.filesystem.root.ListFilesCommand.parseStringForHybridFile(ListFilesCommand.kt:201)
 at com.amaze.filemanager.filesystem.root.ListFilesCommand.listFiles(ListFilesCommand.kt:59)
 at com.amaze.filemanager.asynchronous.asynctasks.LoadFilesListTask.doInBackground(LoadFilesListTask.java:223)
 at com.amaze.filemanager.asynchronous.asynctasks.LoadFilesListTask.doInBackground(LoadFilesListTask.java:67)
 at android.os.AsyncTask$3.call(AsyncTask.java:378)
 at java.util.concurrent.FutureTask.run(FutureTask.java:266)
 ... 3 more

</details>
<hr>

VishalNehra:
Dupe #2635

PR #2491, opened 2021-05-26
Filter ls entries ending with . and ..

TranceLove:
From deprecated PR while working on #2015. Quoting log message:

> Added filter to ls results for some devices (including AVD emulator) where `toybox` is used instead of `busybox`, to filter out entries "." and "..".

#### Issue tracker   

Fixes #
-or-   
Addresses #

#### Release  
Addresses release/
  
#### Test cases
- [ ] Covered
  
#### Manual testing
- [ ] Done  
  
If yes,  
- Device:
- OS:

#### Build tasks success  

#### Related PR  
Related to PR #

#### Additional Info


VishalNehra:
Can create an issue for this pr and mark that issue to 3.6?

TranceLove:
Closing since I couldn't reproduce the problem. `ls -la` did showed `.` and `..` on both Toybox and Busybox, but never affected file listing when using root explorer, with or without legacy listing option (which legacy listing is currently broken, possibly as #2460).

PR #2508, opened 2021-05-31
Suppress jetifier transform jar output during Github build

### LINKED ISSUE #2506: Runs have logs that are too long
Any run has logs that easily have more than 20000 lines for Run test cases. That is unusable.

TranceLove:
Fixes #2506.

Reference: https://stackoverflow.com/a/52353446


EmmanuelMess:
Probably good idea to remove "history is available." and "Build cache is disabled" too.

EmmanuelMess (approved):

PR #2535, opened 2021-06-04
chore: fix build deprecated waring and clean code

Bambooin:
#### Automatic tests

- [ ] Added test cases
  
#### Manual tests
- [x] Done  
  

#### Build tasks success  


EmmanuelMess:
That was not the issue, the issue was that we should have two different translatable strings one uppercase and one lowecase, and use the one that corresponds where they correspond.

Issue #2142, opened 2020-12-14
Crash on rename file in SD Card


VishalNehra:
## Exception
* User Action: UI Error
* Request: Application crash
* Version: 3.5.2
* OS: Linux Android 9 - 28
* Device: willow
* Model: Redmi Note 8T
* Product: willow_eea
<details><summary><b>Crash log </b></summary><p>

java.lang.NullPointerException: Attempt to invoke virtual method 'androidx.documentfile.provider.DocumentFile androidx.documentfile.provider.DocumentFile.findFile(java.lang.String)' on a null object reference
 at com.amaze.filemanager.filesystem.FileUtil.getDocumentFile(FileUtil.java:949)
 at com.amaze.filemanager.filesystem.files.FileUtils.scanFile(FileUtils.java:243)
 at com.amaze.filemanager.filesystem.files.FileUtils.lambda$scanFile$1(FileUtils.java:225)
 at com.amaze.filemanager.filesystem.files.-$$Lambda$FileUtils$nQFhhNVP8RswRigIhHzl4wuQz60.run(Unknown Source:4)
 at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:245)
 at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
 at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
 at java.lang.Thread.run(Thread.java:764)

</details>

PR #2539, opened 2021-06-05
Fix ExternalSdCardOperation.getDocumentFile()

TranceLove:
It was found that the arguments passed to String.split() in splitting given path was incorrectly interpreted, causing some SD card related operations to fail, especially when there is folder prefix(es).

Original code in Java should be calling
https://developer.android.com/reference/java/lang/String#split(java.lang.String)

while it was then calling this instead, when code was transformed into Kotlin, not recognizing the argument as regex but normal string
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/split.html

For example, for a file on the SD card `"tmp/DummyFile"`, it was interpreted as `["tmp/DummyFile"]` but not `["tmp", "DummyFile"]`, i.e. not splitted correctly. After this change paths should be splitted correctly.

I read that the regex in original code should be just splitting with `/`... so be it. Only a single `/` would be enough.

#### Issue tracker   
Addresses #2142 (I don't know if it would fix the issue, so just _addresses_ here)

#### Manual tests
- [x] Done  
  
- Device: Fairphone 3
- OS: LineageOS 16.0 (9.0)

on copying files to a subfolder on the same external SD card. Before the fix when the paste happens, SAF dialog will popup even when I had already grant SAF permission to Amaze; afterwards the paste operation would fail.

After applying the changeset this problem will disappear.

#### Build tasks success  

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/filesystem/ExternalSdCardOperation.kt
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/ExternalSdCardOperation.kt
@@ -83,7 +83,7 @@ object ExternalSdCardOperation {
             return document
         }
 
-        val parts = relativePath.split("\\/").toTypedArray()
+        val parts = relativePath.split("/").toTypedArray()
         for (i in parts.indices) {
             if (document == null) {
                 return null


TranceLove:
build test apk

VishalNehra (approved):

PR #2551, opened 2021-06-05
2149 Implement viewmodel for MainFragment

### LINKED ISSUE #1797: Searching more than once will preserve previous search result
Bug description
Search result should only show current search result rather than show current and previous result at the same time.

Step to reproduce
1. Input search text
2. Click "search" button
3. Search same text or different text again
4. Search result preserve previous result

Screenshot
Screenshot

Smartphone
360 N6 Pro

### LINKED ISSUE #2149: Id nonexistent on MainFragment
## Exception
* __User Action:__ UI Error
* __Version:__ 3.5.2
* __OS:__ Linux Nokia/Plate2_00WW/PL2_sprout:10/QKQ1.190828.002/00WW_4_16B:user/release-keys 10 - 29
* __Device:__ PL2_sprout
* __Model:__ Nokia 6.1
* __Product:__ Plate2_00WW
Crash log

```
java.lang.IllegalArgumentException: No view found for id 0x7f0901b9 (com.amaze.filemanager:id/pager) for fragment MainFragment{7e4d7b (9857c5b7-4bd9-4211-976f-de8b30d15daf) id=0x7f0901b9}
at androidx.fragment.app.FragmentManagerImpl.moveToState(FragmentManagerImpl.java:875)
at androidx.fragment.app.FragmentManagerImpl.moveFragmentToExpectedState(FragmentManagerImpl.java:1238)
at androidx.fragment.app.FragmentManagerImpl.moveToState(FragmentManagerImpl.java:1303)
at androidx.fragment.app.FragmentManagerImpl.dispatchStateChange(FragmentManagerImpl.java:2659)
at androidx.fragment.app.FragmentManagerImpl.dispatchActivityCreated(FragmentManagerImpl.java:2613)
at androidx.fragment.app.FragmentController.dispatchActivityCreated(FragmentController.java:246)
at androidx.fragment.app.FragmentActivity.onStart(FragmentActivity.java:542)
at androidx.appcompat.app.AppCompatActivity.onStart(AppCompatActivity.java:201)
at android.app.Instrumentation.callActivityOnStart(Instrumentation.java:1426)
at android.app.Activity.performStart(Activity.java:7825)
at android.app.ActivityThread.handleStartActivity(ActivityThread.java:3295)
at android.app.servertransaction.TransactionExecutor.performLifecycleSequence(TransactionExecutor.java:221)
at android.app.servertransaction.TransactionExecutor.cycleToPath(TransactionExecutor.java:201)
at android.app.servertransaction.TransactionExecutor.executeLifecycleState(TransactionExecutor.java:173)
at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:97)
at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2017)
at android.os.Handler.dispatchMessage(Handler.java:107)
at android.os.Looper.loop(Looper.java:214)
at android.app.ActivityThread.main(ActivityThread.java:7397)
at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:935)

```

### LINKED ISSUE #2525: Exit on AppsListFragment should exit app
Exit on AppsListFragment should exit app

VishalNehra:
#### Issue tracker   

Fixes #2149 
Fixes #2525 
Fixes #1797 

#### Automatic tests

- [ ] Added test cases
  
#### Manual tests
- [x] Done  
  

#### Build tasks success  

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/ui/fragments/AppsListFragment.java
+++ b/app/src/main/java/com/amaze/filemanager/ui/fragments/AppsListFragment.java
@@ -127,9 +127,6 @@ public class AppsListFragment extends ListFragment
       case R.id.sort:
         showSortDialog(((MainActivity) requireActivity()).getAppTheme());
         return true;
-      case R.id.exit:
-        ((MainActivity) requireActivity()).goToMain(null);
-        return true;
       default:
         return super.onOptionsItemSelected(item);
     }
--- a/app/src/main/java/com/amaze/filemanager/ui/fragments/AppsListFragment.java
+++ b/app/src/main/java/com/amaze/filemanager/ui/fragments/AppsListFragment.java
@@ -127,6 +127,9 @@ public class AppsListFragment extends ListFragment
       case R.id.sort:
         showSortDialog(((MainActivity) requireActivity()).getAppTheme());
         return true;
+      case R.id.exit:
+        requireActivity().finish();
+        return true;
       default:
         return super.onOptionsItemSelected(item);
     }
--- a/app/src/main/java/com/amaze/filemanager/ui/fragments/AppsListFragment.java
+++ b/app/src/main/java/com/amaze/filemanager/ui/fragments/AppsListFragment.java
@@ -136,7 +136,9 @@ public class AppsListFragment extends ListFragment
   public void onSaveInstanceState(@NonNull Bundle b) {
     super.onSaveInstanceState(b);
 
-    b.putParcelable(KEY_LIST_STATE, getListView().onSaveInstanceState());
+    if (this.isAdded()) {
+      b.putParcelable(KEY_LIST_STATE, getListView().onSaveInstanceState());
+    }
   }
 
   private void updateViews(MainActivity mainActivity, UtilitiesProvider utilsProvider) {
--- a/app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
+++ b/app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
@@ -1031,11 +1031,7 @@ public class MainFragment extends Fragment
       mainFragmentViewModel.setStopAnims(true);
 
       if (mainFragmentViewModel.getOpenMode() != OpenMode.CUSTOM) {
-        if (mainFragmentViewModel.getCurrentPath() != null) {
-          mainFragmentViewModel
-              .getDataUtils()
-              .addHistoryFile(mainFragmentViewModel.getCurrentPath());
-        }
+        mainFragmentViewModel.getDataUtils().addHistoryFile(mainFragmentViewModel.getCurrentPath());
       }
 
       listView.setAdapter(adapter);
@@ -1851,11 +1847,12 @@ public class MainFragment extends Fragment
           mainFragmentViewModel.setStopAnims(false);
         }
         return false;
+      default:
+        return true;
     }
-    return true;
   }
 
   public MainFragmentViewModel getMainFragmentViewModel() {
-    return this.mainFragmentViewModel;
+    return new ViewModelProvider(this).get(MainFragmentViewModel.class);
   }
 }
--- a/app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
+++ b/app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
@@ -341,8 +341,7 @@ public class MainFragment extends Fragment
 
   public void switchView() {
     boolean isPathLayoutGrid =
-        mainFragmentViewModel
-                .getDataUtils()
+        DataUtils.getInstance()
                 .getListOrGridForPath(mainFragmentViewModel.getCurrentPath(), DataUtils.LIST)
             == DataUtils.GRID;
     reloadListElements(false, mainFragmentViewModel.getResults(), isPathLayoutGrid);
@@ -862,7 +861,7 @@ public class MainFragment extends Fragment
                 FileUtils.openFile(new File(path), (MainActivity) getActivity(), sharedPref);
                 break;
             }
-            mainFragmentViewModel.getDataUtils().addHistoryFile(layoutElementParcelable.desc);
+            DataUtils.getInstance().addHistoryFile(layoutElementParcelable.desc);
           }
         }
       }
@@ -941,7 +940,7 @@ public class MainFragment extends Fragment
               mSwipeRefreshLayout.setRefreshing(false);
               if (data != null && data.second != null) {
                 boolean isPathLayoutGrid =
-                    mainFragmentViewModel.getDataUtils().getListOrGridForPath(path, DataUtils.LIST)
+                    DataUtils.getInstance().getListOrGridForPath(path, DataUtils.LIST)
                         == DataUtils.GRID;
                 setListElements(data.second, back, path, data.first, false, isPathLayoutGrid);
                 setListElements(data.second, back, path, data.first, false, isPathLayoutGrid);
@@ -1059,7 +1058,7 @@ public class MainFragment extends Fragment
       mainFragmentViewModel.setStopAnims(true);
 
       if (mainFragmentViewModel.getOpenMode() != OpenMode.CUSTOM) {
-        mainFragmentViewModel.getDataUtils().addHistoryFile(mainFragmentViewModel.getCurrentPath());
+        DataUtils.getInstance().addHistoryFile(mainFragmentViewModel.getCurrentPath());
       }
 
       listView.setAdapter(adapter);
@@ -1356,13 +1355,12 @@ public class MainFragment extends Fragment
                   int i;
                   AppConfig.toast(requireContext(), getString(R.string.unknown_error));
                   if ((i =
-                          mainFragmentViewModel
-                              .getDataUtils()
+                          DataUtils.getInstance()
                               .containsServer(mainFragmentViewModel.getSmbPath()))
                       != -1) {
                     getMainActivity()
                         .showSMBDialog(
-                            mainFragmentViewModel.getDataUtils().getServers().get(i)[0],
+                            DataUtils.getInstance().getServers().get(i)[0],
                             mainFragmentViewModel.getSmbPath(),
                             true);
                   }
@@ -1446,7 +1444,7 @@ public class MainFragment extends Fragment
       mainFragmentViewModel.getSearchHelper().clear();
     }
     for (SmbFile aMFile : mFile) {
-      if ((mainFragmentViewModel.getDataUtils().isFileHidden(aMFile.getPath()) || aMFile.isHidden())
+      if ((DataUtils.getInstance().isFileHidden(aMFile.getPath()) || aMFile.isHidden())
           && !showHiddenFiles) {
         continue;
       }
@@ -1510,7 +1508,7 @@ public class MainFragment extends Fragment
   private LayoutElementParcelable addTo(HybridFileParcelable mFile) {
     File f = new File(mFile.getPath());
     String size = "";
-    if (!mainFragmentViewModel.getDataUtils().isFileHidden(mFile.getPath())) {
+    if (!DataUtils.getInstance().isFileHidden(mFile.getPath())) {
       if (mFile.isDirectory()) {
         size = "";
         LayoutElementParcelable layoutElement =
@@ -1580,7 +1578,7 @@ public class MainFragment extends Fragment
   }
 
   public void hide(String path) {
-    mainFragmentViewModel.getDataUtils().addHistoryFile(path);
+    DataUtils.getInstance().addHistoryFile(path);
     File file = new File(path);
     if (file.isDirectory()) {
       File f1 = new File(path + "/" + ".nomedia");
--- a/app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
+++ b/app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
@@ -73,6 +73,7 @@ import com.amaze.filemanager.ui.views.DividerItemDecoration;
 import com.amaze.filemanager.ui.views.FastScroller;
 import com.amaze.filemanager.ui.views.WarnableTextInputValidator;
 import com.amaze.filemanager.utils.BottomBarButtonPath;
+import com.amaze.filemanager.utils.DataUtils;
 import com.amaze.filemanager.utils.MainActivityHelper;
 import com.amaze.filemanager.utils.OTGUtil;
 import com.amaze.filemanager.utils.Utils;
@@ -102,7 +103,6 @@ import android.view.LayoutInflater;
 import android.view.Menu;
 import android.view.MenuInflater;
 import android.view.MenuItem;
-import android.view.MotionEvent;
 import android.view.View;
 import android.view.ViewGroup;
 import android.view.ViewTreeObserver;
@@ -133,7 +133,7 @@ import jcifs.smb.SmbException;
 import jcifs.smb.SmbFile;
 
 public class MainFragment extends Fragment
-    implements BottomBarButtonPath, ViewTreeObserver.OnGlobalLayoutListener, View.OnTouchListener {
+    implements BottomBarButtonPath, ViewTreeObserver.OnGlobalLayoutListener {
 
   public ActionMode mActionMode;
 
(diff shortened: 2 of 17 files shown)


app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
```diff
@@ -1826,4 +1817,45 @@ public int getRootDrawable() {
...
+          stopAnimation();
+          mainFragmentViewModel.setStopAnims(false);
+        }
+        return false;
+    }
+    return true;
+  }
+
+  public MainFragmentViewModel getMainFragmentViewModel() {
+    return this.mainFragmentViewModel;
```
EmmanuelMess, line 1859: This is wrong, this should get an instance of the lifecycle independent holder of the model, and get an instance of the model from there.

EmmanuelMess (approved):

EmmanuelMess (changes requested):
The ViewModel should be lifecycle independent.

app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
```diff
@@ -1826,4 +1845,37 @@ public int getRootDrawable() {
...
+    } else {
+      mToolbarContainer.getViewTreeObserver().removeGlobalOnLayoutListener(this);
+    }
+  }
+
+  public @Nullable MainFragmentViewModel getMainFragmentViewModel() {
+    try {
+      return new ViewModelProvider(this).get(MainFragmentViewModel.class);
+    } catch (Exception e) {
+      Log.e(getClass().getSimpleName(), "Failed to create viewmodel", e);
```
EmmanuelMess, line 1876: Why the generic Exception catching?

app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
```diff
@@ -176,124 +157,106 @@
...
-   * boolean identifying if the search task should be re-run on back press after pressing on
-   * any of the search result
-   */
-  private boolean mRetainSearchTask = false;
-
-  /** For caching the back button */
-  private LayoutElementParcelable back = null;
-
-  private int dragAndDropPreference;
+  private MainFragmentViewModel mainFragmentViewModel;
```
EmmanuelMess, line 164: Why do you keep this? In some cases `mainFragmentViewModel` will be null when `getMainFragmentViewModel()` wouldn't be.

EmmanuelMess (changes requested):
We still need to check for leaks of context in the MainFragmentViewModel after these changes.

app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
EmmanuelMess: Please use a `Obejcts.requireNonNull()`

VishalNehra: Because it may throw fragment not initialised or any other fragment lifecycle exceptions in `new ViewModelProvider(this)`. Using `requireNonNull` won't help.

app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
VishalNehra: A null check is there for all public methods or `getMainFragmentViewModel()` This class level variable is defined for private methods so that I don't need to create it again and again.

VishalNehra:
No new code is introduced in this viewmodel. And `MainFragmentViewModel` 's lifecycle is managed as per the attached fragments. There are no new leaks in this.

app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
EmmanuelMess: The documentation doesn't say that [ViewModelProvider.get](https://developer.android.com/reference/androidx/lifecycle/ViewModelProvider#get(java.lang.Class%3CT%3E)) throws exceptions, did I miss something?

app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
EmmanuelMess: > I don't need to create it again and again.

Create what? The `ViewModelProvider`?

EmmanuelMess: Why don't you want to create it again and again?

VishalNehra: Yes, I don't need to declare in each method that's why it's at class level

app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
VishalNehra: It was throwing when i tested. I wasn't sure which ones that's why it's catching generic exception. 
Maybe the ones being thrown are runtime exception.

EmmanuelMess:
> No new code is introduced in this viewmodel. And MainFragmentViewModel 's lifecycle is managed as per the attached fragments. There are no new leaks in this.

The [documentation recommends that no Context instances are maintained in `ViewModel`](https://developer.android.com/topic/libraries/architecture/viewmodel) because that would cause a context to exist for longer than it should (across fragment destruction in this case).

app/src/main/java/com/amaze/filemanager/ui/fragments/data/MainFragmentViewModel.kt
```diff
@@ -0,0 +1,195 @@
...
+    lateinit var openMode: OpenMode
+
+    // defines the current visible tab, default either 0 or 1
+    // private int mCurrentTab;
+    /** For caching the back button  */
+    var back: LayoutElementParcelable? = null
+
+    var dragAndDropPreference = 0
+
+    val dataUtils: DataUtils = DataUtils.getInstance()
```
EmmanuelMess, line 63: Please don't keep an instance of a Singleton in the ViewModel.

VishalNehra: Instance isn't kept here.. this is just a reference to the original instance, right. Instance is maintained by dataUtils itself.

app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
EmmanuelMess: > Yes, I don't need to declare in each method that's why it's at class level

But you lose the guarantee that it won't be null. You'd have to check that for each usage.

app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
EmmanuelMess: > It was throwing when i tested. 

Detached fragment? This seems like another issue that's being shadowed here by this generic try catch.

app/src/main/java/com/amaze/filemanager/ui/fragments/AppsListFragment.java
```diff
@@ -127,9 +127,6 @@ public boolean onOptionsItemSelected(@NonNull MenuItem item) {
       case R.id.sort:
         showSortDialog(((MainActivity) requireActivity()).getAppTheme());
         return true;
-      case R.id.exit:
```
EmmanuelMess, line 130: This fixes another issue AFAIK. Please add fixes.

app/src/main/java/com/amaze/filemanager/ui/fragments/data/MainFragmentViewModel.kt
```diff
@@ -0,0 +1,195 @@
...
+import com.amaze.filemanager.ui.fragments.preference_fragments.PreferencesConstants
+import com.amaze.filemanager.utils.DataUtils
+import java.util.*
+
+class MainFragmentViewModel : ViewModel() {
+
+    var currentPath: String? = null
+
+    /** This is not an exact copy of the elements in the adapter  */
+    var listElements: ArrayList<LayoutElementParcelable>? = null
```
EmmanuelMess, line 39: This should never be null.

app/src/main/java/com/amaze/filemanager/ui/fragments/data/MainFragmentViewModel.kt
```diff
@@ -0,0 +1,195 @@
...
+    /**
+     * Initialize column number from preference
+     */
+    fun initColumns(sharedPreferences: SharedPreferences) {
+        if (columns == 0) {
+            sharedPreferences.getString(
+                PreferencesConstants
+                    .PREFERENCE_GRID_COLUMNS,
+                "-1"
+            )?.toInt()?.run {
```
EmmanuelMess, line 146: `sharedPreferences.getString` is never null AFAIK

app/src/main/java/com/amaze/filemanager/ui/fragments/data/MainFragmentViewModel.kt
EmmanuelMess: Sorry, I meant, don't keep a reference to a Singleton, it is already accessible via DataUtils.getInstance().

app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
EmmanuelMess: This look way to defensive, isn't there a better way to do this?

EmmanuelMess: The documentation asserts that the get() is nonnull, but here you are losing that guarantee, and then you have quite a lot of code rechecking to obtain again the guarantee of nonnullness.

app/src/main/java/com/amaze/filemanager/ui/fragments/data/MainFragmentViewModel.kt
VishalNehra: It's used by mainFragment. Either I keep it here or in mainfragment (which isn't good IMO since it's view layer). We'll slowly keep migrating to viewmodels and remove DataUtils singleton alltogether.

EmmanuelMess: Can't you replace all calls for `MainFragmentViewModel.getDataUtils()` with `DataUtils.getInstance()`?

app/src/main/java/com/amaze/filemanager/ui/fragments/data/MainFragmentViewModel.kt
VishalNehra: It's marked as Nullable

app/src/main/java/com/amaze/filemanager/ui/fragments/data/MainFragmentViewModel.kt
VishalNehra: It was initialized as null in previous implementation in MainFragment. Not taking any risk to change any logic for now.

app/src/main/java/com/amaze/filemanager/ui/fragments/AppsListFragment.java
VishalNehra: Updated PR description

app/src/main/java/com/amaze/filemanager/ui/fragments/data/MainFragmentViewModel.kt
EmmanuelMess: Yes, because you can put `null` in the default. In this case it can't be null AFAIK.

app/src/main/java/com/amaze/filemanager/ui/fragments/data/MainFragmentViewModel.kt
EmmanuelMess: An issue should be created to fix this. Will create.

app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
VishalNehra: For private members in MainFragment - it's initialized in oncreateview so it'll not be null. In public methods exposed by MainFragment I'm making that check everywhere until we have a better approach. We shouldn't be exposing fragment methods like this. These are last minute changes so I didn't want to make much changes than required.
It's either this or I put try catch in each method, only to replace them later by a better approach.

EmmanuelMess: > In public methods exposed by MainFragment I'm making that check everywhere until we have a better approach. 

Fixing [this](https://github.com/TeamAmaze/AmazeFileManager/pull/2551#pullrequestreview-676876595) should fix this as no more null check would be required. Especially inside the Fragment.

app/src/main/java/com/amaze/filemanager/ui/fragments/data/MainFragmentViewModel.kt
VishalNehra: But compiler suggest it can be null. Because `getString` returns `String?` and so I have to keep ?. at `toInt` as well. It's all compiler suggested. If I remove I need to use `!!`

app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
VishalNehra: For detached fragment, without try/catch, I don't see any other fix apart from making major changes in codebase and use callbacks / observers.

app/src/main/java/com/amaze/filemanager/ui/fragments/data/MainFragmentViewModel.kt
EmmanuelMess: I thought you could do a soft cast. Nevertheless you can still do a less-than-soft cast:
```kt
val a: String? = sharedPref.getString(PreferencesConstants.PREFERENCE_DIRECTORY_SORT_MODE, "0")
a as String // a will not be null because default value is not null.
```

app/src/main/java/com/amaze/filemanager/ui/fragments/data/MainFragmentViewModel.kt
VishalNehra: I just don't see the issue here. We can have soft references of singleton without any problems afaik. Regardless, I've made the change.

app/src/main/java/com/amaze/filemanager/ui/fragments/data/MainFragmentViewModel.kt
VishalNehra: Casting is dangerous. I don't see the urgent need. It's just a null check that too natively supported by kt. In future consider someone changes the default value to null but forgets to remove the casting, will be huge problem then.

EmmanuelMess: > Casting is dangerous.

Not here.

> I don't see the urgent need.

Ok.

EmmanuelMess:
Crash on search, go to another app, the activity gets destroyed and go back to amaze
```text
java.lang.NullPointerException: Attempt to invoke virtual method 'java.util.Iterator java.util.ArrayList.iterator()' on a null object reference
  at com.amaze.filemanager.adapters.RecyclerAdapter.setItems(RecyclerAdapter.java:405)
  at com.amaze.filemanager.adapters.RecyclerAdapter.<init>(RecyclerAdapter.java:176)
  at com.amaze.filemanager.ui.fragments.MainFragment.reloadListElements(MainFragment.java:1053)
  at com.amaze.filemanager.ui.fragments.MainFragment.addSearchResult(MainFragment.java:1666)
  at com.amaze.filemanager.ui.activities.MainActivity.onProgressUpdate(MainActivity.java:1961)
  at com.amaze.filemanager.asynchronous.asynctasks.SearchAsyncTask.onProgressUpdate(SearchAsyncTask.java:110)
  at com.amaze.filemanager.asynchronous.asynctasks.SearchAsyncTask.onProgressUpdate(SearchAsyncTask.java:39)
  at android.os.AsyncTask$InternalHandler.handleMessage(AsyncTask.java:671)
  at android.os.Handler.dispatchMessage(Handler.java:102)
  at android.app.ActivityThread.main(ActivityThread.java:5417)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
```

EmmanuelMess:
Nevermind, that crash is also on release/3.6

EmmanuelMess (approved):
I hope I didn't forget to test some critical crash somewhere.

EmmanuelMess (approved):

EmmanuelMess:
Please stop trying to bundle a lot of fixes in one PR.

PR #2552, opened 2021-06-05
Fix: replace add icon with cut icon (drag n drop)

### LINKED ISSUE #2544: Drag and drop icon shows a plus
**Describe the bug**
Drag and drop icon shows a plus, not the icon of the thing/s being dragged.

rbrishabh:
Fixes #2544

Hi @VishalNehra!

I don't think is quite right yet! I just replaced the add icon with the cut icon for now (a small win that I couldn't find where to actually change the code)

Do we need to add a new custom cut/copy icon, or does one already exist?

Looking forward to your review!

thanks!

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/adapters/RecyclerAdapter.java
+++ b/app/src/main/java/com/amaze/filemanager/adapters/RecyclerAdapter.java
@@ -963,7 +963,7 @@ public class RecyclerAdapter extends RecyclerView.Adapter<RecyclerView.ViewHolde
   }
 
   private int getDragIconReference(String rememberMovePreference) {
-    int iconRef = R.drawable.ic_add_white_24dp;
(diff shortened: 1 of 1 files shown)


EmmanuelMess:
> Do we need to add a new custom cut/copy icon, or does one already exist?

It probably doesn't exist. A fast look at https://fonts.google.com/icons?selected=Material+Icons&icon.query=copy doesn't reveal any promising icons IMO.

VishalNehra:
Something like this perhaps? I've seen this somewhere 🤔 
Ps. pardon my drawing skills 😅

EmmanuelMess:
> 
> Something like this perhaps? I've seen this somewhere thinking
> Ps. pardon my drawing skills sweat_smile

Yes, like that.

VishalNehra:
Closing till we have a relevant icon

PR #2594, opened 2021-06-07
Preference with multiple selection will show selected option as summary

rupenms:
Fix: Preference with multiple selection like Theme, number of grid columns and drag/drop preference settings will now show a summary of the current selected option below the title

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/PrefFrag.java
+++ b/app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/PrefFrag.java
@@ -99,6 +99,15 @@ public class PrefFrag extends PreferenceFragmentCompat
     addPreferencesFromResource(R.xml.preferences);
 
     sharedPref = PreferenceManager.getDefaultSharedPreferences(getActivity());
+    String themeTitle = sharedPref.getString(PreferencesConstants.FRAGMENT_THEME_SUMMARY,getString(R.string.light));
+    String numberOfColumns = sharedPref.getString(PreferencesConstants.PREFERENCE_GRID_COLUMNS,getString(R.string.default_string));
+    if (numberOfColumns.equals("-1")) {
+      numberOfColumns = getString(R.string.default_string);
+    }
+    String dragDropPreference = sharedPref.getString(PreferencesConstants.PREFERENCE_DRAG_AND_DROP_SUMMARY,getString(R.string.drag_to_select));
+    findPreference(PreferencesConstants.FRAGMENT_THEME).setSummary(themeTitle);
+    findPreference(PreferencesConstants.PREFERENCE_GRID_COLUMNS).setSummary(numberOfColumns);
+    findPreference(PreferencesConstants.PREFERENCE_DRAG_AND_DROP_PREFERENCE).setSummary(dragDropPreference);
 
     for (String PREFERENCE_KEY : PREFERENCE_KEYS) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/PrefFrag.java
```diff
@@ -231,12 +241,14 @@ public boolean onPreferenceClick(Preference preference) {
                   sharedPref
                       .edit()
                       .putInt(PreferencesConstants.PREFERENCE_DRAG_AND_DROP_PREFERENCE, which)
+                      .putString(PreferencesConstants.PREFERENCE_DRAG_AND_DROP_SUMMARY, getDragDropSummary(which))
```
EmmanuelMess, line 244: Why are you adding a new thing to the SharedPreferences?

rupenms: for storing the selected option text in the preference so that when preferences are created again it will display stored selection in summary

VishalNehra: This is absolutely not necessary. just set the summary based on preference value you're getting from preferences dynamically.

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/PrefFrag.java
```diff
@@ -231,12 +241,14 @@ public boolean onPreferenceClick(Preference preference) {
...
                       .apply();
                   sharedPref
                       .edit()
                       .putString(PreferencesConstants.PREFERENCE_DRAG_AND_DROP_REMEMBERED, null)
                       .apply();
                   dialog.dismiss();
+                  restartPC(getActivity());
```
VishalNehra, line 251: Why do you want to restart?

VishalNehra (changes requested):

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/PrefFrag.java
rupenms: we get only the preference id we dont get title for that id so for that I have used a seperate preference constant to store title which I am fetching from that id through getDragDropSummary method

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/PrefFrag.java
rupenms: settings page doesnot reflect the selected option as summary immediately after clicking on the option you need to refresh the page same as setting a theme

EmmanuelMess: Are you can't just invalidate the specific preference?

rupenms: I dont think so, the preferences are loaded from xml file in oncreatepreferences. In order to reflect that selected option the preferences needs to be loaded again from xml file. and as its loaded on oncreatepreferences the page needs to be refreshed in order to reflect the selected option as summary

EmmanuelMess:
Don't save more information than it is already in SharedPreferences for this.

PR #2608, opened 2021-06-11
Fix value for legacy_listing_summary

TranceLove:
## Description
Previous string value for `legacy_listing_summary` in string.xml contradicts its behaviour - Amaze will use old method (ls) to list files when preference PreferenceConstants.PREFERENCE_ROOT_LEGACY_LISTING is true, while the translation string reads contradictory meaning.

#### Manual tests
- [x] Done  
  
- Device: Fairphone 3
- OS: LineageOS 16.0 (9.0)

#### Build tasks success  

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -696,7 +696,7 @@
     <string name="clear_open_file_summary">Clears selected default file opening apps</string>
     <string name="ftp_readonly">Read-only access</string>
     <string name="legacy_listing_title">Use legacy listing for root</string>
-    <string name="legacy_listing_summary">If enabled, uses stat for listing, else uses ls</string>
+    <string name="legacy_listing_summary">If enabled, uses legacy method to list files</string>
     <string name="error_unsupported_v5_rar">RAR archive \"%s\" is unsupported RAR v5 archive.</string>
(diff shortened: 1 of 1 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -696,7 +696,7 @@
     <string name="clear_open_file_summary">Clears selected default file opening apps</string>
     <string name="ftp_readonly">Read-only access</string>
     <string name="legacy_listing_title">Use legacy listing for root</string>
-    <string name="legacy_listing_summary">If enabled, uses stat for listing, else uses ls</string>
+    <string name="legacy_listing_summary">If enabled, uses old method (ls) for listing files, else uses stat (new method)</string>
```
VishalNehra, line 699: Old method and new method is kinda redundant. Plus the ls and stat is something user wouldn't care. We should minimise text here.
Say 'If enabled uses legacy listing method, else uses new method'

VishalNehra (approved):

app/src/main/res/values/strings.xml
TranceLove: I'll then say **If enabled, uses legacy method to list files**.

VishalNehra (approved):

PR #2649, opened 2021-06-20
Verify package name before initializing ACRA and ErrorActivity

### LINKED ISSUE #2641: Restrict crash reporting to amaze package name
Due to limited resources quota on cloud platform and email service, better check package name before allowing user to send crash reports. 
Also show a message to user stating that their version is not official within the app itself instead of replying to their email.

TranceLove:
Implementation is a bit primitive, and given the source code is publicly available, we can only slow down, but not prohibiting others from pirating without a decent code signing mechanism that will work across github, f-droid and derivatives, and google play.

#### Issue tracker   
Fixes #2641

#### Manual tests
- [x] Done  
  
- Device: Pixel 2 emulator
- OS: Android 11

that ErrorActivity will immediately finish() if app package name doesn't start with our designated bundle ID.

#### Build tasks success  

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/application/AppConfig.java
+++ b/app/src/main/java/com/amaze/filemanager/application/AppConfig.java
@@ -109,7 +109,9 @@ public class AppConfig extends GlideApplication {
   @Override
   protected void attachBaseContext(Context base) {
     super.attachBaseContext(base);
-    initACRA();
+    if (base.getPackageName().equals(BuildConfig.APPLICATION_ID)) {
+      initACRA();
+    }
   }
 
   @Override
--- a/app/src/main/java/com/amaze/filemanager/crashreport/ErrorActivity.java
+++ b/app/src/main/java/com/amaze/filemanager/crashreport/ErrorActivity.java
@@ -126,11 +126,13 @@ public class ErrorActivity extends ThemedActivity {
 
   private static void startErrorActivity(
       final Context context, final ErrorInfo errorInfo, final List<Throwable> el) {
-    final Intent intent = new Intent(context, ErrorActivity.class);
-    intent.putExtra(ERROR_INFO, errorInfo);
-    intent.putExtra(ERROR_LIST, elToSl(el));
-    intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
-    context.startActivity(intent);
+    if (context.getPackageName().equals(BuildConfig.APPLICATION_ID)) {
+      final Intent intent = new Intent(context, ErrorActivity.class);
+      intent.putExtra(ERROR_INFO, errorInfo);
+      intent.putExtra(ERROR_LIST, elToSl(el));
+      intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+      context.startActivity(intent);
+    }
   }
 
   public static void reportError(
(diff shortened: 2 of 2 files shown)


EmmanuelMess:
> Implementation is a bit primitive, and given the source code is publicly available, we can only slow down, but not prohibiting others from pirating without a decent code signing mechanism that will work across github, f-droid and derivatives, and google play.

Protection via obfuscation would be ideal here.

app/src/main/java/com/amaze/filemanager/application/AppConfig.java
```diff
@@ -109,7 +109,9 @@ public void onCreate() {
   @Override
   protected void attachBaseContext(Context base) {
     super.attachBaseContext(base);
-    initACRA();
+    if (base.getPackageName().equals(BuildConfig.APPLICATION_ID)) {
```
EmmanuelMess, line 112: Isn't build config generated on compile?

app/src/main/java/com/amaze/filemanager/crashreport/ErrorActivity.java
```diff
@@ -149,6 +151,7 @@ public static void reportError(
     final String[] el = new String[] {report.getString(ReportField.STACK_TRACE)};
 
     final Intent intent = new Intent(context, ErrorActivity.class);
+    intent.setPackage("context.getPackageName()");
```
EmmanuelMess, line 154: did you forget to delete the quotes?

app/src/main/java/com/amaze/filemanager/crashreport/ErrorActivity.java
```diff
@@ -174,74 +177,77 @@ private static String getStackTrace(final Throwable throwable) {
...
-          Utils.openURL(ERROR_GITHUB_ISSUE_URL, this);
-        });
-
-    // normal bugreport
-    buildInfo(errorInfo);
-    if (errorInfo.message != 0) {
-      errorMessageView.setText(errorInfo.message);
+    if (ErrorInfo.comparePackageInfo(intent.getPackage()) != 0) {
+      finish();
     } else {
```
EmmanuelMess, line 183: No else needed AFAIK.

EmmanuelMess (changes requested):
See comments

TranceLove:
@EmmanuelMess Surely do.

I found this for encrypting strings, but it also noted degrading performance if too many strings are encrypted. Possibly some code reviews required before proceeding.
 https://github.com/christopherney/Enigma

TranceLove:
@TranceLove But also, since other peiple can have full access to source code, it seems code signing shall be the more appropriate method.

app/src/main/java/com/amaze/filemanager/crashreport/ErrorActivity.java
TranceLove: At least a `return` to get out of there would make me feel more safe ;)

app/src/main/java/com/amaze/filemanager/crashreport/ErrorActivity.java
```diff
@@ -496,5 +502,9 @@ public void writeToParcel(final Parcel dest, final int flags) {
       dest.writeString(this.request);
       dest.writeInt(this.message);
     }
+
+    public static int comparePackageInfo(String packageName) {
+      return packageName.indexOf("com.amaze.filemanager");
```
VishalNehra, line 507: How is this going to work on debug package?

VishalNehra (changes requested):

app/src/main/java/com/amaze/filemanager/crashreport/ErrorActivity.java
TranceLove:
```kotlin
"com.amaze.filemanager.debug".indexOf("com.amaze.filemanager") == 0
```

EmmanuelMess (approved):

PR #2667, opened 2021-07-03
Fix split apks broken icon

### LINKED ISSUE #2360: Split apks have broken icons
[AmazeFileManager.zip](https://github.com/TeamAmaze/AmazeFileManager/files/6137147/AmazeFileManager.zip)

imagen

HoussemNasri:
## Description
- `ApkImageDataFetcher` is responsible for fetching apk icons, it does so by calling `getPackageManager().getPackageArchiveInfo()`, this method will return `null` when it fails to parse the apk (corrupted apk, split apk...), my solution is to set a placeholder icon for the apk when this method returns `null`. 
- the build failed because `core-ktx` version was set to `.+` which will evaluate to `1.7.0-alpha01`, this version has a `minCompileSdk` of `30` but the current `compileSdkVersion` of Amaze is `29`, I fixed it by setting version to the last stable one `1.6.0`.
https://issuetracker.google.com/issues/192484318
#### Issue tracker   
Fixes #2360 

#### Automatic tests

- [ ] Added test cases
  
#### Manual tests
- [x] Done  
  
- Device: PIXEL API 29
- OS: Android 10.0

#### Build tasks success  


HoussemNasri:
# Scrennshots

> Material Light



> Material Dark



> Black (for OLED)

app/build.gradle
```diff
@@ -226,7 +226,7 @@ dependencies {
 
     implementation project(':commons_compress_7z')
     implementation project(':file_operations')
-    implementation "androidx.core:core-ktx:+"
+    implementation "androidx.core:core-ktx:1.6.0"
```
VishalNehra, line 229: Please use `implementation "androidx.core:core-ktx:$kotlin_version"`

VishalNehra (changes requested):

app/build.gradle
HoussemNasri: Done.

VishalNehra (approved):

app/src/main/res/drawable/ic_android_white_24dp.xml
```diff
@@ -4,6 +4,6 @@
         android:viewportWidth="24.0"
         android:viewportHeight="24.0">
     <path
-        android:fillColor="@color/accent_black"
```
EmmanuelMess, line 7: We used this for the drawer, won't the drawer have a green android icon?

EmmanuelMess (changes requested):
Code reviews

app/src/main/res/drawable/ic_android_white_24dp.xml
HoussemNasri: I thought the same thing, but it seems the original color is ignored and replaced with a theme color, `?attr/colorPrimary` when active else `?attr/colorOnSurface`.
Although I am still not sure why APK files icon kept their color, maybe because the context object doesn't have a theme associated with it?
https://material.io/components/navigation-drawer/android#anatomy

EmmanuelMess (approved):
Tested on AVD API 23 and LG K11alpha

EmmanuelMess:
(We'll merge as soon as Codacy ends scheduled maintenance)

PR #2708, opened 2021-07-27
Implement DocumentFile mode of HybridFile

### LINKED ISSUE #2015: Android 11, Can't access /storage/emulated/0/android/data

TranceLove:
For fixing #2015, so ugly assumptions here and there.

- Added `DOCUMENT_FILE` mode in `HybridFile`, with the internals sharing code with `OTG` where possible
- Implement `SafRootHolder` for holding the SAF root URI. Ugly, and will have problems when two tabs are handling different SAF root URIs - e.g. one is at `Android/data` and one is at `Android/obb` - but OTG implementation is having the same problem on handling multiple OTG devices, so...

#### Issue tracker   
Fixes #2015

#### Automatic tests
- [ ] Added test cases
  
#### Manual tests
- [x] Done  
  
- Device: Pixel 2 emulator
- OS: Android 11
Tested for create/rename/delete folder, file, open text file in `TextEditorActivity`.

#### Build tasks success  

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/DeleteTask.java
+++ b/app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/DeleteTask.java
@@ -31,6 +31,7 @@ import com.amaze.filemanager.database.CryptHandler;
 import com.amaze.filemanager.file_operations.exceptions.ShellNotRunningException;
 import com.amaze.filemanager.filesystem.HybridFile;
 import com.amaze.filemanager.filesystem.HybridFileParcelable;
+import com.amaze.filemanager.filesystem.SafRootHolder;
 import com.amaze.filemanager.filesystem.cloud.CloudUtil;
 import com.amaze.filemanager.filesystem.files.CryptUtil;
 import com.amaze.filemanager.filesystem.files.FileUtils;
@@ -67,14 +68,14 @@ public class DeleteTask
   private final DataUtils dataUtils = DataUtils.getInstance();
 
   public DeleteTask(@NonNull Context cd) {
-    this.cd = cd;
+    this.cd = cd.getApplicationContext();
     rootMode =
         PreferenceManager.getDefaultSharedPreferences(cd)
             .getBoolean(PreferencesConstants.PREFERENCE_ROOTMODE, false);
   }
 
   public DeleteTask(@NonNull Context cd, CompressedExplorerFragment compressedExplorerFragment) {
-    this.cd = cd;
+    this.cd = cd.getApplicationContext();
     rootMode =
         PreferenceManager.getDefaultSharedPreferences(cd)
             .getBoolean(PreferencesConstants.PREFERENCE_ROOTMODE, false);
@@ -154,6 +155,10 @@ public class DeleteTask
     switch (file.getMode()) {
       case OTG:
         DocumentFile documentFile = OTGUtil.getDocumentFile(file.getPath(), cd, false);
+        return documentFile.delete();
+      case DOCUMENT_FILE:
+        documentFile =
+            OTGUtil.getDocumentFile(file.getPath(), SafRootHolder.getUriRoot(), cd, false);
         return documentFile.delete();
       case DROPBOX:
       case BOX:
--- a/app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/DeleteTask.java
+++ b/app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/DeleteTask.java
@@ -62,22 +62,23 @@ public class DeleteTask
     extends AsyncTask<ArrayList<HybridFileParcelable>, String, AsyncTaskResult<Boolean>> {
 
   private ArrayList<HybridFileParcelable> files;
-  private final Context cd;
+  private final Context applicationContext;
   private final boolean rootMode;
   private CompressedExplorerFragment compressedExplorerFragment;
   private final DataUtils dataUtils = DataUtils.getInstance();
 
-  public DeleteTask(@NonNull Context cd) {
-    this.cd = cd.getApplicationContext();
+  public DeleteTask(@NonNull Context applicationContext) {
+    this.applicationContext = applicationContext.getApplicationContext();
     rootMode =
-        PreferenceManager.getDefaultSharedPreferences(cd)
+        PreferenceManager.getDefaultSharedPreferences(applicationContext)
             .getBoolean(PreferencesConstants.PREFERENCE_ROOTMODE, false);
   }
 
-  public DeleteTask(@NonNull Context cd, CompressedExplorerFragment compressedExplorerFragment) {
-    this.cd = cd.getApplicationContext();
+  public DeleteTask(
+      @NonNull Context applicationContext, CompressedExplorerFragment compressedExplorerFragment) {
+    this.applicationContext = applicationContext.getApplicationContext();
     rootMode =
-        PreferenceManager.getDefaultSharedPreferences(cd)
+        PreferenceManager.getDefaultSharedPreferences(applicationContext)
             .getBoolean(PreferencesConstants.PREFERENCE_ROOTMODE, false);
     this.compressedExplorerFragment = compressedExplorerFragment;
   }
@@ -85,7 +86,7 @@ public class DeleteTask
   @Override
   protected void onProgressUpdate(String... values) {
     super.onProgressUpdate(values);
-    Toast.makeText(cd, values[0], Toast.LENGTH_SHORT).show();
+    Toast.makeText(applicationContext, values[0], Toast.LENGTH_SHORT).show();
   }
 
   @Override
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/DeleteTask.java
```diff
@@ -106,14 +108,14 @@ protected void onProgressUpdate(String... values) {
       // delete file from media database
       if (!file.isSmb()) {
         try {
-          deleteFromMediaDatabase(cd, file.getPath());
+          deleteFromMediaDatabase(cd.get(), file.getPath());
```
EmmanuelMess, line 111: Stop doing this, `cd.get()` could return null unpredictably, if you need context during the background operations you need a reference to the application context.

app/src/main/java/com/amaze/filemanager/filesystem/EditableFileAbstraction.java
```diff
@@ -57,23 +57,27 @@ public EditableFileAbstraction(@NonNull Context context, @NonNull Uri uri) {
...
+              */
+              tempName = c.getString(c.getColumnIndex(OpenableColumns.DISPLAY_NAME));
+            } catch (Exception e) {
+              tempName = null;
+            }
+            c.close();
           }
-          c.close();
+        } catch (UnsupportedOperationException e) {
+          tempName = uri.getLastPathSegment();
```
EmmanuelMess, line 80: What's this for?

app/src/main/java/com/amaze/filemanager/filesystem/HybridFileParcelable.java
```diff
@@ -154,6 +156,14 @@ public void setPermission(String permission) {
     this.permission = permission;
   }
 
+  public Uri getFullUri() {
+    return fullUri;
```
EmmanuelMess, line 160: This should crash if the full uri is obtained for anything that is not a DOCUMENT_FILE.

app/src/main/java/com/amaze/filemanager/filesystem/HybridFileParcelable.java
```diff
@@ -154,6 +156,14 @@ public void setPermission(String permission) {
...
+  }
+
+  public void setFullUri(Uri fullUri) {
+    this.fullUri = fullUri;
```
EmmanuelMess, line 164: This should crash if the full uri is added for anything that is not a DOCUMENT_FILE.

app/src/main/java/com/amaze/filemanager/filesystem/Operations.java
```diff
@@ -143,17 +161,21 @@ protected Void doInBackground(Void... params) {
...
-          } else errorCallBack.done(file, false);
+          safCreateDirectory.apply(
+              new Pair<>(
+                  OTGUtil.getDocumentFile(file.getPath(), context, false),
+                  OTGUtil.getDocumentFile(file.getParent(context), context, false)));
+          return null;
+        } else if (file.isDocumentFile()) {
+          DocumentFile toCreate =
+              OTGUtil.getDocumentFile(
+                  file.getPath(), SafRootHolder.INSTANCE.getUriRoot(), context, false);
```
EmmanuelMess, line 172: Can't `getUriRoot` be JvmStatic?

app/src/main/java/com/amaze/filemanager/filesystem/SafRootHolder.kt
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.amaze.filemanager.filesystem
+
+import android.net.Uri
+
+object SafRootHolder {
```
EmmanuelMess, line 25: Make functions here that set up the fields, then call them from where it is appropiate, as there is quite a few lines of code that need to be executed to set up the process and it should be centralized.

EmmanuelMess (changes requested):

app/src/main/java/com/amaze/filemanager/filesystem/EditableFileAbstraction.java
EmmanuelMess: The catch I mean, why is it there.

TranceLove: Yes of course. The problem is, I even forgotten why I inserted this catch at the beginning 😵️ must be to prevent some empty cursor I think.

This is a good example of what a good programmer should do - document as you code.

app/src/main/java/com/amaze/filemanager/filesystem/SafRootHolder.kt
TranceLove: Sorry but I'm a bit confused about the code for setting up the `SafRootHolder`.

EmmanuelMess: It seems this code is used: 

```java
  private ActivityResultLauncher<Intent> handleDocumentUriForRestrictedDirectories =
      registerForActivityResult(
          new ActivityResultContracts.StartActivityForResult(),
          result -> {
            if (SDK_INT >= Q) {
              getContext()
                  .getContentResolver()
                  .takePersistableUriPermission(
                      result.getData().getData(),
                      Intent.FLAG_GRANT_READ_URI_PERMISSION
                          | Intent.FLAG_GRANT_WRITE_URI_PERMISSION);
              SafRootHolder.setUriRoot(result.getData().getData());
              loadlist(result.getData().getDataString(), false, OpenMode.DOCUMENT_FILE);
            }
          });
```

app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
```diff
@@ -928,12 +958,53 @@ public void loadlist(final String path, final boolean back, final OpenMode openM
       loadFilesListTask.cancel(true);
     }
 
+    OpenMode _openMode = openMode;
+    String _path = FileProperties.remapPathForApi30OrAbove(path, false);
```
EmmanuelMess, line 962: Don't use this naming convention to keep the variable final.

app/src/main/java/com/amaze/filemanager/utils/OTGUtil.kt
```diff
@@ -118,27 +141,47 @@ object OTGUtil {
...
-                nextDocument = rootUri.createFile(part.substring(part.lastIndexOf(".")), part)
+            if (part.startsWith("primary%3AA")) {
+                part.substringAfterLast(
+                    SafRootHolder.uriRoot.toString()
+                        .substringAfterLast("/")
+                ).split("%2F").filterNot { it.isEmpty() or it.isBlank() }.forEach {
+                    retval = retval!!.findFile(it)
+                }
+            } else {
+                var nextDocument = retval!!.findFile(part)
```
EmmanuelMess, line 177: `!!`

app/src/main/java/com/amaze/filemanager/filesystem/SafRootHolder.kt
TranceLove: Sorry but I'm really out of ideas at this point :(

EmmanuelMess: How do you know the uses of these variables won't NPE? Is there an intuive way to check the flow?

EmmanuelMess (dismissed):
This will probably need some refractoring as we move to include newer android versions, but it is ok for now.

TranceLove:
build test apk

VishalNehra:
Few this here:
1. Can we aim this for 3.6.2 as lots of users are asking for this and there's still time for 3.7
2. I notice pressing back button when I'm inside android/data leads me back to internal storage root instead of /android
3. I'm getting a crash when i press on bottom bar and from paths i choose android folder

It's thinking that current filesystem is OTG? 

```
## Issue explanation (write below this line)



## Exception
* __App Name:__ Amaze File Manager
* __Package:__ com.amaze.filemanager.debug
* __Version:__ 3.6.0
* __User Action:__ UI Error
* __OS:__ Linux Android 11 - 30
* __Device:__ OnePlus9R
* __Model:__ LE2101
* __Product:__ OnePlus9R_IND
<details><summary><b>Crash log </b></summary><p>

```
java.lang.RuntimeException: An error occurred while executing doInBackground()
	at android.os.AsyncTask$4.done(AsyncTask.java:415)
	at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:383)
	at java.util.concurrent.FutureTask.setException(FutureTask.java:252)
	at java.util.concurrent.FutureTask.run(FutureTask.java:271)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
	at java.lang.Thread.run(Thread.java:923)
Caused by: java.lang.NullPointerException
	at com.amaze.filemanager.utils.OTGUtil.getDocumentFiles(OTGUtil.kt:118)
	at com.amaze.filemanager.asynchronous.asynctasks.LoadFilesListTask.listDocumentFiles(LoadFilesListTask.java:570)
	at com.amaze.filemanager.asynchronous.asynctasks.LoadFilesListTask.doInBackground(LoadFilesListTask.java:209)
	at com.amaze.filemanager.asynchronous.asynctasks.LoadFilesListTask.doInBackground(LoadFilesListTask.java:69)
	at android.os.AsyncTask$3.call(AsyncTask.java:394)
	... 3 more

```
</details>
<hr>
```

TranceLove:
Build failed with unable to publish test report to Codacy due to missing API token in env vars. Did I miss anything?

PR #2719, opened 2021-08-06
Dark Theme

### LINKED ISSUE #2056: Support system dark theme on Android 10
Per user feedback on Google Play. May become an theme selection option in preferences.

Reference: https://developer.android.com/guide/topics/ui/look-and-feel/darktheme

VishnuSanal:
#### Issue tracker
Fixes #2056 

#### Automatic tests
- [ ] Added test cases 
  
#### Manual tests
- [x] Done 

#### Build tasks success

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -57,7 +57,8 @@
             android:label="@string/appbar_name"
             android:launchMode="singleInstance"
             android:name=".ui.activities.MainActivity"
-            android:theme="@style/appCompatBlack">
+            android:theme="@style/appCompatBlack"
+            android:configChanges="uiMode" >
 
             <intent-filter android:label="@string/appbar_name">
                 <action android:name="android.intent.action.MAIN" />
@@ -121,7 +122,7 @@
 
         <activity
             android:uiOptions="splitActionBarWhenNarrow"
-            android:configChanges="orientation|screenSize"
+            android:configChanges="orientation|screenSize|uiMode"
             android:label="@string/setting"
             android:name=".ui.activities.PreferencesActivity"
             android:theme="@style/appCompatBlack">
@@ -134,6 +135,7 @@
         <activity
             android:label="@string/textreader"
             android:name=".ui.activities.texteditor.TextEditorActivity"
+            android:configChanges="uiMode"
             android:theme="@style/appCompatBlack">
             <intent-filter
                 tools:ignore="AppLinkUrlError"
@@ -151,6 +153,7 @@
             android:label="@string/databasereader"
             android:name=".ui.activities.DatabaseViewerActivity"
             android:theme="@style/appCompatBlack"
+            android:configChanges="uiMode"
             android:screenOrientation="locked">
             <intent-filter
                 tools:ignore="AppLinkUrlError"
@@ -164,11 +167,13 @@
 
         <activity android:name=".ui.activities.AboutActivity"
             android:theme="@style/aboutBlack"
+            android:configChanges="uiMode"
             android:label="About"
             />
 
         <activity android:name=".crashreport.ErrorActivity"
             android:theme="@style/appCompatBlack"
+            android:configChanges="uiMode"
             android:label="Error"
             />
 
--- a/app/src/main/java/com/amaze/filemanager/crashreport/ErrorActivity.java
+++ b/app/src/main/java/com/amaze/filemanager/crashreport/ErrorActivity.java
@@ -40,11 +40,15 @@ import com.amaze.filemanager.R;
 import com.amaze.filemanager.filesystem.files.FileUtils;
 import com.amaze.filemanager.ui.activities.MainActivity;
 import com.amaze.filemanager.ui.activities.superclasses.ThemedActivity;
+import com.amaze.filemanager.ui.fragments.preference_fragments.PreferencesConstants;
+import com.amaze.filemanager.ui.theme.AppTheme;
 import com.amaze.filemanager.utils.Utils;
 import com.google.android.material.snackbar.Snackbar;
 
+import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
+import android.content.res.Configuration;
 import android.graphics.Color;
 import android.os.Build;
 import android.os.Bundle;
@@ -61,6 +65,7 @@ import android.widget.EditText;
 import android.widget.TextView;
 import android.widget.Toast;
 
+import androidx.annotation.NonNull;
 import androidx.annotation.StringRes;
 import androidx.appcompat.app.ActionBar;
 import androidx.appcompat.widget.Toolbar;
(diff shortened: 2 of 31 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -121,7 +122,7 @@
 
         <activity
             android:uiOptions="splitActionBarWhenNarrow"
-            android:configChanges="orientation|screenSize"
+            android:configChanges="orientation|screenSize|uiMode"
```
EmmanuelMess, line 125: Why?

app/src/main/java/com/amaze/filemanager/ui/activities/AboutActivity.java
```diff
@@ -235,7 +271,10 @@ public void onClick(View v) {
                 .withAboutSpecial1Description(getString(R.string.amaze_license))
                 .withLicenseShown(true);
 
-        switch (getAppTheme().getSimpleTheme()) {
+        switch (getAppTheme()
+            .getSimpleTheme(
+                (getResources().getConfiguration().uiMode & Configuration.UI_MODE_NIGHT_MASK)
```
EmmanuelMess, line 276: Deduplicate this, it seems to be copied *a lot* of times.

EmmanuelMess (changes requested):

app/src/main/java/com/amaze/filemanager/crashreport/ErrorActivity.java
```diff
@@ -271,6 +276,38 @@ public boolean onOptionsItemSelected(final MenuItem item) {
     return false;
   }
 
+  @Override
+  public void onConfigurationChanged(@NonNull Configuration newConfig) {
```
EmmanuelMess, line 280: These methods are duplicated a few times. Try to minimize code duplications, maybe a class hierarchy?

app/src/main/java/com/amaze/filemanager/ui/theme/AppTheme.java
```diff
@@ -59,14 +61,16 @@ public static AppTheme getTheme(int index) {
...
+        return SYSTEM;
     }
   }
 
   /**
    * @return The Theme enum to provide to {@link
    *     com.afollestad.materialdialogs.MaterialDialog.Builder}
    */
-  public Theme getMaterialDialogTheme() {
+  public Theme getMaterialDialogTheme(boolean isNightMode) {
```
EmmanuelMess, line 73: The code for "isNightTheme" is always the same, who not write it here?

EmmanuelMess (changes requested):

app/src/main/AndroidManifest.xml
VishnuSanal: To handle the theme changes whilst the app is being used (as mentioned [here](https://developer.android.com/guide/topics/ui/look-and-feel/darktheme#config-changes))

app/src/main/java/com/amaze/filemanager/ui/theme/AppTheme.java
VishnuSanal: It is indeed the same, but can we get the `Context` for doing that in that enum? This is why I had to pass that boolean everytime the method gets called!

`(c.getResources().getConfiguration().uiMode & Configuration.UI_MODE_NIGHT_MASK) == Configuration.UI_MODE_NIGHT_YES))`

EmmanuelMess: > It is indeed the same, but can we get the `Context` for doing that in that enum? This is why I had to pass that boolean everytime the method gets called!
> 
> `(c.getResources().getConfiguration().uiMode & Configuration.UI_MODE_NIGHT_MASK) == Configuration.UI_MODE_NIGHT_YES))`

Getting a Context generally hinges on whether keeping a strong reference is a good idea or not. In this case the strong reference shouldn't be kept longer than the scope of this method.

app/src/main/AndroidManifest.xml
EmmanuelMess: Why don't we want an automatic restart of the activities? In which activities wouldn't we want an automatic restart?

app/src/main/java/com/amaze/filemanager/crashreport/ErrorActivity.java
EmmanuelMess: I think we had a ThemedActivity to deal with themes, but I am unsure how adaptable that class is for this.

VishnuSanal: > I think we had a ThemedActivity to deal with themes, but I am unsure how adaptable that class is for this.

I've moved 'em to `PreferenceActivity`...

app/src/main/AndroidManifest.xml
VishnuSanal: > Why don't we want an automatic restart of the activities? In which activities wouldn't we want an automatic restart?

 Didn't get you... We require the onConfigChanged() to get called when the change happens, hence the uiMode attribute...

EmmanuelMess: > Didn't get you... We require the onConfigChanged() to get called when the change happens, hence the uiMode attribute...

Not necessarily, you can just let the activity restart.

app/src/main/java/com/amaze/filemanager/crashreport/ErrorActivity.java
EmmanuelMess: > `PreferenceActivity`

Why? Isn't that related to configurations?

VishnuSanal: > Why?

`AboutActivity` extends `PreferenceActivity` directly (which is extended by `ThemedActivity` too), so I chose `PreferenceActivity`.

But, moving to `ThemedActivity` would work too, by updating `AboutActivity`, will do :)

app/src/main/AndroidManifest.xml
EmmanuelMess: Probably only some activities should update their state not restarting.

VishnuSanal: I'm sorry, I didn't get you again. `uiMode` was added to get `onConfigurationChanged()` called when the user changes the system-level dark theme whilst using our app - so that we can apply our dark theme in the app.

app/src/main/java/com/amaze/filemanager/ui/activities/AboutActivity.java
VishnuSanal: ✅ Done

app/src/main/java/com/amaze/filemanager/crashreport/ErrorActivity.java
VishnuSanal: ✅ Done

app/src/main/java/com/amaze/filemanager/ui/theme/AppTheme.java
VishnuSanal: ✅ Done

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/ColorPref.kt
```diff
@@ -303,17 +305,19 @@ class ColorPref : PreferenceFragmentCompat(), Preference.OnPreferenceClickListen
...
+
             selectedColors.setColors(
                 userColorPreferences.primaryFirstTab,
                 userColorPreferences.primarySecondTab,
                 userColorPreferences.accent,
                 userColorPreferences.iconSkin
             )
-            if (activity.appTheme.materialDialogTheme == Theme.LIGHT) {
+
+            if (activity.appTheme.getMaterialDialogTheme(activity.applicationContext) == Theme.LIGHT)
```
EmmanuelMess, line 316: Fix this brackets, @VishalNehra this shouldn't pass codacy like this

EmmanuelMess (approved):

VishalNehra (approved):

PR #2723, opened 2021-08-08
Make SAF-based backend optional in FTP server

TranceLove:
As of 3.6.x cycle the DocumentFile-based implementation for FTP server is causing problems to a number of users. This PR adds an option to allow user to fallback to the original java.io.File based filesystem.

If user had been using the DocumentFile-based UI but changed back to java.io.File based filesystem, shared path will be reset to internal storage's root; user will need to change the path on their own.

(It should be possible to help users to migrate the path between the two FTP filesystem implementations however.)

I am making this against 3.6.2, but just take it against 3.7 if you think it's better.

#### Automatic tests
- [ ] Added test cases
  
#### Manual tests
- [x] Done  
  
- Device: Fairphone 3
- OS: LineageOS 16.0 (9.0)
- Device: Pixel 2 emulator
- OS: Android 11

#### Build tasks success  

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/asynchronous/services/ftp/FtpService.kt
+++ b/app/src/main/java/com/amaze/filemanager/asynchronous/services/ftp/FtpService.kt
@@ -29,7 +29,7 @@ import android.content.SharedPreferences
 import android.net.ConnectivityManager
 import android.net.NetworkCapabilities
 import android.net.wifi.WifiManager
-import android.os.Build
+import android.os.Build.VERSION.SDK_INT
 import android.os.Build.VERSION_CODES.KITKAT
 import android.os.Build.VERSION_CODES.M
 import android.os.Environment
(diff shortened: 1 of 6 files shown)


EmmanuelMess:
Sorry, I am a bit lost, how would a user experiencing the issue fix it from the FTP fragment itself? is there a checkbox?

TranceLove:
@EmmanuelMess Actually if time allows I'd think of making FTP server settings a preference page😏

EmmanuelMess:
> @EmmanuelMess Actually if time allows I'd think of making FTP server settings a preference pagesmirk

That wouldn't really help though 😅 the user should be able to fix the issue from the ftp fragment. Having the preference is not a bad idea, but *only* having the preference is IMO.

EmmanuelMess (approved):

VishalNehra (approved):

VishalNehra:
Can close #2723 ?

PR #2806, opened 2021-09-15
Fix MainActivity does not respond to view compressed file intents

### LINKED ISSUE #2802: MainActivity should handle view intents of compressed archives
**Describe the bug**
Intents to view compressed archives should trigger open archive dialog.

Similar problem also affects users having both release and debug builds installed on the same device.

**To Reproduce**
1. Choose an archive in Amaze
2. Instead of simply tap on the archive, tap on the three dots and choose "Open with"
3. Choose "Other" as file type
4. Activity list pops up
5. Choose Amaze again
6. No response in Amaze

**Expected behavior**
After 5 above, Archive open dialog should pop up, asking user to view or extract the archive, just like simply tapping the archive in Amaze.

 - Device: Pixel 2 emulator
 - OS: Android 6
 - Rooted: No
 - Version: includes, but not limited to, 3.6.1

**Additional context**
2 directions can be think of.
- To begin with, view intent triggered from Amaze should include compressed files' MIME type
- Alternately, in MainActivity handling view intents, possibly add code to detect compressed archives based on filename/URI
both directions should end up at popping up the archive open dialog.

TranceLove:
## Description
- Propagate MIME type when triggering OpenFileDialogFragment
- Detect URI on view intents in MainActivity. If it's a compressed file, triggers MainActivity.openzip()
- Use constant in MimeTypes.ALL_MIME_TYPES for "*/*"

#### Issue tracker   
Fixes #2802, as a band-it fix. Since the open archive dialog assumed supplied path is a local file, hence had to resort to use only CompressedExplorerFragment to handle which supports `content://` URIs (expecting potential problems though).

For a full fix, archive extraction have to support `content://` URIs too; alternately at least open archive dialog should be able to translate `content://` URIs to local files.

#### Manual tests
- [x] Done  
  
- Device: Pixel 2 emulator
- OS: Android 11
With scenario mentioned in #2802, after selecting Amaze as app to open compressed file Amaze should open the compressed file directly.

#### Build tasks success  


VishnuSanal:
Was testing this PR...

Click on a .zip file -> choose Amaze Debug -> the content opens -> click back button -> Amaze closes

I think this is a bad user flow. We would want to let the user back to the previous location. What do you think?

TranceLove:
> Was testing this PR...
> 
> Click on a .zip file -> choose Amaze Debug -> the content opens -> click back button -> Amaze closes
> 
> I think this is a bad user flow. We would want to let the user back to the previous location. What do you think?

OK, as of this PR, when back button is pressed, it goes to here:

https://github.com/TeamAmaze/AmazeFileManager/blob/31b33f8cc9f0b505bad902afa55fce55d90e03f5/app/src/main/java/com/amaze/filemanager/ui/activities/MainActivity.java#L848-L868

When the zip is opened in an Intent, `openzip` will be `true` here, causing line 854 to run, which will result in a `finish()`.

Any idea to work around this without breaking previous behaviour(?), or should this behaviour be changed? Note however, open compressed file Intents shall not be triggered solely from Amaze, but possibly from other apps as well.

VishnuSanal:
> Note however, open compressed file Intents shall not be triggered solely from Amaze, but possibly from other apps as well.

Hmm. makes sense. My concern can be a new issue.

VishalNehra (approved):

PR #2828, opened 2021-09-21
Implement DOCUMENT_FILE mode of HybridFile, for accessing Android/data for API >= 30

### LINKED ISSUE #2015: Android 11, Can't access /storage/emulated/0/android/data

TranceLove:
Supercedes #2708, including fixes by @VishalNehra. Original PR description below.
-------
For fixing #2015, so ugly assumptions here and there.

- Added `DOCUMENT_FILE` mode in `HybridFile`, with the internals sharing code with `OTG` where possible
- Implement `SafRootHolder` for holding the SAF root URI. Ugly, and will have problems when two tabs are handling different SAF root URIs - e.g. one is at `Android/data` and one is at `Android/obb` - but OTG implementation is having the same problem on handling multiple OTG devices, so...

#### Issue tracker   
Fixes #2015

#### Automatic tests
- [ ] Added test cases
  
#### Manual tests
- [x] Done  
  
- Device: Pixel 2 emulator
- OS: Android 11
Tested for create/rename/delete folder, file, open text file in `TextEditorActivity`.

#### Build tasks success  

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/DeleteTask.java
+++ b/app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/DeleteTask.java
@@ -29,6 +29,7 @@ import com.amaze.filemanager.R;
 import com.amaze.filemanager.application.AppConfig;
 import com.amaze.filemanager.database.CryptHandler;
 import com.amaze.filemanager.file_operations.exceptions.ShellNotRunningException;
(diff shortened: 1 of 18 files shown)


bittuilovu:
I am not ble to read my model file from "/storage/emulated/0/" or "/data/media/0/Android/data/" in android JNI c++

PR #2912, opened 2021-10-16
Fix SSH with password authentication

### LINKED ISSUE #2843: SCP/SFTP support does not work correctly
**Describe the bug**
I have set-up an SCP connection to a Linux system, Amaze is able to connect to it and displays the initial directory content, but it is not able to display the content of sub-folders, every selected folder remains empty. 

Additionally if you select a file or folder and execute the "Copy" function nothing happens (no "persistent toast" is shown that allows you to copy a file to a different directory.

Also viewing a file on a SCP mount is not possible, if you do that Chrome opens on the address `127.0.0.1:7871/home 

From logcat I can see that `sftp subsystem`  is used.

**To Reproduce**
1. Set-up an SCP connection to a Linux system using username and key file
2. Select the SCP entry, enter the key password
3. The file and folder list from the Linux server is shown
4. Select a folder that contains files
5. The folder is opened but no files are displayed inside

Logcat does not show any exceptions (just the common failed attempts to load the ssh keyfile for the wrong format).

 - Device: Tablet FullHD
 - OS: Android 9
 - Rooted: No
 - Amaze Version: 3.6.1 from Google Play Store

TranceLove:
## Description
- Fix SftpEntryDao SQL declarations. sqlite is case insensitive, but just to be sure
- Use Kotlin's null/empty check in SshAuthenticationTask and SshConnectionPool
- Trim trailing \n when encrypting SSH passwords
- returns SSH path with encrypted password out to Drawer when calling UtilsHandler.getSftpList()
- Fix handling of encrypted passwords at SshClientUtils.extractBaseUriFrom(), use our own parsing routine instead of Uri.parse which will break when encrypted password contains slash characters

#### Issue tracker   
Fixes #2843

#### Automatic tests
- [x] Added test cases
On handling of SSH URI with encrypted passwords
  
#### Manual tests
- [x] Done  
  
Device: Pixel 2 emulator
OS: Android 11

Manual tests include
- create, update, delete SSH connection settings with SSH server requiring password authentication
- create, update, delete SSH connection settings with SSH server requiring key authentication
- open connection to SSH server should continue function as normal
- passwords, even encrypted passwords, should never be visible at Amaze's bottom bar

#### Build tasks success  


TranceLove:
Build test apk

PR #2966, opened 2021-10-29
fix: remove obsolete 'dexOptions' gradle element

Bambooin:
./gradlew app:assembleFdroidRelease

> Configure project :app
WARNING:DSL element 'dexOptions' is obsolete and should be removed.
It will be removed in version 8.0 of the Android Gradle plugin.
Using it has no effect, and the AndroidGradle plugin optimizes dexing automatically.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -8,11 +8,6 @@ apply plugin: "com.starter.easylauncher"
 android {
     compileSdkVersion 29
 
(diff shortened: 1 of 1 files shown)


EmmanuelMess (approved):
This complies so I suppose its OK.

VishalNehra (approved):

PR #3008, opened 2021-11-11
Proper Intent and Uri for all file access permission request

### LINKED ISSUE #3012: Reason for requesting "All files access" is not mentioned
**Describe the bug**
Recently I upgraded the Amaze to v3.6.7. After that when I opened the app it asked me for the "All files access" permission out of the blue and it was mentioned that the app would *not* function without it. Screenshot of the dialog:

image

Honestly, my knee-jerk reaction is that the dialog had a threatening tone. While the intention would've been good, the words are failing to convey it. It would be nice if the request was put in a more polite way.

One other thing that totally confused: Why does the app even need the "All files access" permission mandatorily out of the blue? I've been using the app on an Android 11 device for all this time without much issue. So, I couldn't really understand what caused the app to request for the "All files access". It would be wonderful if someone could explain the reasoning behind this. Also, if possible do evaluate if this could be made optional. I for one haven't observed any glaring issues while using the app when it didn't have the "All files access" permission.

 - Device: OnePlus Nord
 - OS: Android 11
 - Rooted: No
 - Version: 3.6.7

TranceLove:
## Description
So the prompt dialog would take user directly to the switch instead.

Reference: https://stackoverflow.com/a/65876928
Fixes #3012 

#### Manual tests
- [x] Done  
  
- Device: Oneplus 2
- OS: LineageOS 18 (11)
Upon tapping grant on all file access permission request dialog, user should be taken to the file access grant toggle switch screen, instead of getting a list of apps on user's device having access to all files on device.

#### Build tasks success  

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/ui/activities/superclasses/PermissionsActivity.java
+++ b/app/src/main/java/com/amaze/filemanager/ui/activities/superclasses/PermissionsActivity.java
@@ -197,7 +197,9 @@ public class PermissionsActivity extends ThemedActivity
                 Utils.disableScreenRotation(this);
                 permissionCallbacks[ALL_FILES_PERMISSION] = onPermissionGranted;
                 try {
-                  Intent intent = new Intent(Settings.ACTION_MANAGE_ALL_FILES_ACCESS_PERMISSION);
+                  Intent intent =
+                      new Intent(Settings.ACTION_MANAGE_APP_ALL_FILES_ACCESS_PERMISSION)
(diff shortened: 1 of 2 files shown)


namedme9 (approved):
Access to all 9

VishnuSanal:
Spotless and gradle build fails (maybe because it needs the changes from #3039) Also, parts of code is already on the target branch - should this be rebased?

VishnuSanal (approved):

VishalNehra (approved):

PR #3020, opened 2021-11-14
Explicit MIME type declaration on querying available packages for opening files

TranceLove:
## Description
As noted by Android Studio on the use of `android.permission.QUERY_ALL_PACKAGES` permission, use `<queries>` with explicitly defined action and MIME type instead. This looks safer (minimal permission principle), and also reduces possibility of unrelated apps from popping up in the list of available apps.

Additionally:
- as noted in https://stackoverflow.com/a/55071463, change SQLITE MIME type to `application/vnd.sqlite3`
- added explicit `android:exported` attribute to `<activity>`s in AndroidManifest.xml where the activity is not necessary to export
- added back `rar` to `FileUtils.COMPRESSED_FILE_EXTENSIONS` to fix problem that opening a RAR file using Amaze does not load up the `CompressedExplorerFragment`

#### Manual tests
- [x] Done  
  
Device: Pixel 2 emulator
OS: Android 11
Device: Fairphone 3
OS: LineageOS 16.0 (9.0)
Device: Pixel 2 emulator
OS: Android 6.0

- Tapping files (expect those MIME type not explicitly declared in the manifest) should never be affected - user should be given a list of available apps on device for the specified file
- Tapping open a RAR file from Amaze should open `CompressedExplorerFragment` for viewing contents inside the archive

#### Build tasks success  


TranceLove:
Pending #3022 before continuing. Will add queries for commonly used document types.

VishalNehra:
I don't prefer this. We have an inbuilt file opener. we need to query any and all mime types.

TranceLove:
> I don't prefer this. We have an inbuilt file opener. we need to query any and all mime types.

Then use a wildcard. It is still the preferred way to declare MIME type query to open files as noted by Android Studio hints.

VishalNehra:
@TranceLove It's preferable in other usecases, for us mime type is basically all, so defining a subset is gonna hurt us only. I think we can ignore this warning by ide. 
One usecase that comes to mind is, if a user choose 'other' in 'open with' option while opening file... He should see all possible intent filters in system, but defining a query here will let him see just a subset.

VishalNehra:
From play store:
```
Important dates for QUERY_ALL_PACKAGES permission declaration
If your app does not require use of the QUERY_ALL_PACKAGES permission, you must remove the permission from your app manifest. For apps that require this permission, you can now submit a declaration form in Play Console.

You'll need to provide:

A description of the core feature in your app that requires this permission
A short video showing the core feature in your app that requires this permission
To prepare for the questions you'll be required to answer, review this [Help Center](https://support.google.com/googleplay/android-developer/answer/9214102) article.

Starting July 12, you will not be able to submit new apps or app updates if the declaration is missing. You have until July 12 to submit the declaration, or remove the permission from your app manifest. Apps that fail to meet the [policy requirements](https://support.google.com/googleplay/android-developer/answer/9888170#package-vis) or do not submit the declaration form may be removed from Google Play starting July 12.
```
I'll be declaring because as per the policy, file managers should be allowed to query all packages, but just in case it doesn't go through we should have this PR as a backup. Thanks for working on this @TranceLove

Bladfool (approved):

Bladfool (approved):

EmmanuelMess:
@VishalNehra @TranceLove whats the state of this?

PR #2789, opened 2021-09-10
Small fixes to TabFragment

EmmanuelMess:
#### Automatic tests

- [ ] Added test cases
  
#### Manual tests
- [ ] Done  
  

#### Build tasks success  

Successfully running following tasks on local:
- [x] `./gradlew assembledebug`
- [x] `./gradlew spotlessCheck`

PR #3039, opened 2021-11-24
Fix failing build

VishnuSanal:
#### Manual tests
- [x] Done  
 
#### Build tasks success  

#### Related PR  
Related to PR #2789

### DIFF
--- a/.github/workflows/android-main.yml
+++ b/.github/workflows/android-main.yml
@@ -48,33 +48,3 @@ jobs:
         run: |
           export CODACY_PROJECT_TOKEN=${{ secrets.CODACY_TOKEN }}
           bash <(curl -Ls https://coverage.codacy.com/get.sh)
-  build_release_fdroid:
-    name: FDroid release build
-    runs-on: ubuntu-20.04
-    steps:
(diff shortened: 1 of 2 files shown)


EmmanuelMess (dismissed):

EmmanuelMess:
What's the rationale behind removing the builds?

VishnuSanal:
> What's the rationale behind removing the builds?

It was failing all these while because we were building release flavours. Also, as we discussed on Telegram, release should build fine if debug builds.

EmmanuelMess:
> It was failing all these while because we were building release flavours.

I still think building release flavour is indicative of how well releases are, for example, playRelease alerts to a lot of unused strings, also, the debug keystore should work out of the box for amaze, so that play release can be easily tested.

VishnuSanal:
> > It was failing all these while because we were building release flavours.
> 
> I still think building release flavour is indicative of how well releases are, for example, playRelease alerts to a lot of unused strings, also, the debug keystore should work out of the box for amaze, so that play release can be easily tested.

Agreed, but, release build requires keystore etc & we can't expose them.

PR #3071, opened 2021-12-17
Clear data packages if not empty when adding first datapoint

### LINKED ISSUE #1793: Crash not first datapoint
**Describe the bug**
crash when extracting zip file

**To Reproduce**
1. select a zip file and a dialog appears
2. tap 'veiw'
3. tap the zip file again
4. tap 'EXTRACT'
5. tap the file again and a dialog 'Please Wait' appears
6. keep tapping the file for several times

**Screenshots**
Screen Shot

 - Device: emulator
 - OS: Android 4.4.4
 - Rooted: No
 - Version: 3.3.2

**Additional context**

Stack Trace:

E/AndroidRuntime: FATAL EXCEPTION: AsyncTask #4
                                                   Process: com.amaze.filemanager, PID: 20416
                                                   java.lang.RuntimeException: An error occured while executing doInBackground()
                                                       at android.os.AsyncTask$3.done(AsyncTask.java:300)
                                                       at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:355)
                                                       at java.util.concurrent.FutureTask.setException(FutureTask.java:222)
                                                       at java.util.concurrent.FutureTask.run(FutureTask.java:242)
                                                       at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:231)
                                                       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
                                                       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
                                                       at java.lang.Thread.run(Thread.java:841)
                                                    Caused by: java.lang.IllegalStateException: This is not the first datapoint!
                                                       at com.amaze.filemanager.asynchronous.services.AbstractProgressiveService.addFirstDatapoint(AbstractProgressiveService.java:212)
                                                       at com.amaze.filemanager.asynchronous.services.ExtractService$DoWork$1.onStart(ExtractService.java:257)
                                                       at com.amaze.filemanager.filesystem.compressed.extractcontents.helpers.ZipExtractor.extractWithFilter(ZipExtractor.java:70)
                                                       at com.amaze.filemanager.filesystem.compressed.extractcontents.Extractor.extractFiles(Extractor.java:57)
                                                       at com.amaze.filemanager.asynchronous.services.ExtractService$DoWork.doInBackground(ExtractService.java:286)
                                                       at com.amaze.filemanager.asynchronous.services.ExtractService$DoWork.doInBackground(ExtractService.java:208)
                                                       at android.os.AsyncTask$2.call(AsyncTask.java:288)
                                                       at

VishalNehra:
#### Issue tracker   

Fixes #1793 

#### Automatic tests

- [ ] Added test cases
  
#### Manual tests
- [x] Done  
  

#### Build tasks success  


app/src/main/java/com/amaze/filemanager/asynchronous/services/AbstractProgressiveService.java
```diff
@@ -240,14 +241,16 @@ private void publishCompletedResult(int id1) {
   }
 
   protected void addFirstDatapoint(String name, int amountOfFiles, long totalBytes, boolean move) {
-    if (!getDataPackages().isEmpty()) {
-      throw new IllegalStateException("This is not the first datapoint!");
+    synchronized (getDataPackages()) {
+      if (!getDataPackages().isEmpty()) {
+        Log.w(getClass().getSimpleName(), "Warning: This is not the first datapoint!");
```
EmmanuelMess, line 246: This should be error.

app/src/main/java/com/amaze/filemanager/asynchronous/services/AbstractProgressiveService.java
```diff
@@ -240,14 +241,16 @@ private void publishCompletedResult(int id1) {
   }
 
   protected void addFirstDatapoint(String name, int amountOfFiles, long totalBytes, boolean move) {
-    if (!getDataPackages().isEmpty()) {
-      throw new IllegalStateException("This is not the first datapoint!");
+    synchronized (getDataPackages()) {
```
EmmanuelMess, line 244: Why the synchronized? Why isn't this using an actual lock if its needed? Or a concurrent list?

app/src/main/java/com/amaze/filemanager/asynchronous/services/AbstractProgressiveService.java
EmmanuelMess, line 246: This should be error.

EmmanuelMess (changes requested):

app/src/main/java/com/amaze/filemanager/asynchronous/services/AbstractProgressiveService.java
VishalNehra: I doubt even concurrent list can't address the issue here. If line 247 is executed and before putting first data package in 252 it's possible someone else can add another data package again. Hence the synchronized block. I'm not sure what do you mean by the 'actual lock'

EmmanuelMess: Every usage of the list should be syncronized for this to work.

EmmanuelMess: Syncronization on objects is currently not enforced either statically or in runtime, that's why I proposed using an actual lock or the concurrent list.

app/src/main/java/com/amaze/filemanager/asynchronous/services/AbstractProgressiveService.java
```diff
@@ -240,19 +242,21 @@ private void publishCompletedResult(int id1) {
...
-    DatapointParcelable intent1 =
-        DatapointParcelable.Companion.buildDatapointParcelable(
-            name, amountOfFiles, totalBytes, move);
-    putDataPackage(intent1);
   }
 
   protected void addDatapoint(DatapointParcelable datapoint) {
     if (getDataPackages().isEmpty()) {
-      throw new IllegalStateException("This is the first datapoint!");
+      Log.e(getClass().getSimpleName(), "Warning: This is the first datapoint!");
```
EmmanuelMess, line 259: Why the "Warning: "? This is already marked as error in logs.

EmmanuelMess:
Please remind me the thread structure for calls to "addDatapoint", is it called always from the same thread? is the list read only from the ui thread for showing the notification?

app/src/main/java/com/amaze/filemanager/asynchronous/services/AbstractProgressiveService.java
VishalNehra: Concurrent list will not help here because after clearing the list we want to add new element to it, and before reaching the line of code we want to avoid someone else to add to the list first (in `addDatapoint`). 
I do agree that we need synchronization on every usage, so i'll be putting synchronized on both `addFirstDatapoint` and `addDatapoint`. 

All this said, the real culprit here was `clearDatapackages` which wasn't called after the extract service's onDestroy, so the list already had elements on subsequent calls to extract service. I've added clear datapackages in abstract service's onDestroy. This race condition should almost never happen now, but regardless, I'll put synchronized on both  the methods.

VishalNehra:
yes, the list is accessed through ui thread, but can be written by multiple threads (there's a thread pool in service watcher, also the first datapoint is published by a background thread in the service itself), but notification data is published directly without using these, we use these datapoints for showing progress in form of graphs in the process viewer fragment.

app/src/main/java/com/amaze/filemanager/asynchronous/services/AbstractProgressiveService.java
EmmanuelMess: 1.`onDestroy` will not always be called, regardless, why does the list get presisted after the `onDestroy` is called?
2. If adding `clearDatapackages` to `onDestroy` fixes the issue and then syncronization is not needed, syncronization should be removed.
3. Otherwise you have to make sure the subclass doesn't do things with the list while it is synchronized in `AbstractProgressiveService` which could mean moving the list up to `AbstractProgressiveService` and synchronizing *on the list* not on the class object.

VishalNehra: 1. I'm not sure, but this is the only logical explanation behind this crash. Also,  clearing it in onDestroy doesn't harm either.
2. Point 1 as well as there *might be* a small chance a race condition can occur, again, I don't see any harm in adding synchronized on these methods, because clearly the methods are adding the first data point, we wouldn't want a race condition that during it cleared the existing list and when it's again adding the first one, someone else doesn't modify the list.
3. Making the list synchronized isn't really necessary. We're simply putting new datapoints in the list, it's just when the list is being created for the first time that requires these changes. 
Again, we can get away without putting synchronized on these blocks but I don't see any harm in doing so.

EmmanuelMess: It doesn't seem a good idea to add things that are redundant, specially when related to threading.

EmmanuelMess (approved):

PR #3074, opened 2021-12-18
Fix back button not responding when browsing at second level of archive

TranceLove:
## Description
There was a bug in CompressedExplorerFragment.goBack() that prevent browsing back to the first level of archive. This change fixes this, by defaulting parent path to "" if no parent path can be derived.

#### Manual tests
- [x] Done  
  
- Device: Pixel 2 emulator
- OS: Android 11

#### Build tasks success  


VishalNehra:
@TranceLove Great! Thanks for fixing.

VishnuSanal (approved):

EmmanuelMess (approved):

EmmanuelMess:
(Sorry for closing the button moved while I was clicking)

PR #3119, opened 2022-01-11
Limit TimeZone setting to individual class instead of build script

### LINKED ISSUE #2979: Test always fails in local env
Running `:app:testPlayDebugUnitTest` always fails with:

```
com.amaze.filemanager.asynchronous.asynctasks.compress.EncryptedRarHelperTaskTest > testRoot[16] FAILED
    java.lang.Exception at AndroidTestEnvironment.java:470
        Caused by: java.lang.AssertionError at Assert.java:89
```

TranceLove:
## Description
So tests can be run without using gradle.

Was tried to do `TimeZone.setDefault()` at `@BeforeClass` `@AfterClass` static methods, but didn't work - TimeZone was reset in the middle of nowhere, hence have to put them to `@Before` and `@After` instead.

#### Issue tracker   
Fixes #2979.

#### Build tasks success  


VishalNehra:
Timezone is also being set through github actions, that should be redundant now right?

EmmanuelMess (approved):

TranceLove:
@VishalNehra Let's see it.

VishalNehra (approved):

EmmanuelMess (approved):

PR #3140, opened 2022-01-22
Overhauls the settings screen

### LINKED ISSUE #2968: Add different panes to preferences
Add different panes to preferences activity. 

-----

This now has a bounty, [here](https://app.bountysource.com/issues/104727912-add-different-panes-to-preferences), bounty expires in one month from 2022-01-13, if you have been working, please alert us so that we can extend it.

Please claim after a PR fixing this issue is accepted.
Remember that you can also back the issue adding to the bounty!

nicbn:
## Description

Video demonstration: https://www.youtube.com/watch?v=xo8mEtx8GCg

#### Issue tracker   

Closes #2968 
  
#### Manual tests

- [x] Done  

Tested on API 30 and API 16

#### Build tasks success  


EmmanuelMess:
Do not move stuff to utils, the name is meaningless.

EmmanuelMess:
Please beware of tests that might fail after the name changes.

nicbn:
Should I move the constants to the base package? I don't think it makes sense to have these constants at the fragments package, as most of their usage doesn't concern fragments, and is related to SharedPreferences keys instead.

EmmanuelMess:
> Should I move the constants to the base package? 

No.

> I don't think it makes sense to have these constants at the fragments package, as most of their usage doesn't concern fragments, and is related to SharedPreferences keys instead.

I would move it to where most sense it makes for to be at, but I am not quite sure that place exists here, feel free to create one if needed.

> Or should I revert these changes and PR a refactor separately later?

As you wish.

app/src/main/java/com/amaze/filemanager/ui/activities/superclasses/ThemedActivity.java
```diff
@@ -85,27 +82,6 @@ public void onCreate(Bundle savedInstanceState) {
     setTheme();
   }
 
-  @Override
```
nicbn, line 88: This was removed because it was causing state to be lost when orientation changed, or activity recreated for any other reason.

Seems like nothing else changed with this change, app runs fine if the theme is not set or set to default.

EmmanuelMess: What? That annotation doesn't really do anything AFAIK, the override is created nontheless.

nicbn: I meant the methods below the annotation, onConfigurationChanged and restartPC. This is just the first line

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/PrefsFragment.kt
```diff
@@ -0,0 +1,88 @@
...
+                    startActivity(Intent(activity, AboutActivity::class.java))
+                    false
+                }
+
+        findPreference<Preference>("feedback")!!
+                .onPreferenceClickListener = Preference.OnPreferenceClickListener {
+            val emailIntent = Utils.buildEmailIntent(null, Utils.EMAIL_SUPPORT)
+
+            val activities = activity.packageManager.queryIntentActivities(
+                    emailIntent, PackageManager.MATCH_DEFAULT_ONLY)
```
nicbn, line 73: `spotlessCheck` fails here

```
Step 'ktlint' found problem in 'app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/PrefsFragment.kt':
Error on line: 73, column: 38
Exceeded max line length (100)
```

I tried even making this whole callback into a comment and it didn't work. No idea why this is failing here. Any ideas?

VishnuSanal: > 



> `spotlessCheck` fails here
> 
> ```
> Step 'ktlint' found problem in 'app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/PrefsFragment.kt':
> Error on line: 73, column: 38
> Exceeded max line length (100)
> ```
> 
> I tried even making this whole callback into a comment and it didn't work. No idea why this is failing here. Any ideas?

It's exceeding the max line length - try renaming the variables to something shorter & see if that works.

nicbn: Problem solved, the issue was another line down below that was too large. Weird that the number was completely off.

app/src/main/java/com/amaze/filemanager/ui/activities/superclasses/ThemedActivity.java
EmmanuelMess: > Seems like nothing else changed with this change, app runs fine if the theme is not set or set to default.

Does it still correctly set the theme immediately (in the settings screen) on theme change?

app/src/main/java/com/amaze/filemanager/ui/activities/PreferencesActivity.kt
```diff
@@ -0,0 +1,164 @@
...
+import com.amaze.filemanager.ui.fragments.preference_fragments.PrefsFragment
+import com.amaze.filemanager.ui.theme.AppTheme
+import com.amaze.filemanager.utils.PreferenceUtils
+import com.amaze.filemanager.utils.Utils
+import com.readystatesoftware.systembartint.SystemBarTintManager
+import java.io.File
+
+class PreferencesActivity : ThemedActivity(), FolderChooserDialog.FolderCallback {
+    val layout: View
+        get() = _layout!!
```
EmmanuelMess, line 49: No `!!` please, change `_layout` to `lateinit var` and unify with `layout`

app/src/main/java/com/amaze/filemanager/ui/activities/PreferencesActivity.kt
```diff
@@ -0,0 +1,164 @@
...
+        }
+    }
+
+    override fun onBackPressed() {
+        if (supportFragmentManager.backStackEntryCount > 0) {
+            supportFragmentManager.popBackStack()
+            return
+        }
+
+        val x = Intent(this, MainActivity::class.java)
```
EmmanuelMess, line 84: No single letter variable name. Use `intent` or similar here.

app/src/main/java/com/amaze/filemanager/ui/activities/PreferencesActivity.kt
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    fun invalidateNavBar() {
+        val primaryColor = ColorPreferenceHelper
+            .getPrimary(currentColorPreference, MainActivity.currentTab)
+        if (Build.VERSION.SDK_INT == 20 || Build.VERSION.SDK_INT == 19) {
+            val tintManager = SystemBarTintManager(this)
+            tintManager.isStatusBarTintEnabled = true
+            tintManager.setStatusBarTintColor(primaryColor)
+            val p = findViewById<View>(R.id.activity_preferences).layoutParams
```
EmmanuelMess, line 132: No single letter variable name. Use `layoutParams` or similar here.

app/src/main/java/com/amaze/filemanager/ui/dialogs/ColorPickerDialog.java
```diff
@@ -284,6 +284,18 @@ public void onDialogClosed(boolean positiveResult) {
     }
   }
 
+  public static int getTitle(int i) {
```
EmmanuelMess, line 287: Use `@StringRes` annotation for this method.

EmmanuelMess, line 287: No single letter variables.

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/AppearancePrefsFragment.kt
```diff
@@ -0,0 +1,132 @@
...
+import android.os.Bundle
+import androidx.preference.Preference
+import com.afollestad.materialdialogs.MaterialDialog
+import com.amaze.filemanager.R
+import com.amaze.filemanager.ui.theme.AppTheme
+
+class AppearancePrefsFragment : BasePrefsFragment() {
+    override val title = R.string.appearance
+
+    private var gridColumnPref: Preference? = null
```
EmmanuelMess, line 33: This seems like it should be `lateinit var`

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/AppearancePrefsFragment.kt
```diff
@@ -0,0 +1,132 @@
...
+                .title(R.string.theme)
+                .build()
+                .show()
+
+            true
+        }
+
+        val navPref = findPreference<Preference>(
+            PreferencesConstants.PREFERENCE_COLORED_NAVIGATION
+        )!!
```
EmmanuelMess, line 69: No `!!` please

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/AppearancePrefsFragment.kt
```diff
@@ -0,0 +1,132 @@
...
+        colorPrefs.onPreferenceClickListener = Preference.OnPreferenceClickListener {
+            activity.pushFragment(ColorPrefsFragment())
+
+            true
+        }
+
+        val gridColumnItems = resources.getStringArray(R.array.columns)
+        gridColumnPref = findPreference(PreferencesConstants.PREFERENCE_GRID_COLUMNS)
+        updateGridColumnSummary()
+        gridColumnPref!!.onPreferenceClickListener = Preference.OnPreferenceClickListener {
```
EmmanuelMess, line 91: No `!!`.

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/BehaviorPrefsFragment.kt
```diff
@@ -0,0 +1,74 @@
...
+import com.amaze.filemanager.ui.dialogs.OpenFileDialogFragment.Companion.clearPreferences
+import java.io.File
+
+class BehaviorPrefsFragment : BasePrefsFragment(), FolderChooserDialog.FolderCallback {
+    override val title = R.string.behavior
+
+    override fun onCreatePreferences(savedInstanceState: Bundle?, rootKey: String?) {
+        setPreferencesFromResource(R.xml.behavior_prefs, rootKey)
+
+        findPreference<Preference>("clear_open_file")!!.onPreferenceClickListener =
```
EmmanuelMess, line 38: No `!!`

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/BookmarksPrefsFragment.kt
```diff
@@ -36,74 +34,65 @@ import com.amaze.filemanager.database.UtilsHandler
...
+
+    companion object {
+        private val dataUtils = DataUtils.getInstance()!!
+    }
 
-    private var sharedPrefs: SharedPreferences? = null
     private val position: MutableMap<Preference, Int> = HashMap()
-    private var dataUtils: DataUtils? = null
-    private var utilsHandler: UtilsHandler? = null
+    private var bookmarksList: PreferenceCategory? = null
```
EmmanuelMess, line 50: This should probably be `lateinit var`

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/BookmarksPrefsFragment.kt
```diff
@@ -36,74 +34,65 @@ import com.amaze.filemanager.database.UtilsHandler
...
-            preferenceScreen.addPreference(p)
-        }
-    }
+        setPreferencesFromResource(R.xml.bookmarks_prefs, rootKey)
 
-    override fun onResume() {
-        super.onResume()
-        onCreate(null)
-    }
+        findPreference<Preference>("add_bookmarks")!!.onPreferenceClickListener =
```
EmmanuelMess, line 63: No `!!`.

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/ColorPrefsFragment.kt
```diff
@@ -0,0 +1,161 @@
...
+    override fun onDisplayPreferenceDialog(preference: Preference) {
+        if (preference.key == PreferencesConstants.PRESELECTED_CONFIGS) {
+            showPreselectedColorsConfigDialog()
+        } else {
+            super.onDisplayPreferenceDialog(preference)
+        }
+    }
+
+    private fun showPreselectedColorsConfigDialog() {
+        val dialog = ColorPickerDialog.newInstance(
```
EmmanuelMess, line 82: Name shadowing, maybe should be corrected. @TranceLove ?

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/PrefsFragment.kt
```diff
@@ -0,0 +1,94 @@
...
+import com.amaze.filemanager.ui.activities.AboutActivity
+import com.amaze.filemanager.utils.Utils
+
+class PrefsFragment : BasePrefsFragment() {
+    override val title = R.string.setting
+
+    override fun onCreatePreferences(savedInstanceState: Bundle?, rootKey: String?) {
+        setPreferencesFromResource(R.xml.preferences, rootKey)
+
+        findPreference<Preference>("appearance")!!.onPreferenceClickListener =
```
EmmanuelMess, line 38: No `!!`. All the file

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/QuickAccessesPrefsFragment.kt
```diff
@@ -20,52 +20,41 @@
...
+        val currentValue = TinyDB.getBooleanArray(activity.prefs, KEY, DEFAULT)!!
 
-        // Load the preferences from an XML resource
-        addPreferencesFromResource(R.xml.fastaccess_prefs)
-        preferences = PreferenceManager.getDefaultSharedPreferences(activity)
-        currentValue = TinyDB.getBooleanArray(preferences!!, KEY, DEFAULT)
-        for (i in 0 until preferenceScreen.preferenceCount) {
-            preferenceScreen.getPreference(i).onPreferenceClickListener = this
+        val onChange = Preference.OnPreferenceClickListener {
+            currentValue[prefPos[it.key]!!] = (it as SwitchPreference).isChecked
```
EmmanuelMess, line 50: No `!!` probably a let is better.

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/QuickAccessesPrefsFragment.kt
```diff
@@ -20,52 +20,41 @@
...
+
+            true
         }
-    }
 
-    override fun onPreferenceClick(preference: Preference): Boolean {
-        currentValue!![prefPos[preference.key]!!] = (preference as SwitchPreference).isChecked
-        TinyDB.putBooleanArray(preferences!!, KEY, currentValue!!)
-        return true
+        for (k in KEYS) {
```
EmmanuelMess, line 56: No single letter variable names, use `key`.

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/SecurityPrefsFragment.kt
```diff
@@ -0,0 +1,212 @@
...
+                resources.getString(R.string.authenticate_password),
+                decryptedPassword,
+                true
+            ) { _, _ -> }
+            masterPasswordDialogBuilder.theme(
+                activity.utilsProvider.appTheme.getMaterialDialogTheme(requireContext())
+            )
+            masterPasswordDialogBuilder.positiveText(resources.getString(R.string.ok))
+            masterPasswordDialogBuilder.negativeText(resources.getString(R.string.cancel))
+            masterPasswordDialogBuilder.positiveColor((getActivity() as ThemedActivity?)!!.accent)
```
EmmanuelMess, line 164: Should probably be `requireActivity() as ThemedActivity`

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/UiPrefsFragment.kt
```diff
@@ -0,0 +1,89 @@
...
+
+class UiPrefsFragment : BasePrefsFragment() {
+    override val title = R.string.ui
+
+    private var dragAndDropPref: Preference? = null
+
+    override fun onCreatePreferences(savedInstanceState: Bundle?, rootKey: String?) {
+        setPreferencesFromResource(R.xml.ui_prefs, rootKey)
+
+        findPreference<Preference>("sidebar_bookmarks")!!.onPreferenceClickListener =
```
EmmanuelMess, line 36: `!!`

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/UiPrefsFragment.kt
```diff
@@ -0,0 +1,89 @@
...
+
+import android.os.Bundle
+import androidx.preference.Preference
+import com.afollestad.materialdialogs.MaterialDialog
+import com.amaze.filemanager.R
+
+class UiPrefsFragment : BasePrefsFragment() {
+    override val title = R.string.ui
+
+    private var dragAndDropPref: Preference? = null
```
EmmanuelMess, line 31: `lateinit var`

app/src/main/java/com/amaze/filemanager/ui/views/preference/PathSwitchPreference.kt
```diff
@@ -33,32 +32,21 @@ class PathSwitchPreference(context: Context?) : Preference(context) {
     var lastItemClicked = -1
         private set
 
+    var onEdit: (PathSwitchPreference) -> Unit = {}
```
EmmanuelMess, line 35: Shouldn't these be part of the constructor?

EmmanuelMess (changes requested):
Needs a few changes.

EmmanuelMess:
There is a detail, why does it take so long between the click and the opening of the section on my Nokia? It should be almost instant, but it takes a few frames, the "touch" animation almost ends before the screen opens.

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/AppearancePrefsFragment.kt
nicbn: This is assigned via `findPreference` so it must either be `Preference?` or `!!` must be used on assignment

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/BookmarksPrefsFragment.kt
nicbn: Same as above: `findPreference` returns `Preference?`

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/UiPrefsFragment.kt
nicbn: Same as above: `findPreference` returns `Preference?`

nicbn:
That's because of the transition I was using. Changed it to a fade transition which feels way more responsive.

This is not what is used in Android settings, which uses the "open" transition, but I found it a bit distracting.

EmmanuelMess:
> That's because of the transition I was using. Changed it to a fade transition which feels way more responsive.
> 
> This is not what is used in Android settings, which uses the "open" transition, but I found it a bit distracting.

Its better now.

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/AppearancePrefsFragment.kt
EmmanuelMess: Makes sense, then leave it as is, and use "?" on every usage.

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/BookmarksPrefsFragment.kt
EmmanuelMess: Makes sense, then leave it as is, and use "?" on every usage.

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/UiPrefsFragment.kt
EmmanuelMess: Makes sense, then leave it as is, and use "?" on every usage.

EmmanuelMess (approved):
Thanks!

EmmanuelMess:
@nicbn sorry for the delay, Codacy had an issue with their services and the Codacy status wasn't getting through to the PR, could you fix the Codacy issues so we can merge please?

nicbn:
Done

EmmanuelMess (dismissed):

app/src/main/java/com/amaze/filemanager/ui/activities/superclasses/PreferenceActivity.java
```diff
@@ -57,9 +57,9 @@
 
   @Override
   public void onCreate(final Bundle savedInstanceState) {
-    super.onCreate(savedInstanceState);
-
     sharedPrefs = PreferenceManager.getDefaultSharedPreferences(this);
+
+    super.onCreate(savedInstanceState);
```
VishnuSanal, line 62: Why?!

app/src/main/java/com/amaze/filemanager/ui/activities/superclasses/ThemedActivity.java
VishnuSanal: > Seems like nothing else changed with this change, app runs fine if the theme is not set or set to default.

Changing the system theme whilst using the app (with theme set to follow system) doesn't work as expected. The app theme stays intact until the app is restated.

> This was removed because it was causing state to be lost when orientation changed, or activity recreated for any other reason.

Can you revert this change and handle the above condition in some other way?

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/BasePrefsFragment.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2014-2021 Arpit Khurana <arpitkh96@gmail.com>, Vishal Nehra <vishalmeham2@gmail.com>,
+ * Copyright (C) 2014-2020 Arpit Khurana <arpitkh96@gmail.com>, Vishal Nehra <vishalmeham2@gmail.com>,
```
VishnuSanal, line 2: Was this changed by spotless?

VishnuSanal (changes requested):
@nicbn Excellent work!

- Please address the comments
- Also, where are the new icons from? Did you create them on your own or imported? If they're imported, please link us to the source.

nicbn:
> Also, where are the new icons from? Did you create them on your own or imported? If they're imported, please link us to the source.

They are clip arts built into Android Studio, IIRC they are from https://fonts.google.com/icons

app/src/main/java/com/amaze/filemanager/ui/activities/superclasses/PreferenceActivity.java
nicbn: Because fragments are created before the `super` call returns, so we must initialize `sharedPrefs` before the `super` call otherwise it cannot be used by fragments

app/src/main/java/com/amaze/filemanager/ui/activities/superclasses/ThemedActivity.java
nicbn: In this case, I will try to handle it more gracefully than before. The way it was if the theme was set to follow the system, every configuration change (such as orientation) would cause the activity to be restarted (not recreated, restarted as in a new activity is created from the ground up), if I understand correctly.

This is bad because it messes up with saved bundle badly, which can cause lots of frustration when you are trying to make fragments work and they just reset whenever orientation changes. There are some hacky solutions which were being used before but it is not elegant at all.

I believe the best alternative is instead to call `Activity` `recreate` whenever the system theme changes and we should use the same theme as the system. This causes a screen flicker, but still IMO is a better solution.

app/src/main/java/com/amaze/filemanager/ui/fragments/preference_fragments/BasePrefsFragment.kt
nicbn: Not sure, I will revert

app/src/main/java/com/amaze/filemanager/ui/activities/superclasses/ThemedActivity.java
```diff
@@ -136,6 +115,26 @@ public void initStatusBarResources(View parentView) {
     }
   }
 
+  @Override
```
nicbn, line 118: See here for the new logic on system night theme

app/src/main/java/com/amaze/filemanager/ui/activities/superclasses/ThemedActivity.java
VishnuSanal: Great! I should've done that in the first place whilst implementing too 🙈

app/src/main/java/com/amaze/filemanager/ui/activities/superclasses/ThemedActivity.java
```diff
@@ -136,6 +115,27 @@ public void initStatusBarResources(View parentView) {
...
+    // System theme change
+    if (uiModeNight != newUiModeNight) {
+      uiModeNight = newUiModeNight;
+
+      if (getPrefs().getString(PreferencesConstants.FRAGMENT_THEME, "4").equals("4")) {
+        getUtilsProvider().getThemeManager().setAppTheme(AppTheme.getTheme(this, 4));
+        // Recreate activity, handling saved state
+        //
+        // Not smooth, but will only be called if the user changes the system theme, not
+        // the app theme.
```
VishnuSanal, line 133: I can't notice any change on a real device. So, +1 from my side.

VishnuSanal (approved):

VishalNehra:
What happen to 'send feedback' option? It used to option email app prior to this. Other changes seem really good.

nicbn:
It seems to be working the same in my device

EmmanuelMess (approved):

EmmanuelMess:
[Here](https://github.com/TeamAmaze/AmazeFileManager/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3ABounty) are some other bounties.

PR #3185, opened 2022-02-26
App backup: add support for apps that have multiple APK files

### LINKED ISSUE #3184: App Manager backup only copies base apk file and ignores split apk files
When you open the "App Manager" in Amaze and select an app downloaded from Play Store that is a split APK (consists of multiple APK file: one base APK file and one or more split APK files) and execute the function "Backup" then only the base apk file is copied to the `app_backup` folder on SD card.

This means the backup is incomplete as the other split APK files contain native libraries, localization and necessary resources like images for the device screen resolution.

jpstotz:
As described in #3184 the app backup function of Amaze does only copy the main APK file of an app even if it is a split APK that consists of multiple APK files.

This PR adds the necessary functionality to copy all APK files that belong to an app when the user makes an app backup using Amaze.

Fixes #3184

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/adapters/AppsRecyclerAdapter.kt
+++ b/app/src/main/java/com/amaze/filemanager/adapters/AppsRecyclerAdapter.kt
@@ -68,7 +68,6 @@ import com.amaze.filemanager.utils.AnimUtils.marqueeAfterDelay
 import com.amaze.filemanager.utils.Utils
 import com.amaze.filemanager.utils.safeLet
 import java.io.File
-import java.util.*
 import kotlin.collections.ArrayList
 import kotlin.math.roundToInt
 
@@ -425,17 +424,8 @@ class AppsRecyclerAdapter(
     }
 
     private fun popupBackup(appDataParcelable: AppDataParcelable) {
-        Toast.makeText(
-            fragment.context,
-            fragment.getString(R.string.copyingapk) +
-                Environment.getExternalStorageDirectory().path +
-                "/app_backup",
-            Toast.LENGTH_LONG
-        )
-            .show()
-        val f = File(appDataParcelable.path)
-        val ab =
-            ArrayList<HybridFileParcelable>()
+        val baseApkFile = File(appDataParcelable.path)
+        val filesToCopyList = ArrayList<HybridFileParcelable>()
         val dst = File(
             Environment.getExternalStorageDirectory()
                 .path + "/app_backup"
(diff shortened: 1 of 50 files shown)


jpstotz:
BTW: I just noticed that the current spotless configuration for Java sources (and may be Kotlin, too) is ineffective.

Executing `./gradle spotlessApply` does neither correct indentation nor remove unnecessary imports. It looks like multiple `target` definitions are not supported by the plugin. If I combine all target definitions into one it works: 

`target 'app/src/**/*.java', 'commons_compress_7z/src/**/*.java', 'file_operations/src/**/*.java'`

EmmanuelMess:
> I noticed this a while back and did raise it on our telegram group @EmmanuelMess Thanks for pointing the issue.

Wut?

EmmanuelMess:
Is this a work in progress or is it ready for review?

jpstotz:
@EmmanuelMess From my point of view this PR is ready for a review. The necessary changes were not so extensive.

EmmanuelMess:
Please apply this patch to remove unused strings: [0001-Removed-ununsed-strings.zip](https://github.com/TeamAmaze/AmazeFileManager/files/8153439/0001-Removed-ununsed-strings.zip)

jpstotz:
@EmmanuelMess Sorry I forgot to remove the now unused strings in all the languages. I applied you patch.

VishalNehra (dismissed):

EmmanuelMess:
How do I test this? Is there a known multi apk application?

app/src/main/java/com/amaze/filemanager/adapters/AppsRecyclerAdapter.kt
```diff
@@ -445,15 +435,40 @@ class AppsRecyclerAdapter(
...
+            }
+        }
+        intent.putParcelableArrayListExtra(CopyService.TAG_COPY_SOURCES, filesToCopyList)
         intent.putExtra(CopyService.TAG_COPY_TARGET, dst.path)
         intent.putExtra(CopyService.TAG_COPY_OPEN_MODE, 0)
+
+        val dest = Environment.getExternalStorageDirectory().path + "/app_backup"
+        Toast.makeText(
+                fragment.context,
+                fragment.getString(R.string.copyingapks, filesToCopyList.size, dest),
```
VishnuSanal, line 467: Cant this just be `dst.path`?

VishnuSanal (changes requested):
Awesome work!

Please see the comments.

Also, as mentioned by Emmanuel, please suggest an app so that we can test this too.

app/src/main/java/com/amaze/filemanager/adapters/AppsRecyclerAdapter.kt
jpstotz: @VishnuSanal You are correct, the `dest` variable is no longer necessary. In the original code the toast was right at the beginning of the `popupBackup` method, at that point `dst` was not yet defined. I will update that.

jpstotz:
@EmmanuelMess At the moment I would assume than more than 40% of all apps installed via Google Play Store should consist of multiple APK files. For example on my device the stable version of Amaze itself, installed via Google Play consists of 5 APK files (base/main, xxhdpi, arm64_v8a, de, en).

Other example apps that have multiple APK files are on my device [Plantnet](https://play.google.com/store/apps/details?id=org.plantnet), [Google Duo](https://play.google.com/store/apps/details?id=com.google.android.apps.tachyon), [Google Docs](https://play.google.com/store/apps/details?id=com.google.android.apps.docs.editors.docs).
GMail, Google Maps, Google Chrome are also good candidates that should consist of multiple APK files if the sued device is not running an very old Android version.

VishnuSanal (approved):

EmmanuelMess (approved):

PR #3190, opened 2022-02-27
Better error message for missing decompressors/extractors

### LINKED ISSUE #3142: Decompressor instance must not be null
## Exception
* __App Name:__ Panel de navegación de ejemplo
* __Package:__ com.amaze.filemanager
* __Version:__ 3.6.7
* __User Action:__ UI Error
* __OS:__ Linux
TCL/4049G/U3_3G:6.0/MRA58K/4049G_AVCO_V2.7:user/release-keys 6.0 - 23
* __Device:__ U3_3G
* __Model:__ 4049G
* __Product:__ 4049G
<details><summary><b>Crash log </b></summary><p>

```
java.lang.NullPointerException: getCompressorInstance(requireContext(),
this) must not be null
at
com.amaze.filemanager.ui.fragments.CompressedExplorerFragment.onViewCreated(CompressedExplorerFragment.kt:235)
at androidx.fragment.app.Fragment.performViewCreated(Fragment.java:2974)
at
androidx.fragment.app.FragmentStateManager.createView(FragmentStateManager.java:543)
at
androidx.fragment.app.FragmentStateManager.moveToExpectedState(FragmentStateManager.java:282)
at
androidx.fragment.app.FragmentManager.executeOpsTogether(FragmentManager.java:2177)
at
androidx.fragment.app.FragmentManager.removeRedundantOperationsAndExecute(FragmentManager.java:2088)
at
androidx.fragment.app.FragmentManager.execPendingActions(FragmentManager.java:1990)
at androidx.fragment.app.FragmentManager$5.run(FragmentManager.java:524)
at android.os.Handler.handleCallback(Handler.java:815)
at android.os.Handler.dispatchMessage(Handler.java:104)
at android.os.Looper.loop(Looper.java:207)
at android.app.ActivityThread.main(ActivityThread.java:5765)
at
com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:806)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:696)

```
</details>
<hr>

EmmanuelMess:
#### Issue tracker   

Fixes #3142

#### Automatic tests

- [ ] Added test cases
  
#### Manual tests
- [ ] Done  
  

#### Build tasks success  

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/filesystem/compressed/CompressedHelper.java
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/compressed/CompressedHelper.java
@@ -22,6 +22,7 @@ package com.amaze.filemanager.filesystem.compressed;
 
 import java.io.File;
 
+import com.amaze.filemanager.BuildConfig;
 import com.amaze.filemanager.file_operations.utils.UpdatePosition;
 import com.amaze.filemanager.filesystem.compressed.extractcontents.Extractor;
 import com.amaze.filemanager.filesystem.compressed.extractcontents.helpers.Bzip2Extractor;
@@ -49,10 +50,11 @@ import com.amaze.filemanager.filesystem.compressed.showcontents.helpers.ZipDecom
 import com.amaze.filemanager.utils.Utils;
 
 import android.content.Context;
+import android.util.Log;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 
-/** @author Emmanuel on 23/11/2017, at 17:46. */
 public abstract class CompressedHelper {
 
   /**
@@ -80,7 +82,10 @@ public abstract class CompressedHelper {
   public static final String fileExtensionGz = "gz";
   public static final String fileExtensionBzip2 = "bz2";
 
+  private static final String TAG = CompressedHelper.class.getSimpleName();
+
   /** To add compatibility with other compressed file types edit this method */
+  @Nullable
   public static Extractor getExtractorInstance(
       @NonNull Context context,
       @NonNull File file,
@@ -118,13 +123,18 @@ public abstract class CompressedHelper {
     } else if (isBzip2(type)) {
       extractor = new Bzip2Extractor(context, file.getPath(), outputPath, listener, updatePosition);
     } else {
-      return null;
+      if(BuildConfig.DEBUG) {
+        throw new IllegalArgumentException("The compressed file has no way of opening it: " + file);
+      }
+      Log.e(TAG, "The compressed file has no way of opening it: " + file);
+      extractor = null;
     }
 
     return extractor;
   }
 
   /** To add compatibility with other compressed file types edit this method */
+  @Nullable
   public static Decompressor getCompressorInstance(@NonNull Context context, @NonNull File file) {
     Decompressor decompressor;
     String type = getExtension(file.getPath());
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/amaze/filemanager/filesystem/compressed/CompressedHelper.java
```diff
@@ -118,13 +118,14 @@ public static Extractor getExtractorInstance(
     } else if (isBzip2(type)) {
       extractor = new Bzip2Extractor(context, file.getPath(), outputPath, listener, updatePosition);
     } else {
-      return null;
+      throw new IllegalArgumentException("The current compressed file has no way of opening it!" + file);
```
VishalNehra, line 121: The crash report points to getCompressorInstance method but the change is made in getExtractInstance.
Also to point out, throwing an illegal argument exception isn't user friendly, either throw a toast message saying 'operation failed'

VishalNehra (changes requested):

app/src/main/java/com/amaze/filemanager/filesystem/compressed/CompressedHelper.java
EmmanuelMess: > Also to point out, throwing an illegal argument exception isn't user friendly, either throw a toast message saying 'operation failed'

The problem is checking if the decompressor can decompress isn't this function's resposability, that resposability belies in `isFileExtractable`. If this crashes, it means someone forgot to add a case here or forgot to check if the compressed file is actually extractable.

EmmanuelMess: Also, we have no way to track down the issue if the error is "operation failed".

VishalNehra: Sorry I should rephrase. thowing illegal argument exception only isn't enough here. This'll still be a crash page for user.
I do understand your point, 'operation failed' doesn't give better understanding to us, that's why we log this crash wherever this method call is used and show a toast to user instead. 
In future I do have plans to extract all logs in a session to a log file and allow user to send us that logfile. That'll allow us to find the exact cause for this and other similar 'operation failed' scenarios around the app.

EmmanuelMess: Hmm, I'll make it a crash if DEBUG and a toast if not, but I'll have to see how to make it work.

app/src/main/res/values/strings.xml
```diff
@@ -745,5 +745,7 @@ You only need to do this once, until the next time you select a new location for
 
     <string name="quick_desc">Files recently accessed through Amaze</string>
     <string name="recent_desc">Recently created or modified files</string>
+    <string name="error_cannot_find_way_open">Could not find way to open requested file</string>
```
VishnuSanal, line 748: I think this can be `Could not find a way to open the requested file` :)

app/src/main/res/values/strings.xml
```diff
@@ -745,5 +745,7 @@ You only need to do this once, until the next time you select a new location for
...
+    <string name="error_cant_decompress_that_file">Can\'t extract or show that archive</string>
```
VishnuSanal, line 749: `that archive` can be replaced with `the requested archive` IMO

VishnuSanal (changes requested):
Awesome work, please see the comments.

app/src/main/res/values/strings.xml
EmmanuelMess: Is `the archive` ok? `requested` seems unnecessarily long

VishnuSanal: > Is `the archive` ok? `requested` seems unnecessarily long

How about using `Can't process the requested archive` and removing `extract or show`?

EmmanuelMess: Fixed!

VishnuSanal (dismissed):

VishnuSanal (approved):

TranceLove (approved):

PR #3204, opened 2022-03-05
Fix Android connected test compatibility with main code base

TranceLove:
By downgrading awaitility to 3.1.6, the last version we had been using. 4.x series depends on Java 8 which we shall avoid.


EmmanuelMess (approved):
All emulator tests pass on my PC

VishalNehra (approved):

PR #3182, opened 2022-02-24
AESCrypt support

TranceLove:
## Description

- Split PasswordUtil and SecretKeygen away from CryptUtil
- CryptUtil code refactoring to reduce duplication
- Add AESCrypt support. Code is Kotlintized with adaptations to Android platform

#### Issue tracker   
Fixes #2841

#### Automatic tests
- [x] Added test cases
  
#### Manual tests
- [x] Done  
  
Device:
- Pixel 5 emulator running Android 12
- Pixel emulator running Android 4.4

#### Build tasks success  
Successfully running following tasks on local:
- [x] `./gradlew assembledebug`
- [x] `./gradlew spotlessCheck`

Issue #3156, opened 2022-02-05
Files encryption "Warning" related suggestions


shuvashish76:
Current **Warning** screen👇🏿
![Screenshot_2022-02-05-12-58-54](https://user-images.githubusercontent.com/48226312/152632989-0e92b4ea-b481-46a9-91d9-af035dcc2464.jpeg)

Usually we never read such disclaimer/warning, we just click yes/ok/agree/accept. Though Amaze has another extra button for it called "Don't Inform Again" which makes it worse.

Suggestions: 
1. add a timer to current "Got it" button (check screen recording)
2. add a checkbox for "Don't inform again" to the bottom_left of the screen(check screenshot)
3. Replace "Don't inform again" with Exit

https://user-images.githubusercontent.com/48226312/152633297-67582d93-3747-44d3-a1ba-eb0437a4c6fb.mp4

source: [SecondScreen](https://github.com/farmerbb/SecondScreen) app

![Screenshot_2022-02-05-11-36-31](https://user-images.githubusercontent.com/48226312/152633334-6535542b-8cbb-4536-97ee-2bf7bff7f3b7.jpeg)

source: [AppManager](https://github.com/MuntashirAkon/AppManager) app

---
Now ppl just can't "_mark the checkbox-click Got it_" & Encrypt quickly.
We make sure users read the Disclaimer by adding 3 steps in total. (_though for some ppl 5sec is not enough to read full disclaimer it prevents users to skip the warning subconsciously/auto-mode_)

This is an important issue as ppl losing their personal files & IMO Amaze should provide a quick-hotfix as soon as possible till Amaze add #2841.

ThankYou🙏🏿

VishnuSanal:
> Usually we never read such disclaimer/warning, we just click yes/ok/agree/accept. Though Amaze has another extra button for it called "Don't Inform Again" which makes it worse.

> Suggestions:
> 
> 1. add a timer to current "Got it" button (check screen recording)
> 2. add a checkbox for "Don't inform again" to the bottom_left of the screen(check screenshot)
> 3. Replace "Don't inform again" with Exit

> Now ppl just can't "_mark the checkbox-click Got it_" & Encrypt quickly. We make sure users read the Disclaimer by adding 3 steps in total. (_though for some ppl 5sec is not enough to read full disclaimer it prevents users to skip the warning subconsciously/auto-mode_)

[...]

PR #3235, opened 2022-03-24
Fix at encrypt save as dialog, tapping the info button have no response

TranceLove:
Fix problem at #3182: https://github.com/TeamAmaze/AmazeFileManager/pull/3182#issuecomment-1077663402

Additionally, added a contentIsHtml flag at AlertDialog.show() as the aze encryption format description is composed of HTML.

#### Issue tracker   
Addresses problem at #3182

#### Automatic tests
- [x] Added test cases
  
#### Manual tests
- [x] Done  
  
- Device: Pixel 2 emulator
- OS: Android 11

#### Build tasks success  

#### Related PR  
Related to PR #3182

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -628,10 +628,12 @@
     <string name="encrypt_file_save_as">Save Encrypted File As&#8230;</string>
     <string name="encrypt_folder_save_as">Save Encrypted Folder As&#8230;</string>
     <string name="encrypt_option_use_aze">Use Amaze Encryption Format</string>
-    <string name="encrypt_option_use_aescrypt_title">About Amaze Encryption Format</string>
     <string name="encrypt_option_use_aescrypt_desc"><![CDATA[
-        <p>Choosing this option, Amaze will encrypt the files using its own encryption format.</p>
-        <p>Different than AESCrypt format, Amaze\'s Encryption Format can only be decrypted by this device; also you must not uninstall Amaze File Manager from this device, otherwise you will have no way to recover the encrypted file.</p>
+        <p>AESCrypt is an open source file encryption format based on AES encryption. It is also portable, can be decrypted with AESCrypt programs on other platforms.</p>
+        <p><a href="https://www.aescrypt.com" target="_blank>AESCrypt website</a></p>]]>
+    </string>
+    <string name="encrypt_option_use_azecrypt_desc"><![CDATA[
+        <p>Amaze\'s Encryption Format is also based on AES encryption; but different from AESCrypt format, it can only be decrypted by this device; also you must not uninstall Amaze File Manager from this device nor change the way you lock screen, otherwise you will have no way to recover the encrypted file.</p>
     ]]></string>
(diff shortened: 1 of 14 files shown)


VishnuSanal (approved):

VishnuSanal (changes requested):
Awesome work! Please consider these too.

1. Can we convert the info button on encrypt dialog to a TextView below the checkbox? As of now, it has a greater chance of getting ignored. As suggested on #3156 Or as Vishal suggested, the dialog appears just when the checkbox is checked.

2. The text fields can be empty and the OK button can be clicked! (some ways to do this would be to check the checkbox, uncheck the checkbox & focus on a text field etc) Can we have checks for this? It's okay if these are checked even after O.K button is clicked, we can set error to the TextInputLayout then too.

VishalNehra:
@TranceLove can you convert this pr to draft please if it's in wip

TranceLove:
Let's see if this is what you're looking for.

Screenshot_20220331_001928

Screenshot_20220331_001942

Meanwhile, because of the way we do field validations, this is probably the best we may have for the form. But please feel free to comment on a better way to do form validation on situations like this - I too think the error messages pop up at the wrong time.

VishalNehra:
> Let's see if this is what you're looking for.
> 
> Meanwhile, because of the way we do field validations, this is probably the best we may have for the form. But please feel free to comment on a better way to do form validation on situations like this - I too think the error messages pop up at the wrong time.

This looks good to me. Please also mention that amaze encryption is based on aes as well.

app/src/main/res/values/strings.xml
```diff
@@ -628,10 +628,12 @@
...
     <string name="encrypt_option_use_aze">Use Amaze Encryption Format</string>
-    <string name="encrypt_option_use_aescrypt_title">About Amaze Encryption Format</string>
     <string name="encrypt_option_use_aescrypt_desc"><![CDATA[
-        <p>Choosing this option, Amaze will encrypt the files using its own encryption format.</p>
-        <p>Different than AESCrypt format, Amaze\'s Encryption Format can only be decrypted by this device; also you must not uninstall Amaze File Manager from this device, otherwise you will have no way to recover the encrypted file.</p>
+        <p>AESCrypt is an open source file encryption format based on AES encryption. It is also portable, can be decrypted with AESCrypt programs on other platforms.</p>
+        <p><a href="https://www.aescrypt.com" target="_blank>AESCrypt website</a></p>]]>
+    </string>
+    <string name="encrypt_option_use_azecrypt_desc"><![CDATA[
+        <p>Amaze\'s Encryption Format is also based on AES encryption; but different from AESCrypt format, it can only be decrypted by this device; also you must not uninstall Amaze File Manager from this device, otherwise you will have no way to recover the encrypted file.</p>
```
VishnuSanal, line 636: Should we add that they shouldn't change the screen lock too?

VishnuSanal (approved):

VishnuSanal (approved):
🚀

VishalNehra (approved):

Issue #3207, opened 2022-03-06
Searching causes crashes


sjl872964789:
**Describe the bug**
I'm not sure which of my previous steps caused the problem, now amaze crashes every time I search t.

**To Reproduce**
Not sure

**Expected behavior**
Don't crash

**Screenshots**
![77](https://user-images.githubusercontent.com/15941953/156930516-0199211e-ba95-4685-85d9-d21db519d72f.png)

**Smartphone (please complete the following information):**
 - OS: Android 8.0
 - Rooted: No
 - Version: 3.6.7

**Additional context**

* __App Name:__ Amaze File Manager
* __Package:__ com.amaze.filemanager
* __Version:__ 3.6.7
* __User Action:__ UI Error
* __Request:__ Application crash
* __OS:__ Linux Android 8.0.0 - 26
* __Device:__ generic_x86
* __Model:__ Android SDK built for x86
* __Product:__ sdk_gphone_x86
<details><summary><b>Crash log </b></summary><p>

```
java.lang.NullPointerException: Attempt to read from field 'java.lang.String com.amaze.filemanager.adapters.data.LayoutElementParcelable.title' on a null object reference
 at com.amaze.filemanager.adapters.RecyclerAdapter.onBindViewHolder(RecyclerAdapter.java:765)
 at androidx.recyclerview.widget.RecyclerView$Adapter.onBindViewHolder(RecyclerView.java:7065)
 at androidx.recyclerview.widget.RecyclerView$Adapter.bindViewHolder(RecyclerView.java:7107)
 at androidx.recyclerview.widget.RecyclerView$Recycler.tryBindViewHolderByDeadline(RecyclerView.java:6012)
 at androidx.recyclerview.widget.RecyclerView$Recycler.tryGetViewHolderForPositionByDeadline(RecyclerView.java:6279)
 at androidx.recyclerview.widget.RecyclerView$Recycler.getViewForPosition(RecyclerView.java:6118)
 at androidx.recyclerview.widget.RecyclerView$Recycler.getViewForPosition(RecyclerView.java:6114)
 at androidx.recyclerview.widget.LinearLayoutManager$LayoutState.next(LinearLayoutManager.java:2303)
 at androidx.recyclerview.widget.LinearLayoutManager.layoutChunk(LinearLayoutManager.java:1627)
 at androidx.recyclerview.widget.LinearLayoutManager.fill(LinearLayoutManager.java:1587)
[...]

PR #3242, opened 2022-03-28
[bugfix] searching for hidden files crushes app

DoozyDoz:
## Description
- In response to issue #3207 , the app was crushing everytime  a user searches for a file/folder that is hidden.
- This happens when the adapter tries to access getters of a null `rowItem`.

#### Issue tracker   
Addresses #3207 

#### Automatic tests

- [ ] Added test cases
  
#### Manual tests
- [x] Done  
  
- Device: Huawei P30 lite
- OS: Android 10

#### Build tasks success  


VishalNehra:
This is fine but can we find why the search result has hidden item in it in the first place. We should avoid adding hidden items in search results instead of adding nulls in resulting list.

EmmanuelMess:
The entire method has to be rewritten separating the different item types.

VishalNehra:
I'm going to merge this, but will keep the issue open as this is just a temporary fix.

VishalNehra (approved):

PR #3269, opened 2022-04-21
Integrate amaze utils transfer and analyse in navigation drawer

VishalNehra:
#### Automatic tests

- [ ] Added test cases
  
#### Manual tests
- [x] Done  
  

#### Build tasks success  


EmmanuelMess:
Code looks clean.

app/src/main/java/com/amaze/filemanager/ui/Extensions.kt
```diff
@@ -53,6 +55,23 @@ fun Context.startActivityCatchingSecurityException(intent: Intent) {
     }
 }
 
+/**
+ * Refactor the alias, based on if Amaze Utils is installed, then we disable the alias intent-filters
+ * if not installed then we enable the amaze utilities alias
+ */
+fun Context.refactorAUAlias(shouldEnable: Boolean) {
+    val component = ComponentName(this, "com.amaze.filemanager.amazeutilsalias")
+    if (!shouldEnable) {
```
EmmanuelMess, line 64: Both do the same thing?

app/src/main/java/com/amaze/filemanager/ui/Extensions.kt
```diff
@@ -53,6 +55,23 @@ fun Context.startActivityCatchingSecurityException(intent: Intent) {
     }
 }
 
+/**
+ * Refactor the alias, based on if Amaze Utils is installed, then we disable the alias intent-filters
```
EmmanuelMess, line 59: What do you mean by refractor? Rename maybe? Refactor makes sense only for code.

app/src/main/java/com/amaze/filemanager/utils/Utils.java
```diff
@@ -444,4 +445,15 @@ public static void addShortcut(
 
     ShortcutManagerCompat.requestPinShortcut(context, info, null);
   }
+
+  public static boolean appInstalledOrNot(String uri, PackageManager pm) {
```
EmmanuelMess, line 449: Please don't add things to Utils, as they will get lost, create a specific class for these that make it easier to find later when needed.

app/src/main/java/com/amaze/filemanager/ui/views/drawer/Drawer.java
```diff
@@ -521,6 +523,50 @@ public void refreshDrawer() {
...
+            R.string.analyse_storage,
+            new MenuMetadata(
+                    () -> {
+                      boolean isAUInstalled = Utils.appInstalledOrNot(AboutActivity.PACKAGE_AMAZE_UTILS,
+                              mainActivity.getPackageManager());
+                      if (isAUInstalled) {
+                        try {
+                          Utils.openURL("amaze://teamamaze.xyz/analyse", mainActivity);
+                        } catch (ActivityNotFoundException e) {
+                          GeneralDialogCreation.showDownloadAmazeUtilsDialog(mainActivity);
```
EmmanuelMess, line 561: When does this Exception happen? Shouldn't `isAUInstalled` prevent this?

EmmanuelMess (changes requested):

app/src/main/java/com/amaze/filemanager/ui/Extensions.kt
VishalNehra: one enables the alias, other disables it

app/src/main/java/com/amaze/filemanager/ui/views/drawer/Drawer.java
VishalNehra: yes. this happens in case the deeplink can't be handled by amaze utils, but it is installed.
For people using old version of AU.

app/src/main/java/com/amaze/filemanager/ui/Extensions.kt
EmmanuelMess: Both branches have the same code.

VishalNehra: PackageManager.COMPONENT_ENABLED_STATE_DISABLED vs PackageManager.COMPONENT_ENABLED_STATE_ENABLED

VishnuSanal (approved):

PR #3317, opened 2022-05-15
Convert Java to Kotlin

hojat72elect:
## Description
I'm gradually converting Java parts of the app into kotlin; in my own tests I faced some NPEs and resolved them, but let me know if there are any Nullability-related issues that I haven't taken care of 🤝🙏 
The current implementation is bug free and correct enough but it can definitely get more efficient/nicer than this, for example, all of those base classes could be replaced with kotlin delegations 🤔
Plz let me know how I can make it better ☺️🔧   

#### Automatic tests

- [ ] Added test cases
  
#### Manual tests
- [x] Done  
  

1- Device: Bluestacks 5 --- OS: 7.1.2 Nougat
2- Device: Nexus 6p --- OS: 10 (API 29)

#### Build tasks success  


VishalNehra:
Thanks for your contribution! Please run `./gradlew spotlessApply` at the root as I can see quite a lot of indentation issues.
Also please merge AboutActivity and fix conflicts with upstream. 
Let me know if you face any issues.

hojat72elect:
The merge conflict of "AboutActivity" was resolved 😇
The repo passes SpotlessCheck 👌

There's a tiny possibility that this PR introduces some NPEs to the source code; because these kotlin classes are interacting with java source code and a null object might come from java to kotlin. I have tried to catch them as much as I can but let me know if I have missed anything🤔

after converting the whole repo to kt, almost all NPEs will go away.

ooyellow58:
I do not see any meaning converting java to kotlin

EmmanuelMess:
Please remember to remove "!!" as you convert files.

file_operations/src/main/java/com/amaze/filemanager/fileoperations/filesystem/usb/UsbOtgRepresentation.kt
```diff
@@ -0,0 +1,48 @@
...
+            other.productID &&
+            vendorID == other.vendorID &&
+            (
+                serialNumber == null && other.serialNumber == null ||
+                    serialNumber == other.serialNumber
+                )
+    }
+
+    /**
+     * why? exactly why?
```
EmmanuelMess, line 45: Please don't add questions as documentation. This a common hash function.

PR #3330, opened 2022-05-27
[WIP] FTP client support

TranceLove:
Don't touch. I will make another PR for the real deal.

Working:
- create folder
- create file
- delete file
- download file
- get file properties with `MLST` command
- upload file, which also rely on properly parsing `MLST` command responses
- rename file

Need fix:
- [x] network connection error handling
- [x] Problem after refactoring to `Callable`s, cannot display Toast and crashed app
- [ ] fix upload file test case, byte array comparison is flaky


app/src/main/java/com/amaze/filemanager/ui/activities/MainActivity.java
```diff
@@ -1961,9 +1962,10 @@ public void showSftpDialog(String name, String path, boolean edit) {
       if (i != -1) name = dataUtils.getServers().get(i)[0];
     }
     SftpConnectDialog sftpConnectDialog = new SftpConnectDialog();
-    SshConnectionPool.ConnectionInfo connInfo = new SshConnectionPool.ConnectionInfo(path);
+    FtpConnectionPool.ConnectionInfo connInfo = new FtpConnectionPool.ConnectionInfo(path);
```
anr-detector, line 1965: I'm just wondering why this isn't executed from another thread than the UI thread? This will create some ANR/lag.

PR #3360, opened 2022-06-15
Hotfix/3.7.2 to master

VishalNehra:
#### Automatic tests

- [ ] Added test cases
  
#### Manual tests
- [ ] Done  
  

#### Build tasks success  

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/application/AppConfig.java
+++ b/app/src/main/java/com/amaze/filemanager/application/AppConfig.java
@@ -68,7 +68,7 @@ import jcifs.Config;
     reportSenderFactoryClasses = AcraReportSenderFactory.class)
 public class AppConfig extends GlideApplication {
 
-  private static final Logger LOG = LoggerFactory.getLogger(AppConfig.class);
+  private Logger log = null;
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/amaze/filemanager/application/AppConfig.java
```diff
@@ -68,7 +68,7 @@
     reportSenderFactoryClasses = AcraReportSenderFactory.class)
 public class AppConfig extends GlideApplication {
 
-  private static final Logger LOG = LoggerFactory.getLogger(AppConfig.class);
+  private Logger log = null;
```
EmmanuelMess, line 71: What happened here?

EmmanuelMess (changes requested):

app/src/main/java/com/amaze/filemanager/application/AppConfig.java
VishalNehra: This needs to be initialized at application startup lifecycle instead of static initialization for logfile to be initialized.

EmmanuelMess: Why?

PR #3381, opened 2022-07-02
Prevent Google Play release variant to be able to update itself via apk selection

TranceLove:
## Description

Latest Google Play policy requires apps to block any attempts to update itself, either via online update or even allow user to install update over itself by allowing user to specify the apk to install. This change implements this restriction.

Details [here](https://www.androidpolice.com/google-firms-up-play-store-rules-for-which-apps-can-install-other-apps/).

Only Google Play release variant is affected; F-Droid variant can still do this.

Thanks to our user for the tipping!

#### Automatic tests

- [x] Added test cases (Should only run with play release variant builds)
  
#### Manual tests
- [x] Done  
  
- Device: Pixel 2 emulator
- OS: Android 11
Toast should pop up if attempt to install Amaze's own apk from within Amaze (Play release variant)

#### Build tasks success  

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/filesystem/files/FileUtils.java
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/files/FileUtils.java
@@ -61,6 +61,7 @@ import com.amaze.filemanager.ui.theme.AppTheme;
 import com.amaze.filemanager.utils.DataUtils;
 import com.amaze.filemanager.utils.OTGUtil;
 import com.amaze.filemanager.utils.OnProgressUpdate;
+import com.amaze.filemanager.utils.PackageInstallValidation;
 import com.cloudrail.si.interfaces.CloudStorage;
 import com.cloudrail.si.types.CloudMetaData;
 import com.googlecode.concurrenttrees.radix.ConcurrentRadixTree;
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/amaze/filemanager/filesystem/files/FileUtils.java
```diff
@@ -416,6 +417,18 @@ public static float readableFileSizeFloat(long size) {
    */
   public static void installApk(
       final @NonNull File f, final @NonNull PermissionsActivity permissionsActivity) {
+
+    try {
+      PackageInstallValidation.validatePackageInstallability(f);
+    } catch (IllegalArgumentException e) {
```
VishnuSanal, line 423: Did you mean `PackageCannotBeInstalledException`? When would the `Toast` get fired? Trying to install Amaze crashes the app for me.

VishnuSanal (changes requested):
please see the comment, I am getting a crash when I was expecting a `Toast`

app/src/play/java/com/amaze/filemanager/utils/PackageInstallValidation.kt
```diff
@@ -0,0 +1,68 @@
...
+    @Throws(PackageCannotBeInstalledException::class)
+    fun validatePackageInstallability(f: File) {
+        AppConfig.getInstance().run {
+            val packageInfo: PackageInfo? =
+                packageManager.getPackageArchiveInfo(
+                    f.absolutePath,
+                    PackageManager.GET_ACTIVITIES
+                )
+            if (packageInfo == null) {
+                throw PackageCannotBeInstalledException("Cannot get package info")
```
VishnuSanal, line 50: both cases are handled in the same way, is this intentional? i.e, "Cannot get package info" & "Cannot update myself per Google Play policy" shows the Toast message that the app can't be installed due to Google's limitations. a workaround that comes to my mind is to throw `IllegalStateException` like you did before & handle it with a different Toast message.

VishalNehra:
Changed base branch to release/3.8

VishnuSanal (approved):

app/src/main/java/com/amaze/filemanager/filesystem/files/FileUtils.java
```diff
@@ -416,6 +417,25 @@ public static float readableFileSizeFloat(long size) {
...
+      PackageInstallValidation.validatePackageInstallability(f);
+    } catch (PackageInstallValidation.PackageCannotBeInstalledException e) {
+      Toast.makeText(
+              permissionsActivity,
+              R.string.error_google_play_cannot_update_myself,
+              Toast.LENGTH_LONG)
+          .show();
+      return;
+    } catch (IllegalStateException e) {
+      Toast.makeText(
```
VishalNehra, line 431: Question - in case we can't get package info, we still allow user to install the apk right?

VishalNehra (approved):

VishalNehra:
This must go out asap. I'll publish a version in a day or two.

app/src/main/java/com/amaze/filemanager/filesystem/files/FileUtils.java
VishnuSanal: Yes, I think.

Voykin (approved):

PR #3414, opened 2022-07-28
Add fastlane for F-Droid

linsui:
## Description
These descriptions are copied from fdroiddata repo directly. F-Droid can fetch them from here so it can be updated easily.

#### Automatic tests

- [ ] Added test cases
  
#### Manual tests
- [ ] Done  
  

#### Build tasks success  


fastlane/metadata/android/en-US/short_description.txt
```diff
@@ -0,0 +1 @@
+Manage local files
```
VishalNehra, line 1: This short description is incorrect. Amaze also handles remote files on ftp / smb servers, I think a simply 
`Light and smooth file manager following the Material Design guidelines` should be okay.

VishalNehra (changes requested):

fastlane/metadata/android/en-US/short_description.txt
linsui: Looks like it's a very old description. I'll only keep the en-US and zh-CN descriptions since I can understand them.

VishalNehra (approved):

PR #3451, opened 2022-08-27
3450 Allow to delete drag and drop files

### LINKED ISSUE #3450: Drag and drop to delete files
Add a delete files area on screen when user is dragging files, dropping files in this area will trigger delete.

VishalNehra:
#### Issue tracker   

Fixes #3450 

#### Automatic tests

- [ ] Added test cases
  
#### Manual tests
- [x] Done  
  

#### Build tasks success  

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/ui/Extensions.kt
+++ b/app/src/main/java/com/amaze/filemanager/ui/Extensions.kt
@@ -25,6 +25,7 @@ import android.content.Context
 import android.content.Intent
 import android.content.pm.PackageManager
 import android.text.TextUtils
+import android.view.View
 import android.view.inputmethod.InputMethodManager
 import android.widget.EditText
 import android.widget.Toast
@@ -94,4 +95,14 @@ fun EditText.openKeyboard(context: Context) {
         },
         100
     )
+}
+
+fun View.hideFade(duration: Long) {
+    this.animate().alpha(0f).duration = duration
+    this.visibility = View.GONE
+}
+
+fun View.showFade(duration: Long) {
+    this.animate().alpha(1f).duration = duration
+    this.visibility = View.VISIBLE
 }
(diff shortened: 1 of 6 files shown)


VishnuSanal (commented):
awesome work! 

i noticed that the deletion confirmation dialog doesn't pop up when the list is scrolling (or when the action bar had collapsed, even if there is only one item in the folder). any idea on how to get around this? one idea that comes to my mind is to prevent scrolling when the delete shadow is visible?

PS: the haptic feedback gets triggered, but not the delete confirmation dialog.

VishalNehra:
> awesome work!
> 
> i noticed that the deletion confirmation dialog doesn't pop up when the list is scrolling (or when the action bar had collapsed, even if there is only one item in the folder). any idea on how to get around this? one idea that comes to my mind is to prevent scrolling when the delete shadow is visible?
> 
> PS: the haptic feedback gets triggered, but not the delete confirmation dialog.

Don't understand, please share a video if possible

VishnuSanal:
>Don't understand, please share a video if possible

https://user-images.githubusercontent.com/50027064/187170294-aabda961-c7c7-43c8-a3a0-fecbb0248b8f.mp4

VishalNehra:
So what's the issue here? It's going to scroll if you drag to any bottom part.

VishnuSanal:
> So what's the issue here? It's going to scroll if you drag to any bottom part.

um... A) then why do I get the haptic feedback for deletion? B) the same happens when there's not much files on the folder (say only one file) & when the action bar gets collapsed. C) can we handle that glitch somehow (of a non-checked item's bg getting gray color)?

VishalNehra:
> > So what's the issue here? It's going to scroll if you drag to any bottom part.
> 
> um... A) then why do I get the haptic feedback for deletion? B) the same happens when there's not much files on the folder (say only one file) & when the action bar gets collapsed. C) can we handle that glitch somehow (of a non-checked item's bg getting gray color)?

A) You get haptic feedback because you entered the trash button area. Doesn't mean the location you're currently dropping is also on trash button. You wouldn't know if you've exited it just based on vibration.
C) Non checked items bg getting gray is important as it's a visual clue that item's can be dropped inside a folder. I'll fix for non folders getting gray color.

PR #3452, opened 2022-08-27
3449 Add dont ask again for deletion dialog

### LINKED ISSUE #3449: Allow to delete files without confirmation dialog
Add a checkbox  in delete dialog which once checked will allow user to delete without confirmation next time onwards.

VishalNehra:
#### Issue tracker   

Fixes #3449

#### Automatic tests

- [ ] Added test cases
  
#### Manual tests
- [x] Done  
  

#### Build tasks success  


VishnuSanal:
it is highly dangerous to put a do not ask again checkbox, IMO. a user might check the checkbox that without reading the warning.

how about having a preference in settings - just so that there is more resistance to turning this on? this would ensure that it wouldn't get checked by accident too.

also, there's no way to turn this off as of now!

VishalNehra:
@VishnuSanal There is a way to turn this on/off from settings as well

VishnuSanal:
@VishalNehra Okay, haven't noticed, my bad. But I still think we shouldn't keep the 'do not ask again' checkbox.

VishnuSanal (approved):

PR #3521, opened 2022-09-28
Disable FtpHybridFileTests at all

TranceLove:
After introducing FtpHybridFileTest and deriatives since #3405, test runs were getting more flaky, hence it may be a better idea to disable them for good for the time being, while working on an alternative solution to offload the daemons outside the build, e.g. to Docker.

Band-it fix for sure, but the fact that it runs smooth locally but not on Github Actions is really annoying.


VishnuSanal (approved):
checks pass, must be good to go. preferably, an issue can be created when this get merged noting this.

EmmanuelMess (approved):

PR #3591, opened 2022-10-24
Drive size info in navigation drawer

### LINKED ISSUE #3383: Show storage space in navigation drawer
Since I have a local Nextcloud Server, I would like to see WebDAV support added. Since there's already SMB/and FTP/SFTP support, it would be nice to see support added for another popular NAS type.

Currently users could use the NextCloud app, but it's really not tightly integrated with the default Android File Manager.

Also I happen to notice there is an add-on for proprietary external cloud services, but it's paid only and only on Google Play. What if some user has a device without Google Play Services and no desire to give Google 30% of your payment, or the user is poor and can't afford to buy it? What are they supposed to do? My suggestion would be to implement this functionality free in the FDroid/IzzyOnDroid versions (keep it paid on Google Play), and maybe add a link to donate (like LiberaPay) so we can still support the project.

Third suggestion: It would be nice if the storage space was more upfront, like in Material Files (see screenshot) where users can just glance and see how much space if left without pressing buttons. 

I bring this up since I'm currently working on a custom ROM and I've been trying out some FOSS file managers, and I am thinking of including Amaze as the default file manager, however it is lacking some features like I just mentioned.

VishnuSanal:
#### Issue tracker   
Fixes #3383

#### Automatic tests
- [ ] Added test cases
  
#### Manual tests
- [x] Done  

- Device: Pixel 3A
- OS: Android Tiramisu x86_64
- Device: Realme Narzo 20A
- OS: Android 11
- Device: Pixel 2
- OS: Android API 16

#### Build tasks success


app/src/main/res/values/strings.xml
```diff
@@ -789,5 +789,6 @@ You only need to do this once, until the next time you select a new location for
     <string name="protocol_ftp" translatable="false">FTP</string>
     <string name="protocol_ftps">Secure FTP</string>
     <string name="unavailable">Unavailable</string>
+    <string name="free_of">"free of</string>
```
EmmanuelMess, line 792: Use a string with placeholders.

app/src/main/java/com/amaze/filemanager/ui/views/drawer/Drawer.java
```diff
@@ -1033,4 +1071,23 @@ public String getSecondPath() {
   public Billing getBilling() {
     return this.billing;
   }
+
+  private SpannableString getSpannableText(String text, String freeSpace, String totalSpace) {
+
+    String s =
+        String.format(
```
EmmanuelMess, line 1078: No, use the format provided by getString

app/src/main/java/com/amaze/filemanager/ui/views/drawer/Drawer.java
```diff
@@ -679,9 +694,30 @@ private void addNewItem(
       MenuMetadata meta,
       @DrawableRes int icon,
       @DrawableRes Integer actionViewIcon) {
+    addNewItem(menu, group, order, text, meta, icon, actionViewIcon, null, null);
+  }
+
+  private void addNewItem(
```
EmmanuelMess, line 700: Add NonNull for menu and metadata

EmmanuelMess (changes requested):

VishalNehra (approved):

Issue #3587, opened 2022-10-22
Renaming and only changing case fails


lukash:
**Describe the bug**
When renaming a file, if only changing case of a letter, the operation fails with an "already exists" error.

**To Reproduce**
1. Go to rename a file
2. Only change a case of a letter
3. Attempt to save

**Expected behavior**
Rename happens.

**Smartphone (please complete the following information):**
 - Device: Auss Zenfone 9
 - OS: Android 12
 - Rooted: No
 - Version: 3.8.2

---
This now has a bounty, [here](https://app.bountysource.com/issues/113047903-renaming-and-only-changing-case-fails).

Please claim after a PR fixing this issue is accepted.
Remember that you can also back the issue adding to the bounty!

VishnuSanal:
@lukash thanks for reaching out to us, but I'm afraid there's not much we can do about it since FAT filesystem is case insensitive. @TranceLove do you have anything to add?

TranceLove:
> @lukash thanks for reaching out to us, but I'm afraid there's not much we can do about it since FAT filesystem is case insensitive. @TranceLove do you have anything to add?

Yes that's true - FAT filesystems are case-insensitive, and we can do nothing about it.

lukash:
Well, I can rename "a" to "b" and then rename again to "A".

On that premise:
- as a user that doesn't want to deal with nonsense: A quality app shouldn't give me silly errors, it should handle this for me instead.

- as a developer: Come on, guys.
  Also, to elaborate, if the error is being raised from the FS layer, it either:
  - stores the name case-insensitive and there's a hard collision there, but then it has to store the original case somewhere else (seems inefficient). In that case there could be an API to just change the cased value?
  - or it stores case-sensitive value and for some reason just does an extra case-insensitive check. Worst case scenario, you have to do the double move I described. Yes, it's not atomic anymore. I still feel it's much better than the error.

EmmanuelMess:
This now has a bounty, [here](https://app.bountysource.com/issues/113047903-renaming-and-only-changing-case-fails).

Please claim after a PR fixing this issue is accepted.
Remember that you can also back the issue adding to the bounty!

PR #3593, opened 2022-10-25
Allow case sensitive local file rename

stevealexr:
## Description
The FAT filesystem in android sd card is case-insensitive. This allows case-sensitive renaming by checking whether the changed name and original name is equivalent case-insensitively and doing a double rename. Only the code for local file renaming is changed.
Not sure what indent to use.

#### Issue tracker   
Fixes [#3587](https://github.com/TeamAmaze/AmazeFileManager/issues/3587)

#### Automatic tests
- [ ] Added test cases
  
#### Manual tests
- [x] Done  
  
- Device: Resizable
- OS: Api 33

#### Build tasks success  

These tasks failed even with unmodified source code so probably local configuration issues

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/filesystem/Operations.java
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/Operations.java
@@ -443,8 +443,7 @@ public class Operations {
       private final DataUtils dataUtils = DataUtils.getInstance();
 
       private final boolean isLocalFilesystemCaseInsensitive =
-          oldFile.isLocal()
-              && Objects.requireNonNull(oldFile.getFile()).exists()
+          Objects.requireNonNull(oldFile.getFile()).exists()
               && new File(oldFile.getFile().getParent(), oldFile.getSimpleName().toUpperCase())
                   .exists()
               && new File(oldFile.getFile().getParent(), oldFile.getSimpleName().toLowerCase())
--- a/app/src/main/java/com/amaze/filemanager/filesystem/Operations.java
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/Operations.java
@@ -441,12 +441,18 @@ public class Operations {
 
       private final DataUtils dataUtils = DataUtils.getInstance();
 
-      private final Boolean isCaseSensitiveRename =
+      /**
+       * Determines whether double rename is required based on original and new file name regardless
+       * of the case-sensitivity of the filesystem
+       */
+      private final boolean isCaseSensitiveRename =
           oldFile.getSimpleName().equalsIgnoreCase(newFile.getSimpleName())
               && !oldFile.getSimpleName().equals(newFile.getSimpleName());
 
-      // random string that is appended to file to prevent name collision, max file name is 255
-      // bytes
+      /**
+       * random string that is appended to file to prevent name collision, max file name is 255
+       * bytes
+       */
       private static final String TEMP_FILE_EXT = "u0CtHRqWUnvxIaeBQ@nY2umVm9MDyR1P";
 
       private boolean localRename(@NonNull HybridFile oldFile, @NonNull HybridFile newFile) {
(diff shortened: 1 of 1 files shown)


VishalNehra:
Can you run `./gradlew checkspotless` and apply suggestions to fix the build failure please?

app/src/main/java/com/amaze/filemanager/filesystem/Operations.java
```diff
@@ -441,6 +441,63 @@ public static void rename(
...
+                  RenameFileCommand.INSTANCE.renameFile(file.getPath(), file1.getPath());
+                  result = true;
+                } catch (ShellNotRunningException e) {
+                  LOG.warn("failed to rename file in local filesystem", e);
+                }
+                oldFile.setMode(OpenMode.ROOT);
+                newFile.setMode(OpenMode.ROOT);
+                a = !file.exists() && file1.exists();
+              }
+              errorCallBack.done(newFile, a);
```
VishalNehra, line 483: Need a fallback mechanism in case the second rename fails, fallback should attempt to rename back to original file name.

app/src/main/java/com/amaze/filemanager/filesystem/Operations.java
```diff
@@ -441,6 +441,63 @@ public static void rename(
 
       private final DataUtils dataUtils = DataUtils.getInstance();
 
+      private final Boolean isCaseSensitiveRename =
+              !oldFile.getSimpleName().equals(newFile.getSimpleName()) &&
+              oldFile.getSimpleName().equalsIgnoreCase(newFile.getSimpleName());
+
+      // random string that is appended to file to prevent name collision
+      private static final String TEMP_FILE_EXT = "qwerty";
```
VishalNehra, line 449: This is too less random. Please randomize a bit more. Plus, suggested to put this at the beginning of the filename instead.

VishalNehra (changes requested):

app/src/main/java/com/amaze/filemanager/filesystem/Operations.java
stevealexr: Putting it at the beginning of filename requires paths which is only available for API level 26 or higher. Doing it manually by string manipulation is risky. I have changed the temporary constant to a 32-characters string generated from `!@#$%a-zA-z0-9`.

app/src/main/java/com/amaze/filemanager/filesystem/Operations.java
stevealexr: I added a simple check that attempt to rename back to original file name, not sure if it is sufficient.

stevealexr:
> Can you run `./gradlew spotlessCheck` and apply suggestions to fix the build failure please?

Managed to run spotlessCheck after applying this [fix](https://github.com/diffplug/spotless/issues/834) but assemble debug failure is caused by something else. I am using jdk 11.

app/src/main/java/com/amaze/filemanager/filesystem/Operations.java
```diff
@@ -441,6 +441,61 @@ public static void rename(
 
       private final DataUtils dataUtils = DataUtils.getInstance();
 
+      private final Boolean isCaseSensitiveRename =
```
EmmanuelMess, line 444: Why is the case sensitive rename different?

stevealexr: This approach only cares about the value changed and not the type of filesystem. Although it does unnecessary double renaming in case-sensitive filesystem, it doesn't need to check the type of filesystem. 

The 'localRename' function (can't think of a better name) is extracted from the original code to avoid code duplication.

app/src/main/java/com/amaze/filemanager/filesystem/Operations.java
VishalNehra: Thanks, changes seem fine. It would be a good practice to LOG.warn the inner else condition in case second rename fails.
Also, codacy does seem to complain can you fix the issue:

Check warning on line 673 in app/src/main/java/com/amaze/filemanager/filesystem/Operations.java

Codacy Production
/ Codacy Static Code Analysis
app/src/main/java/com/amaze/filemanager/filesystem/Operations.java#L673
These nested if statements could be combined

app/src/main/java/com/amaze/filemanager/filesystem/Operations.java
VishalNehra: > Why is the case sensitive rename different? Shouldn't there just be filesystems that support case sensitive renames and filesystems that do not? And the renames be the same function?

I think this should be okay as we know for sure most file systems that android supports doesn't support case sensitivity (exception ext4 in which case it needs to be kernel enabled). But I think it would make a marginal performance improvement if second part of the AND condition in this is used first.

VishalNehra:
> > Can you run `./gradlew spotlessCheck` and apply suggestions to fix the build failure please?
> 
> Managed to run spotlessCheck after applying this [fix](https://github.com/diffplug/spotless/issues/834) but assemble debug failure is caused by something else. I am using jdk 11.

I think it's building just fine now remotely. For your local builds, I think you can point to your local jre instead. Make the following change in gradle.properties (if you're in mac).
org.gradle.java.home=/Applications/Android Studio.app/Contents/jre/Contents/Home

app/src/main/java/com/amaze/filemanager/filesystem/Operations.java
stevealexr: I added a log message and combined those if statements.

app/src/main/java/com/amaze/filemanager/filesystem/Operations.java
```diff
@@ -441,6 +441,61 @@ public static void rename(
...
+      private boolean localRename(@NonNull HybridFile oldFile, @NonNull HybridFile newFile) {
+        File file = new File(oldFile.getPath());
+        File file1 = new File(newFile.getPath());
+        boolean result = false;
+
+        switch (oldFile.getMode()) {
+          case FILE:
+            int mode = checkFolder(file.getParentFile(), context);
+            if (mode == 2) {
+              errorCallBack.launchSAF(oldFile, newFile);
```
stevealexr, line 461: The errorCallback.launchSAF needs to be lifted out of this function or else another errorCallback will be called.

app/src/main/java/com/amaze/filemanager/filesystem/Operations.java
stevealexr: I added comment on the purpose of isCaseSensitiveRename variable and refactored double rename part. Hopefully it is clearer.

app/src/main/java/com/amaze/filemanager/filesystem/Operations.java
```diff
@@ -441,6 +441,74 @@ public static void rename(
 
       private final DataUtils dataUtils = DataUtils.getInstance();
 
+      /**
+       * Determines whether double rename is required based on original and new file name regardless
+       * of the case-sensitivity of the filesystem
```
EmmanuelMess, line 446: This _only_ depends on the case sensitivity of the filesystem.

stevealexr: I change it to a filesystem case-sensitivity check by checking whether both upper case and lower case input name for file exists return true.

VishalNehra: This new solution seem to be costly solution and calculating this each time is redundant. If a filesystem is case insensitive then it'll always be case insensitive, so you can persist the result of this in shared preferences and use that from next time.

stevealexr: There are multiple local filesystems like sdcard storage and internal storage. The types of filesystem need to be stored with the mount points. Rolling back to previous solution seems easier.

VishalNehra: Sure please revert back to your old solution. Also note that the new code will double rename even when FS is insensitive and new file and old file names aren't similar at all. 
oldFile.getSimpleName().equalsIgnoreCase(newFile.getSimpleName())
              && !oldFile.getSimpleName().equals(newFile.getSimpleName());

Please also note that I've increased the bounty price for this issue to 10USD you can claim once this PR is merged from the linked issue.

VishalNehra (dismissed):

VishalNehra (approved):

VishalNehra:
Please claim bounty [here](https://app.bountysource.com/issues/113047903-renaming-and-only-changing-case-fails)

stevealexr:
> Please claim bounty [here](https://app.bountysource.com/issues/113047903-renaming-and-only-changing-case-fails) @stevealexr

Screenshot 2022-11-18 093833

I can't submit the claim. No claim button.

PR #3617, opened 2022-11-09
added compile_sdk, min_sdk and target_sdk

shalva97:
#### Issue tracker   
There was 3 different Android SDKs included:
Screenshot 2022-11-09 at 21 52 03

In gradle files, it had different versions per module, so I added it to `ext`. Now it show only one:
Screenshot 2022-11-09 at 21 53 41

Is there a reason to have SDK versions different for modules? at least for me having it in ext block looks cleaner, also 2 less SDK is loaded by the studio and I think it should be faster

#### Automatic tests

- [ ] Added test cases
  
#### Manual tests
- [x] Done  
  

#### Build tasks success  

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -6,7 +6,7 @@ apply plugin: 'com.hiya.jacoco-android'
 apply plugin: "com.starter.easylauncher"
 
 android {
-    compileSdkVersion 31
+    compileSdkVersion compile_sdk
     packagingOptions {
         resources {
             excludes += ['proguard-project.txt', 'project.properties', 'META-INF/LICENSE.txt', 'META-INF/LICENSE', 'META-INF/NOTICE.txt', 'META-INF/NOTICE', 'META-INF/DEPENDENCIES.txt', 'META-INF/DEPENDENCIES']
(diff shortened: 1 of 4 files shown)


JeanPaulLucien (commented):
I don't like this code. You creates new variables for nothing.

Amaze use 3rd party librabries. These libraries have own requirements. Change configs is bad idea, because nobody tested libs with all custom configs.

shalva97:
> These libraries have own requirements

Should not we update them? API 28 was released at 2018...

If you look at build.gradle file commits, you will see that their SDKs were never updated, except the root build.gradle.

> You creates new variables for nothing.

if there were those variables probably it would be updated long ago. Anyways Im not 100% sure it should be same version accross all modules, but I also want to know why not. For me having less libs showing up in the list pleases my OCD...

VishnuSanal (commented):
> @shalva97 hi, thanks a lot for your contribution! but, this currently has merge conflicts + there aren't any significant benefits to the changes. this will be superseded by #4215 too. we can merge it if you can fix the merge conflicts & the tests pass.

closing this for now; feel free to reopen.

PR #3676, opened 2022-12-24
Migrate PemToKeypairTask to Observable 

### LINKED ISSUE #1364: Use multiple PemToKeyPairTask instead of one with infinite loop
At current implementation, one single instance of `PemToKeyPairTask` is used for converting PEM to `KeyPair`, but this is not a recommended approach of using `AsyncTask`s (which encourages distinct AsyncTask for each key-passphrase combination).

TranceLove:
## Description
We are still reusing the same class when decoding PEM, but is looping with RxJava's Observable which should be doing things a lot better.

#### Issue tracker   
Fixes #1364

#### Automatic tests
- [x] Added test cases
  
#### Manual tests
- [ ] Done  
  
#### Build tasks success  


TranceLove:
Pending tasks:

- [ ] Manual test on devices for SSH key-based authentication
- [ ] Add tests for verifying failure scenarios (if possible)

TranceLove:
Manual test completed without problem on Pixel 2 emulator running Android 8.0 and Pixel 6 emulator running Android 13.

Tried with Galaxy Nexus emulator running Android 4.4, but its DocumentsUI didn't even respond with `Activity.RESULT_OK` upon calling file selection, hence couldn't help much here - be it AsyncTask in previous implementation would have failed anyway.

VishalNehra (approved):

.github/workflows/android-feature.yml
```diff
@@ -45,10 +45,3 @@ jobs:
         uses: gradle/gradle-build-action@v2
         with:
           arguments: jacocoTestPlayDebugUnitTestReport --stacktrace --info
-      - name: Publish test results
-        uses: dorny/test-reporter@v1
-        if: always()
-        with:
-          name: test-results
-          path: 'app/build/test-results/testPlayDebugUnitTest/TEST-*.xml'
-          reporter: java-junit
```
VishnuSanal, line 54: why? 🤔

TranceLove: As I read from https://github.com/dorny/test-reporter#recommended-setup-for-public-repositories some refactoring to the workflow is required for test results even from pull requests from forks to deliver properly without causing errors.

Shall discuss with teammates about this later.

PR #3681, opened 2022-12-29
Upgrade commons-compress to 1.22

TranceLove:
## Description
Prevent out of memory errors.

https://security.snyk.io/package/maven/org.apache.commons:commons-compress/1.20

Aside, also backported the sevenz package using classes that are available from API 24 for better compatibility.

#### Issue tracker   
Addresses #3187

#### Automatic tests
- [ ] Added test cases
  
#### Manual tests
- [x] Done  
  
#### Build tasks success  

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -30,7 +30,7 @@ buildscript {
         espressoVersion = "3.4.0"
         materialDialogsVersion = "0.9.6.0"
         jacocoVersion = "0.8.7"
-        commonsCompressVersion = "1.20"
+        commonsCompressVersion = "1.22"
         libsuVersion = "3.2.1"
         mockkVersion = "1.12.2"
         logbackAndroidVersion = "2.0.0"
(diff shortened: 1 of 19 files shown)


TranceLove:
Pending tasks:

- [ ] Confirm other packages are also downward compatible with older Androids
- [ ] Manual tests

TranceLove:
> Confirm other packages are also downward compatible with older Androids

`org.apache.commons.compress.utils.IOUtils` needs forking too due to presence of `java.nio.file.LinkOption` which exists in Android 8.0 only.

> Manual tests

Tested on GPD XD running LegacyROM (4.4.4), successfully unzipped not-password-protected 7z archive.

Robolectric tests can't tell you much that your code is not downward compatible with older Androids, so the best way is to let it test in the field.

VishalNehra (approved):

NanyangTaiji:
Is possible to zip file with password?

Issue #3628, opened 2022-11-13
Attempt to move file over SFTP fails


Cebtenzzre:
**Describe the bug**
When I am using SFTP without accessing an NFS filesystem (which causes #3626), I run into this issue instead. Any attempt to move a file appears to cause some kind of crash - the SFTP connection apparently closes and all folders report "No Files". An exception from "Bundle" is always logged when this happens. This does not happen when I attempt to copy a file instead.

**To Reproduce**
1. Open an SFTP connection
2. Browse to a location not backed by an NFS filesystem
3. Cut a file
4. Browse to a different directory on the SFTP connection and paste the file
5. Amaze refreshes and displays "No Files". An exception is printed to logcat
6. All attempts to browse the SFTP connection now display "No Files", and the SFTP connection cannot be used until closing and reopening the app
7. The file has not actually moved according to the server

**Expected behavior**
I would expect the file to be moved, and to be able to continue to use the SFTP connection.

**Smartphone**
 - Device: Google Pixel 5a
 - OS: Android 13
 - Rooted: Yes, Magisk
 - Version: 3.8.2

**Additional context**
This is the exception that appears in logcat:
```
11-13 16:47:24.358 20885 20885 W Bundle  : Key failedOps expected ArrayList<java.lang.String> but value was of a different type.  The default value <null> was returned.
11-13 16:47:24.358 20885 20885 W Bundle  : Attempt to cast generated internal exception:
11-13 16:47:24.358 20885 20885 W Bundle  : android.os.BadTypeParcelableException: Parcelable creator com.amaze.filemanager.filesystem.HybridFileParcelable is not a subclass of required class java.lang.String provided in the parameter
11-13 16:47:24.358 20885 20885 W Bundle  : 	at android.os.Parcel.readParcelableCreatorInternal(Parcel.java:4852)
11-13 16:47:24.358 20885 20885 W Bundle  : 	at android.os.Parcel.readParcelableInternal(Parcel.java:4763)
11-13 16:47:24.358 20885 20885 W Bundle  : 	at android.os.Parcel.readValue(Parcel.java:4529)
11-13 16:47:24.358 20885 20885 W Bundle  : 	at android.os.Parcel.readValue(Parcel.java:4312)
11-13 16:47:24.358 20885 20885 W Bundle  : 	at android.os.Parcel.readListInternal(Parcel.java:5273)
[...]

PR #3888, opened 2023-07-15
Bugfix/ssh copy fixes 2

TranceLove:
## Description
- Fix broken session caused by unclosed sessions (again)
- Use `du` command at remote to calculate directory size

One known issue though is the current implementation requires rapidly connect and disconnect from the SSH server, as seen in logcat that a lot of calls to attempt fetching SSH private key for authentication. This drastically slows down copying files over SSH, and the implementation itself does not make sense - that it should only disconnect the SFTPClient spawn from SSHClient when finish copying a file, if it cannot survive copying multiple files.

Such fix shall require bigger scale refactoring at CopyService/GenericCopyUtil that it should make up the list of files to copy recursively before starting, instead of walk down the directory tree on demand recursively. Right now it seems to be the only way to make the best of all worlds :(

~~Depends on #3884 before codebase can compile~~ DONE
 
#### Manual tests
- [x] Done  
  
- Device: Fairphone 3
- OS: LineageOS 20 (13)

- Copy folder of files should be working now
- Display thumbnails of folder of images should be functioning without crashing the whole SSH session

#### Build tasks success  


TranceLove:
Pending:

- [ ] check if it can fix #3628
- [ ] endurance test: copy bigger folder structures

VishalNehra (approved):

VishalNehra:
@TranceLove  LGTM, is it ok if I merge this?

TranceLove:
> @TranceLove LGTM, is it ok if I merge this?

Yes sure. Still considering the options for optimizing/fixing the copy process.

VishalNehra (approved):

Issue #3886, opened 2023-07-11
NPE when pasting a deleted/modified file


ssssy1000:
**Describe the bug**
I copy the file and want to paste it, to avoid the duplicate file names, I rename the file, but the  error is happen in paste operation.

**To Reproduce**
1. Copy the file
2. Delete(Rename) the file
3. click Paste
4. The crash happen

**Screenshots**
<img height="600" alt="1689082393588" src="https://github.com/TeamAmaze/AmazeFileManager/assets/98003627/5ee2141a-3041-406f-a96e-3fddd9b2d51d"> <img height="600" alt="1689082443816" src="https://github.com/TeamAmaze/AmazeFileManager/assets/98003627/dc355373-bddb-435c-ae04-669a763bf080"> <img height="600" alt="1689082650898" src="https://github.com/TeamAmaze/AmazeFileManager/assets/98003627/aee6e53a-fd34-44e2-93f3-3c9d50b6220f"> <img height="600" alt="1689082468975" src="https://github.com/TeamAmaze/AmazeFileManager/assets/98003627/92290d08-9678-46c7-b815-faa713215833">

**Smartphone (please complete the following information):**
- Device: Google Pixel 5
- OS: Android 7.1.1
- Rooted: No
- Version: 3.8.5

**Crash logs**
```
[AsyncTask #9        ] Got exception checkout: /storage/emulated/0/DCIM/GrowTracker/2023-07-09-05-34-03/Text.txtjava.lang.NullPointerException: Attempt to invoke virtual method 'android.net.Uri androidx.documentfile.provider.DocumentFile.getUri()' on a null object reference
	at com.amaze.filemanager.filesystem.files.GenericCopyUtil.startCopy(GenericCopyUtil.java:166)
	at com.amaze.filemanager.filesystem.files.GenericCopyUtil.copy(GenericCopyUtil.java:316)
	at com.amaze.filemanager.asynchronous.services.CopyService$DoInBackground$Copy.copyFiles(CopyService.java:522)
	at com.amaze.filemanager.asynchronous.services.CopyService$DoInBackground$Copy.execute(CopyService.java:406)
	at com.amaze.filemanager.asynchronous.services.CopyService$DoInBackground.doInBackground(CopyService.java:270)
	at com.amaze.filemanager.asynchronous.services.CopyService$DoInBackground.doInBackground(CopyService.java:237)
	at android.os.AsyncTask$2.call(AsyncTask.java:305)
[...]

Issue #2078, opened 2020-11-09
Rename downloaded pages if the file name exists


HT-7:
**Is your feature request related to a problem? Please describe.**
While I like the feature which allows downloading pages when sending URLs to it, it has one problem:

When downloading the page a second time (i.e. because it has updated), then the existing HTML file will be **overwritten**.

Overwriting by default is a poor behaviour (not programmer's fault but default in UNIX, as far as I know, thus only a result of neglect), which has caused me to lose my backed up log by 3C Battery Monitor Widget (file name: `bmw_history.txt`).

**Describe the solution you'd like**
If the file name exists, **rename** the new files to `example(1).html`, `example(2).html` , and continuing.

**Describe alternatives you've considered**
Google Chrome for Android allows MHTML downloading, but not in Incognito mode.

Mozilla Firefox for Android allows PDF downloading, but PDF is not quite a website format (better than nothing, obviously).

HT-7:
Side note: Yes, I could indeed rename the existing HTML file in advance, but doing that every single time is inconvenient. And one could forget to do it.

EmmanuelMess:
We should add a dialog to put in a new name.

VishalNehra:
I think dialog would be overkill. Just create new file with appended (1) or (2) should be enough I guess. Google chrome download handles it similarly.

TranceLove:
@VishalNehra I wonder if there's a library handle all this naming stuff?

VishalNehra:
Edit: maybe use the same dialog which we use when copying existing file. We already have it to handle Overwrite / skip actions. We can add one more action to Rename it to (1) or (2) instead. It'll help in copy / cut scenario as well.
 Let me pick this up.
@TranceLove  i don't think we need a library for this. Should be straightforward.

VishalNehra:
Possible duplicate #1589

VishalNehra:
Duplicate #1366 
Related #721
Will pick this and close all the issues.

EmmanuelMess:
v3.5 milestone? It should be at least 3.6 this is not a bug.

PR #3913, opened 2023-08-19
Implement filename renaming library

TranceLove:
## Description

It is a port of https://github.com/jonschlinkert/add-filename-increment for node.js into Kotlin, with some tweaks and tricks specific to Kotlin.

#### Issue tracker   
Addresses #2078

#### Automatic tests
- [x] Added test cases
  
#### Manual tests
- [ ] Done  
  
#### Build tasks success  

#### Related PR  
Related to PR #3886


yatiksihag01:
Hi @TranceLove, I've noticed a couple of issues in your code:
- When renaming directories, the `increment()` function currently appends a dot (.) at the end of the directory name. It should append numbers like (1), (2), and so on. Please check line 155 in `FilenameHelper.kt`.
- If I pass value of startArg 0 or 1 (default), it starts from appending (2), (3) and so on. However, when I use a higher value like 5, it works correctly.

TranceLove:
@yatiksihag01 thanks for the feedback. Added a4ad60f to address the first issue, please take a look.

But for the second problem... may you be more specific about the symptom(s) you have? Say when you call `increment()` with the default `startArg`, what should I expect and why setting `startArg` to high values would work?

app/src/main/java/com/amaze/filemanager/filesystem/FilenameHelper.kt
```diff
@@ -0,0 +1,239 @@
...
+                        "$stem (another copy)"
+                    }
+                    else -> {
+                        "$stem (${toOrdinal(n)} copy)"
+                    }
+                }
+            }
+            // Windows and default formatting are the same.
+            else -> {
+                if (n > 1) {
```
yatiksihag01, line 231: Here is the second problem. Let's say `startArg` initial value is 1. When `increment()` function initially call `format()` , it'll simply return the same filename because `n <= 1`. Then `increment()` will call `format()` for n = 2 and it'll work fine. I don't know if there's specific reason to add this `if (n>1)` condition but in my case this is creating the problem.

TranceLove: I think it's understandable, if looking at the unit test for windows.

https://github.com/TeamAmaze/AmazeFileManager/blob/bda61488a4e75e3d37e21de85a094ed47c9ec35b/app/src/test/java/com/amaze/filemanager/filesystem/FilenameHelperWindowsIncrementNameTest.kt#L40-L48

Couldn't comment if it's right or wrong, as I was doing a direct port... but if I understand correctly, looks like you'd want to have filenames from starting with `file (2).txt` to `file (1).txt`, which is different from the original node.js implementation.

If that is the case, see if 1f2cd85 fits your needs.

yatiksihag01: Yes, now it's working fine.

VishalNehra (approved):

PR #3898, opened 2023-08-04
Fixes #3886

yatiksihag01:
## Description
Disabled file modification while the copy/paste snackbar is visible. This behavior is expected, and it is implemented in almost all popular file manager apps. When users choose to paste a file into the same folder from which they copied it, they will be presented with the option to Rename & Save. This option will automatically append (1), (2), etc., at the end of the file name and save it.

#### Issue tracker   
Fixes #3886 

#### Automatic tests

- [ ] Added test cases
  
#### Manual tests
- [x] Done  
  

- Device: Samsung Galaxy A13
- OS: Android 13

#### Build tasks success  

Successfully running following tasks on local:
- [x] `./gradlew assembledebug`
- [x] `./gradlew spotlessCheck`

PR #3941, opened 2023-09-27
Fixes #3938

### LINKED ISSUE #3938: [Debug Bug] Renaming a file before copying breaks File Renaming
from #3898 

### steps to reproduce

- Copy a directory with files, say `/storage/emulated/0/DCIM/Quotes/` -> `/storage/emulated/0/Pictures/`
- now we have the same contents (say, three PNG files) on `/storage/emulated/0/DCIM/Quotes/` & `/storage/emulated/0/Pictures/Quotes`
- Copy the files from `/storage/emulated/0/DCIM/Quotes/` to `/storage/emulated/0/Pictures/Quotes` & select `Rename` option on the dialog box.
- It works.
- Now, rename `/storage/emulated/0/DCIM/Quotes/` to `/storage/emulated/0/DCIM/QuotesNew/`
- Repeat the same operation & select rename, it fails to paste. 
- Pasting fails if we do cut operation too, but, the source files seems to be deleted
- This doesn't get resolved even after an app restart

ping @yatiksihag01

yatiksihag01:
## Description
- Fixed renaming issue while pasting conflicting files multiple times in same set of two directories.
- Disabled overwrite button if pasting in same directory. In previous implementation, we were showing users a Toast message if they try to overwrite a file in same directory. This new method seems more concise and more expressive.
- Set dialog shown during conflict resolution non-cancellable. This change ensures that users won't accidentally lose progress if they tap elsewhere on the screen.

#### Issue tracker   

Fixes #3938

#### Automatic tests

- [ ] Added test cases
  
#### Manual tests
- [x] Done  
  

- Device: Samsung Galaxy A13
- OS: Android 13

#### Build tasks success  

#### Related PR  
Related to PR #3898

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/movecopy/PreparePasteTask.kt
+++ b/app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/movecopy/PreparePasteTask.kt
@@ -127,6 +127,7 @@ class PreparePasteTask(strongRefMain: MainActivity) {
 
         if (isCloudOrRootMode) {
             startService(filesToCopy, targetPath, openMode, isMove, isRootMode)
+            return
         }
 
         val totalBytes = FileUtils.getTotalBytes(filesToCopy, context.get())
@@ -188,7 +189,7 @@ class PreparePasteTask(strongRefMain: MainActivity) {
                 }
             )
             withContext(Dispatchers.Main) {
-                prepareDialog(conflictingFiles, filesToCopy)
+                prepareDialog(conflictingFiles, conflictingDirActionMap)
                 @Suppress("DEPRECATION")
                 progressDialog?.setMessage(context.get()?.getString(R.string.copying))
             }
@@ -198,7 +199,7 @@ class PreparePasteTask(strongRefMain: MainActivity) {
 
     private suspend fun prepareDialog(
         conflictingFiles: MutableList<HybridFileParcelable>,
-        filesToCopy: ArrayList<HybridFileParcelable>
+        conflictingDirActionMap: HashMap<HybridFileParcelable, String>
     ) {
         if (conflictingFiles.isEmpty()) return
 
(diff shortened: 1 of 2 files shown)


yatiksihag01:
> Disabled overwrite button if pasting in same directory. In previous implementation, we were showing users a Toast message if they try to overwrite a file in same directory. This new method seems more concise and more expressive.

@VishnuSanal tell me what you think about it. Should we keep this implementation or the previous one in which we were showing a toast message.

VishnuSanal:
> > Disabled overwrite button if pasting in same directory. In previous implementation, we were showing users a Toast message if they try to overwrite a file in same directory. This new method seems more concise and more expressive.
> 
> @VishnuSanal tell me what you think about it. Should we keep this implementation or the previous one in which we were showing a toast message.

I vote for the new implementation. #3938 is fixed & works perfect! Consider this an approval from my side, let @VishalNehra & @TranceLove give approvals before merging.

PS: please fix the CI :)

yatiksihag01:
> Execution failed for task ':app:testPlayDebugUnitTest'.
    Process 'Gradle Test Executor 27' finished with non-zero exit value 10
    This problem might be caused by incorrect test process configuration.

I examined CI logs and found that this issue appears to be related to a `org.gradle.api.tasks.TaskExecutionException`. As you can see it suggest that error might be caused by incorrect test configuration. But I didn't change any test configuration. Additionally, there is limited online information available regarding the meaning of exit code 10. Initially Executor 27 initiated the test execution but finished with exit code 10 after the subsequent executors 28, 29,.., 31 had completed their tasks..
@VishnuSanal, @VishalNehra & @TranceLove any idea how can I fix it?

VishnuSanal:
@yatiksihag01 I re-ran the CI & apparently it passes. @TranceLove please confirm whether we can ignore the previous failure.

VishalNehra (approved):

Issue #3756, opened 2023-03-08
`ActivityNotFoundException` on `Utils#openURL`


lightningvr12:
## Exception
* __App Name:__ Amaze File Manager
* __Package:__ com.amaze.filemanager
* __Version:__ 3.8.4
* __User Action:__ UI Error
* __Request:__ Application crash
* __OS:__ Linux Android 10 - 29
* __Device:__ hollywood
* __Model:__ Quest
* __Product:__ hollywood
<details><summary><b>Crash log </b></summary><p>

```
android.content.ActivityNotFoundException: No Activity found to handle Intent { act=android.intent.action.VIEW dat=market://details?id=com.amaze.fileutilities }
	at android.app.Instrumentation.checkStartActivityResult(Instrumentation.java:2058)
	at android.app.Instrumentation.execStartActivity(Instrumentation.java:1716)
	at android.app.Activity.startActivityForResult(Activity.java:5211)
	at androidx.activity.ComponentActivity.startActivityForResult(ComponentActivity.java:597)
	at android.app.Activity.startActivityForResult(Activity.java:5169)
	at androidx.activity.ComponentActivity.startActivityForResult(ComponentActivity.java:583)
	at android.app.Activity.startActivity(Activity.java:5540)
	at android.app.Activity.startActivity(Activity.java:5508)
	at com.amaze.filemanager.utils.Utils.openURL(Utils.java:380)
	at com.amaze.filemanager.ui.activities.UtilitiesAliasActivity.onCreate$lambda-0(UtilitiesAliasActivity.kt:46)
	at com.amaze.filemanager.ui.activities.UtilitiesAliasActivity.$r8$lambda$8quQpF0TkkkHRQu_61tXPbplAbc(Unknown Source:0)
	at com.amaze.filemanager.ui.activities.UtilitiesAliasActivity$$ExternalSyntheticLambda0.onClick(Unknown Source:2)
	at android.view.View.performClick(View.java:7259)
	at android.view.View.performClickInternal(View.java:7236)
	at android.view.View.access$3600(View.java:801)
	at android.view.View$PerformClick.run(View.java:27892)
	at android.os.Handler.handleCallback(Handler.java:886)
	at android.os.Handler.dispatchMessage(Handler.java:102)
	at android.os.Looper.loop(Looper.java:227)
	at android.app.ActivityThread.main(ActivityThread.java:7436)
	at java.lang.reflect.Method.invoke(Native Method)
[...]

PR #3963, opened 2023-10-22
Fix issue ActivityNotFoundException on Utils#openURL

gustamor:
The previous code had a bug that makes an ActivityNotFoundException when a user tries to open a URL on a system with any browser enabled, which is common on Oculus devices, for instance. This commit uses PackageManager to check if there is any webview enabled. It includes the following changes:

- Added PackageManager query to find a match with any enabled webview.
- Starts the activity only if there is a match.
- Logs a warning when there is no webview.
- Displays a Toast message indicating that no browser is available.
- Added some translations of the message

This fix resolves the reported bug and improves the overall user experience when opening URLs.

#### Issue tracker   

Fixes: #3756 

#### Automatic tests
- [ ] Added test cases
  
#### Manual tests
- [x] Done  
- Device: Generic Android Emulator
- OS: Android 10, Android 13 without Google API and browsers disabled

#### Build tasks success  


gustamor:
@TranceLove What do you think of these changes?

VishnuSanal (dismissed):
awesome work, thanks for fixing this!

PS: we do not encourage adding translations directly to the code. please use [Transifex](https://www.transifex.com/amaze/amaze-file-manager/) for translations. @VishalNehra please verify whether we can merge it as is, or should these changes be reverted?

VishalNehra (dismissed):

PR #3986, opened 2023-11-18
Feature: Option for theme to follow power save mode

seelchen:
## Description
This PR adds a new appearance configuration whether the theme should follow the battery saver mode of the device. The option only appears for themes that can be light.

A `BroadcastListener` is added to `ThemedActivity` to be able to respond when the battery saver mode is turned on or off.

https://github.com/TeamAmaze/AmazeFileManager/assets/74261221/db230675-d39f-440e-be07-d67a0d7b919d

https://github.com/TeamAmaze/AmazeFileManager/assets/74261221/0041a94f-957a-4701-bf88-5f5d3508dc8f

#### Issue tracker   

Fixes #2748

#### Automatic tests

- [x] Added test cases
  
#### Manual tests
- [x] Done  
  
Device: 
- Pixel 7 emulator running Android 13
- HTC U11 life running Android 10

#### Build tasks success  

Successfully running following tasks on local:
- [x] `./gradlew assembledebug`
- [x] `./gradlew spotlessCheck`

PR #3994, opened 2023-11-22
Refactor `AppTheme` enum

### LINKED ISSUE #3988: Refactor `AppTheme` enum
After working with the `AppTheme` enum, I think that the information it represents is very confusing. For some parts in the code it represents the preference selected by the user (Light, Dark, Timed, Black, System) but in other parts it represents the theme that should be used to generate the UI (Light, Dark, Black).
Because of the different meaning, there is also an implicit rule that `AppThereme#getSimpleTheme` has to be called when the theme is used to generate the UI.

My suggestion is that it should be split into two enum types. The different types can help developers differentiate between the two. It also changes the implicit rule to call `AppThereme#getSimpleTheme` to an explicit one since this would be the function to convert the one into the other.

Do you agree that this would be an improvement?
Since I'm already working on #2748, I could fix this in that PR or also open a new one for the change.

seelchen:
## Description
This PR splits the `AppTheme` enum into two enums.`AppThemePreference` represents the theme that the user selects in the settings while `AppTheme` represents the theme/color that the app should use for the UI.

The PR is branched off of the branch for #3986 since I made some changes to `AppTheme` there, so this should only be merged after #3986.

#### Issue tracker   

Fixes #3988

#### Automatic tests

- [ ] Added test cases
  
#### Manual tests
- [x] Done  
  
Device:

#### Build tasks success  

#### Related PR  
Related to PR #3986

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/adapters/AppsRecyclerAdapter.kt
+++ b/app/src/main/java/com/amaze/filemanager/adapters/AppsRecyclerAdapter.kt
@@ -292,7 +292,7 @@ class AppsRecyclerAdapter(
             if ((
                 fragment.requireActivity()
                     as MainActivity
-                ).appTheme.getSimpleTheme(fragment.requireContext()) == AppTheme.BLACK
+                ).appTheme == AppTheme.BLACK
             ) {
                 context = ContextThemeWrapper(context, R.style.overflow_black)
             }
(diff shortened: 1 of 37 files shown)


seelchen:
This is ready for review, but the tests fail again because `'Gradle Test Executor 16' finished with non-zero exit value 10`.
I looked around a bit and there is an issue [#8395](https://github.com/robolectric/robolectric/issues/8395) at robolectric that mentions this. The issue doesn't have a solution but this could be the cause for the exception.

VishnuSanal:
It passes on second attempt!

VishalNehra (approved):

Issue #3357, opened 2022-06-14
Can search results be sorted by relevance


LL12138LL:
**Is your feature request related to a problem? Please describe.**
When I search for some files with short names, I find it hard to find them because of too many search results

**Describe the solution you'd like**
Can search results be sorted by relevance

**Additional context**

https://user-images.githubusercontent.com/101189845/173608484-851993cd-12e1-4529-8a4e-15696c7f7233.mp4

seelchen:
Hi, I would like to work on this.

My proposed sorting by relevance would be (from most to least important criteria):
1. Sort based on percentage matched. We would calculate how much of the name of a file is found in the search term (e.g. the search term "my" would match 20% in "my-cat.txt")
2. Prioritize files where the match is in the beginning of the name (e.g. with the search term "my", "my-cat.txt" would be prioritized before "rat-my.txt")
3. Prioritize files where the match could be a whole word in the file name (because it is separated with "-"/"_"/"."/space, e.g. with "my" as search term, "my-cat.txt" is prioritized before "mysterious")
4. Date modified

I also think than directories should not appear first in the search results but should be ranked with the same weight as files. 

It might also be helpful to be able to select which sorting should be applied, e.g. by relevance, by name, by date modified etc.

VishnuSanal:
@seelchen hi, great to see that you are interested in working on Amaze. there's a new search technique being implemented in #3730. please see [this](https://github.com/orgs/TeamAmaze/discussions/3984) discussion for more details. let's discuss this new behavior before you start working on it.

seelchen:
Hi, thank you for the pointers!
I looked at the new search technique and it seems like the results are displayed in the order in which they are loaded in [`ListFileCommand#listFiles`](https://github.com/TeamAmaze/AmazeFileManager/blob/release/4.0/app/src/main/java/com/amaze/filemanager/filesystem/root/ListFilesCommand.kt#L46) (which doesn't explicitly give any guarantees about the order). The old search is also used as `DeepSearch`. I think that sorting by relevance could improve the usability for all three search modes.

seelchen:
My plan is to reuse the `FileListSorter` to sort both `LayoutElementParcables` (used in old search) and `HybridFileParcelables` (used in new search). I would need make them both implement a new interface to access the information needed for sorting. The sorting would be done after all items are found.

PR #4003, opened 2023-12-03
Sorting of search results

seelchen:
## Description
This PR adds the ability to sort the search results in the `SearchView`. For this, a button was added to the `SearchView` which opens a dialog similar to the one for sorting directories.

Additionally, to address issue #3357, this PR adds an option to sort the search results by relevance. This sorting is as described in [this comment](https://github.com/TeamAmaze/AmazeFileManager/issues/3357#issuecomment-1813236314) and is only available for sorting search results and not for directories.

To fully resolve issue #3357, the "old" search has to be embedded in the same search view as the new search view.
However, to avoid bloating this PR, I think it might be better to fix that separately.

#### Issue tracker   

Addresses #3357

#### Automatic tests

- [x] Added test cases
  
#### Manual tests
- [x] Done  
  
Device:

#### Build tasks success  

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/adapters/RecyclerAdapter.java
+++ b/app/src/main/java/com/amaze/filemanager/adapters/RecyclerAdapter.java
@@ -21,7 +21,6 @@
 package com.amaze.filemanager.adapters;
 
 import static com.amaze.filemanager.filesystem.compressed.CompressedHelper.*;
-import static com.amaze.filemanager.filesystem.files.FileListSorter.SORT_NONE_ON_TOP;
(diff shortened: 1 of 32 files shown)


seelchen:
This is ready for review!

VishalNehra:
Yea, descending relevance means least relevant at top and vice versa. I didn't notice this wasn't the original intention.

seelchen:
Alright, then I will invert it :+1:  (although I don't quite understand the logic)

VishalNehra:
It's for the UX, just like you said, it'll be confusing for the user. Not all people using the app are technical, they won't see the relevance as a numerical from less to more. Let's try this way, if we get complaints from user that this behaviour is unexpected, we'll revert to your original implementation.

VishalNehra (approved):

PR #4042, opened 2024-01-10
Move deep search results to `SearchView`

### LINKED ISSUE #3357: Can search results be sorted by relevance

### LINKED ISSUE #4041: Show deep search results in `SearchView`
**Is your feature request related to a problem? Please describe.**
Currently, the search results of the deep search are not shown in the `SearchView` like the other searches but in the `MainView`. Integrating it into the `SearchView` will lead to a more consistent design. It also enables sorting of the deep search results and prevents bugs like using the `MainView`'s context menus (which exits the search).

**Describe the solution you'd like**
Move deep search result list to `SearchView`.

**Additional context**
Would also resolve the remaining tasks of #3357.

seelchen:
## Description
This PR changes the deep search to show the results in the `SearchView` and not the `MainView` anymore. The deep search is refactored to run in a `Coroutine` instead of an `AsyncTask`.

#### Issue tracker   

Fixes #4041 
Fixes #3357

#### Automatic tests

- [x] Added test cases
  
#### Manual tests
- [x] Done  

Device:
-  HTC U11 life running Android 10

  

#### Build tasks success  


seelchen:
I am currently removing all things related to the old search from the `MainActivity` and `MainFragment`. During this I saw the `retainSearchTask` boolean which changes the behavior of `MainFragment#goBack` to go back to the search (which doesn't seem to work as far as I can tell).

I think that it is helpful to be able get back to the search in general, but I don't know if changing the behavior of `goBack` is ideal. I think adding an extra button "Go back to search" might be more intuitive. 

Since the current implementation does not work with the new deep search in the `SearchView`, I will delete it. I think that this should be handled in a separate PR.
What are your thoughts about this?

VishalNehra:
@seelchen  can you please check this codacy issue with your last pr #4020

https://github.com/TeamAmaze/AmazeFileManager/pull/4044/checks?check_run_id=20382785390

seelchen:
This is ready for review! I know that it looks like a lot of changes, but most of it is actually just removing an if statement and reducing the indents for the whole block. Since git stores the diff line-based, every line where the indents are reduced are marked as a change...

VishalNehra:
Great work! Is it possible to include regex search in filter search and index search? Currently it only works when user does deep search. It leaves user confused. We either need to mention this in settings that regex will only work in deep search or find a way to include regex search in filter and index search.

VishalNehra (approved):

VishalNehra:
> I implemented this in #4050 so that this PR won't become even bigger. If you want, I can completely disable regex search here and it would be readded when #4050 is merged.
Sorry that was an oversight on my end. Amazing work here! This should be good to go :)

PR #3730, opened 2023-02-11
Optimize Search

VishnuSanal:
## Description

- [x] store recent search items in `SharedPreferences`
- [x] search in the current folder first
- [x] search in the `MediaStore` next
- [x] search in the whole device, if prompted by the user later

#### Automatic tests
- [ ] Added test cases
  
#### Manual tests
- [x] Done  
- Device: Realme Narzo 20A
- OS: Android 11
- Device: Pixel 3A Emulator
- OS: Android 13

#### Build tasks success  
Successfully running following tasks on local:
- [x] `./gradlew assembledebug`
- [x] `./gradlew spotlessCheck`

PR #4050, opened 2024-01-15
Add missing glob support

### LINKED ISSUE #4043: Missing glob support in basic and indexed search and sort by relevance
**Describe the bug**
There is a config option to enable glob patterns (currently called "Regular Expression") for searching. However, using glob patterns is not supported in the basic and indexed search. Sort by relevance also does not take this option into account when calculating the match percentage.

(I also propose to fix the naming of "BASH style regular expression" since it is only glob pattern matching without regex support)

**To Reproduce**
1. Enable "Regular Expression" option
2. Search using a glob pattern (e.g. `*.txt`)
3. See that no files are found using basic and indexed search

**Expected behavior**
That the option is supported for all search modes and for sort by relevance

**Screenshots**
[search-glob-support](https://github.com/TeamAmaze/AmazeFileManager/assets/74261221/80c45bed-75ef-4174-96db-247c9f195275)

 - Device: emulated Google Pixel 7
 - OS: Android 13
 - Rooted: Yes
 - Version: build from 9a11e3e

seelchen:
## Description
This implements the missing glob support for the basic and indexed search. A base class for the search mode is added where the glob support is handled.
To take glob patterns into account for sort by relevance, a data class is used to be able to store more information about the search result.

#### Issue tracker   

Fixes #4043

#### Automatic tests

- [x] Added test cases
  
#### Manual tests
- [x] Done  
  

#### Build tasks success  

#### Related PR  
Related to PR #4042 because this is branched off of there. This can only be merged after PR #4042 is merged.


VishalNehra:
Can you also please look into why search view doesn't show up at all (I think it's behind the shadow overlay) in landscape.

seelchen:
> Can you also please look into why search view doesn't show up at all (I think it's behind the shadow overlay) in landscape.

I looked into it and it seems that the `SearchView` is behind both the FAB overlay and the `AppBar` in the w720dp layout.

I managed to bring it in front of the `AppBar` by setting the `AppBar` to disabled when the `SearchView` is shown, but I can't manage to bring it in front of the overlay. Neither moving it in the "main_toolbar.xml" nor calling `bringToFront()` works.

As far as I can tell, this glitch was introduced in #3730, so I think it should be addressed in a separate PR.

VishalNehra (approved):

PR #4171, opened 2024-05-07
fix last modified timestamps not getting preserved

### LINKED ISSUE #892: Timestamps not kept when moving
I just used Amaze to move all my photos from internal storage to the SD card, and learned the hard way that Amaze does not keep timestamps when moving files.
Since my default gallery app isn't smart enough to use EXIF data to sort images, this means that one year of photos and videos got shuffled.
(Good thing I know Perl; I eventually managed to make a script that sets the timestamp of files based on their name, so at least I fixed the mess I made.)

Apparently when Amaze "moves" files, it just copies them (without preserving timestamps) and then deletes the original, so the solution might be as simple as setting the copy to preserve timestamps, at least when moving files.

VishnuSanal:
#### Issue tracker   
fixes #892

#### Automatic tests
- [ ] Added test cases
  
#### Manual tests
- [x] Done  

- tested on A14 (cut, copy) * (file, directory).

#### Build tasks success  


VishnuSanal:
build test apk

app/src/main/java/com/amaze/filemanager/filesystem/HybridFile.java
```diff
@@ -1440,8 +1442,15 @@ public Boolean execute(@NonNull Session session) throws IOException {
       // do nothing
       return true;
     } else {
-      File f = getFile();
-      return f.setLastModified(date);
+      if (getFile().setLastModified(date)) return true;
+
+      try {
+        Files.setLastModifiedTime(getFile().toPath(), FileTime.fromMillis(date));
```
TranceLove: What if it runs on Android < 8? Perhaps you may need to trap Exceptions like NoSuchMethodException as well.

https://developer.android.com/reference/java/nio/file/Files#setLastModifiedTime(java.nio.file.Path,%20java.nio.file.attribute.FileTime)

(Although I too expect minimal Android version requirement would need to be raised soon as well)

VishnuSanal: trapping `NoSuchMethodException` gives a compiler error. did version checking to avoid this issue.

VishalNehra (approved):

PR #1941, opened 2020-07-25
Upgrade to Robolectric 4

TranceLove:
#### Issue tracker   
Fixes will automatically close the related issue

Fixes #1918
-or-   
Addresses #

#### Release  
Addresses release/3.5 or release/3.6
  
#### Test cases
- [ ] Covered
  
#### Manual testing
- [ ] Done  
  
If yes,  
- Device:
- OS:

#### Build tasks success  
Successfully running following tasks on local 
- [x] `./gradlew assembledebug`
- [x] `./gradlew spotlessCheck`

#### Related PR  
Related to PR #

#### Additional Info
Summary on changes:
- Upgrade Gradle plugin to 3.6.4, last version to work with SDK 28 while fixing classloading issues when running tests
- Upgrade Robolectric to 4.3.1
- Test code refactoring as recommended by doc, enable coverage to API 28. Robolectric requires Java 9 to run tests against API 29, it's still disabled until we found an agreed and convenient way to enable it
- Little build.gradle cleanup, removed duplicate dependency and moved some library version to variables

PR #4245, opened 2024-08-29
fix build fails

VishnuSanal:
fixes the failing builds


build.gradle
```diff
@@ -148,7 +149,7 @@ subprojects {
             }
         }
         dependencies {
-            compileOnly 'com.github.pengrad:jdk9-deps:1.0'
+//            compileOnly 'com.github.pengrad:jdk9-deps:1.0'
```
VishnuSanal, line 151: @TranceLove what were this dependency used for? this causes the build failure. the builds pass even without this; can we safely remove it?

TranceLove: This was introduced when moving towards JDK 9 per #1941. Perhaps this is no longer needed when officially Android apps can build with JDK 11/17?

VishnuSanal: > This was introduced when moving towards JDK 9 per #1941. Perhaps this is no longer needed when officially Android apps can build with JDK 11/17?

Code complies & the tests pass -- shouldn't it be good? 🤔

VishalNehra:
Needed these changes to fix build failures
https://github.com/TeamAmaze/AmazeFileManager/commit/958ec0a414d9c4de37fc5fc3e5271dea37ed209e

VishalNehra (approved):

VishnuSanal:
> Needed these changes to fix build failures
> https://github.com/TeamAmaze/AmazeFileManager/commit/958ec0a414d9c4de37fc5fc3e5271dea37ed209e

What? Didn't get you. 🤔

PR #4283, opened 2024-11-23
Fix NoClassDefFoundError Lorg/bouncycastle/asn1/ASN1ApplicationSpecific

wkornewald:
#### Issue tracker   

https://github.com/TeamAmaze/AmazeFileManager/issues/4165

#### Automatic tests

- [ ] Added test cases
  
#### Manual tests
- [x] Done  
  

#### Build tasks success  


gradle/libs.versions.toml
```diff
@@ -10,7 +10,7 @@ systembartint = "1.0.3"
 robolectric = "4.9"
 glide = "4.14.2"
 sshj = "0.35.0"
-jcifs = "2.1.9"
+jcifs = "2.1.10"
```
VishnuSanal: does this tag of jcifs solve the bouncycasle dependency issue? can you please share the reference?

VishnuSanal (commented):
@wkornewald thanks for your interest in contributing to amaze! but, I have a question, please check the review comments.

gradle/libs.versions.toml
wkornewald: It's mentioned in the changelogs: https://github.com/AgNO3/jcifs-ng/blob/master/CHANGELOG.txt

```
Update dependencies, compatibility with latest BouncyCastle
```

Corresponding commit: https://github.com/AgNO3/jcifs-ng/commit/0eb193e1fb10f68001db9ed07e545347de5ef69d

The issue is that the old jcifs was built against an old BC while the rest of this app is built against the newer BC and the API has changed in an incompatible way, so jcifs fails to find an imported API that only existed in the old BC.

VishnuSanal: awesome, thanks! :)

VishnuSanal (dismissed):

TranceLove:
With #4371 in place this PR is no longer necessary. But thanks again for your contribution :bow:

PR #4316, opened 2024-12-27
Fix issue #4304

### LINKED ISSUE #4304: Incorrect feedback message 'Password doesn't match' appears after entering only the first password field
**Describe the bug**
Incorrect feedback message 'Password doesn't match' appears after entering only the first password field.

**To Reproduce**
1. Click the 3 Vertical Dots icon next to a file
2. Select 'Encrypt'
3. Enter a password in the first password field (e.g., '123')

**Expected behavior**
The feedback message 'Password doesn't match' should appear when the two password fields have different inputs or after text has been entered in the second password field.

**Screen Recording**

https://github.com/user-attachments/assets/6f60ff14-2d0a-46e1-900a-5596cebfcb2b

- Device: [Google Pixel 5]
- OS: [Android 14]
- Rooted: [No]
- Version: [v3.10]

enescakar2148:
## Description
The cause of the error is that the password validaion logic runs for both password fields at the same time. When entering the first password, the second password field is empty, which causes the 'passwords do not match' error. This can be fixed by updating the code as follows:

Fixes #4304 

#### Issue tracker   
Changes made:

The logic for the first password field validation has been updated:
	•	If the first password is not empty, proceed.
	•	If the second password field is empty or if the passwords match, proceed.
	•	If the file name is valid, proceed with “OK”.

Error conditions:
	•	If the first password is empty: “field_empty” error.
	•	If the second password is entered and does not match: “password_no_match” error.
	•	If the file name is invalid: “empty_string” error.

With this change:
	•	No error will be shown when entering the first password.
	•	When entering the second password, the matching check will be performed.
	•	If the passwords do not match, an error will be shown.

This improves the user experience by avoiding unnecessary error messages when entering the first password, and only showing the mismatch error when the second password is entered.

**Contributors:**
- Enes Çakar (e.52nes@hotmail.com)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "java.configuration.updateBuildConfiguration": "interactive"
```
VishnuSanal: please revert this change

VishnuSanal (changes requested):
@enescakar2148 thanks for your interest in contributiong to amaze! one formatting change, and your're done, this PR LGTM 🚀

.vscode/settings.json
enescakar2148: Thank you. BuildConfiguration: I changed it automatically

VishnuSanal (changes requested):
Pls revert the changes/additions made to the JSON file.

VishnuSanal (dismissed):
LGTM 🚀  keep going!

EmmanuelMess (dismissed):

EmmanuelMess:
Please check CI.

EmmanuelMess:
Any updates @enescakar2148 ? This is already approved.

enescakar2148:
Unfortunately, I couldn't add an update because I was working too hard. I hope I can make new additions as soon as possible🙏

EmmanuelMess:
Okay, the test that is not passing is EncryptAuthenticateDialogTest::testPasswordValidations.

VishnuSanal:
any updates here @enescakar2148, thanks!

EmmanuelMess:
I had always assumed tests to be somewhat selftesting, it seems this one was wrong for quite a while though.

EmmanuelMess (approved):

PR #4344, opened 2025-01-13
fix npes

VishnuSanal:


EmmanuelMess:
A good idea would be to work to have `@Nullable` and  `@NonNull` a much as possible, which would instantly alert to possible issues.

EmmanuelMess (approved):

VishalNehra (approved):

PR #4370, opened 2025-03-22
Update Room library to 2.6.1

TranceLove:
## Description

Seems Room has even more stricter requirements on database column nullability, hence another migration is required.

#### Automatic tests
- [x] Added test cases
  
#### Manual tests
- [x] Done  
  
#### Build tasks success  


EmmanuelMess (approved):
Tested on emulator. Grid state keeping doesn't work between restarts, but the error is also present on release/4.0.

PR #4389, opened 2025-04-15
Query MediaStore for media file URI for setting ringtone

### LINKED ISSUE #4367: Using to select ringtone won't work and is irreversible
**Describe the bug**
Using Amaze to change the phone call ringtone will result in the ringtone not being applied and no possibility of setting a ringtone that works using the default "Sounds" app, rendering the phone ringtone permanently silent unless external tools are used (see additional info).

**To Reproduce**
1. Install Amaze
2. Go to settings -> Sound and Vibration
3. Select Phone ringtone. Android will ask what app to use to resolve the action. Select Amaze and "Just once"
4. Call the phone. The ringtone won't ring
5. Repeat points 2-3 but select the "Sounds" app. The app won't open, making it impossible to restore any ringtone

**Smartphone:**
 - Device: Fairphone 4
 - OS: Android 13
 - Rooted: No
 - Version: 3.10

**Additional context**
After looking online I found someone with a very similar (possibly the same) problem on the fairphone forum, [here](https://forum.fairphone.com/t/sounds-keeps-stopping-message-cant-change-ringtone-or-other-notification-sounds/81657).
They mention they also used Amaze.

A workaround is provided but I couldn't try it yet as it involves the use of a pc and `adb` to manually set a valid ringtone.

TranceLove:
## Description

With reference to #4367, by inspecting the ringtone settings of a device

```sh
settings get system notification_sound
```

It becomes known that ringtone requires a content URI to the selected audio file on the device.

This PR tries to create that URI by querying the MediaStore. Therefore there is a prerequisite that even the audio file is visible in Amaze, it must also be visible to the MediaStore too, that refreshing the MediaStore is required.

Had attempted to do this for end user in code, but only worked on Android 14 or above, so gave up and not reinvent the wheel instead.

#### Issue tracker   
Fixes #4367

#### Automatic tests

- [x] Added test cases
  
#### Manual tests
- [x] Done  

Device:
- Pixel 4 emulator running Android 9
- Pixel 7 emulator running Android 14

1. Make sure the audio file exists in MediaStore. Some apps like Musicolet can help refresh the MediaStore
2. Open settings -> Sound
3. Change ringtone
4. Choose Amaze when app selection pops up
5. Choose the audio file
6. Ringtone name should shown equal to filename without extension, and no Toast pops up
7. Call the device
8. Device should ring without crashing any app

#### Build tasks success  

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/filesystem/MediaStoreHack.java
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/MediaStoreHack.java
@@ -196,7 +196,7 @@ public class MediaStoreHack {
       } else {
         return null;
       }
-    } catch (Exception e) {
+    } catch (UnsupportedOperationException e) {
       Log.e(TAG, "Error querying MediaStore", e);
       return null;
     }
--- a/app/src/main/java/com/amaze/filemanager/filesystem/MediaStoreHack.java
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/MediaStoreHack.java
@@ -40,6 +40,7 @@ import android.provider.BaseColumns;
 import android.provider.MediaStore;
 import android.util.Log;
 
+import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
 /**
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/amaze/filemanager/filesystem/MediaStoreHack.java
```diff
@@ -164,6 +165,43 @@ public static OutputStream getOutputStream(Context context, String str) {
...
+      if (cursor != null && cursor.moveToFirst()) {
+        int id = cursor.getInt(cursor.getColumnIndexOrThrow(MediaStore.Audio.Media._ID));
+        return MediaStore.Audio.Media.EXTERNAL_CONTENT_URI
+            .buildUpon()
+            .appendPath(String.valueOf(id))
+            .build();
+      } else {
+        return null;
+      }
+    } catch (Exception e) {
```
EmmanuelMess: You shouldn't catch a generic `Exception`, just the specific exception expected on this. Here it is `UnsupportedOperationException`.

app/src/main/res/values/strings.xml
```diff
@@ -875,5 +875,6 @@ You only need to do this once, until the next time you select a new location for
     <string name="trash_bin_cleanup_interval_title">Cleanup interval</string>
     <string name="trash_bin_cleanup_interval_summary">Trigger auto-cleanup interval (hours)</string>
     <string name="file_deletion">File Deletion</string>
+    <string name="error_mediastore_query_uri">Error querying MediaStore for content URI of selected media file. Refresh the MediaStore and try again.</string>
```
EmmanuelMess: This text seems too long for a Toast, it gets cut on "Refresh the..." on Pixel 9 emulator.

EmmanuelMess (dismissed):
Tested on Pixel 9 API 34, sets the ringtone correctly.

EmmanuelMess (approved):
Reapproving based on previous tests.

PR #4410, opened 2025-05-25
transifex.yml: add Fastlane descriptions to be translatable

### LINKED ISSUE #4411: Add Fastlate translations to Transifex

yurtpage:
The app store descriptions are also needs to be translated.

Fixes #4411

### DIFF
--- a/transifex.yml
+++ b/transifex.yml
@@ -6,6 +6,16 @@ filters:
     source_file: app/src/main/res/values/strings.xml
     # path expression to translation files, must contain <lang> placeholder
     translation_files_expression: app/src/main/res/values-<lang>/strings.xml
+  - filter_type: file
+    file_format: TXT
+    source_language: en_US
+    source_file: fastlane/metadata/android/en-US/full_description.txt
+    translation_files_expression: fastlane/metadata/android/<lang>/short_description.txt
+  - filter_type: file
+    file_format: TXT
+    source_language: en_US
+    source_file: fastlane/metadata/android/en-US/full_description.txt
+    translation_files_expression: fastlane/metadata/android/<lang>/short_description.txt
 settings:
   language_mapping:
     sv_SE: sv
--- a/transifex.yml
+++ b/transifex.yml
@@ -9,13 +9,13 @@ filters:
   - filter_type: file
     file_format: TXT
     source_language: en_US
-    source_file: fastlane/metadata/android/en-US/full_description.txt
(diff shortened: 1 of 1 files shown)


EmmanuelMess:
This file wasn't actually tracked, so I added it to transifex.

EmmanuelMess:
The file format TXT basically means that the translated strings should all start with "- ".

EmmanuelMess (dismissed):

yurtpage:
I'm not sure if the TXT is better choice but other apps use it
https://github.com/equeim/tremotesf-android/blob/master/.tx/config#L13

transifex.yml
```diff
@@ -6,6 +6,16 @@ filters:
...
     translation_files_expression: app/src/main/res/values-<lang>/strings.xml
+  - filter_type: file
+    file_format: TXT
+    source_language: en_US
+    source_file: fastlane/metadata/android/en-US/full_description.txt
+    translation_files_expression: fastlane/metadata/android/<lang>/short_description.txt
+  - filter_type: file
+    file_format: TXT
+    source_language: en_US
+    source_file: fastlane/metadata/android/en-US/full_description.txt
```
VishnuSanal, line 17: likewise, shouldn't this be  `fastlane/metadata/android/en-US/short_description.txt`?

transifex.yml
```diff
@@ -6,6 +6,16 @@ filters:
     source_file: app/src/main/res/values/strings.xml
     # path expression to translation files, must contain <lang> placeholder
     translation_files_expression: app/src/main/res/values-<lang>/strings.xml
+  - filter_type: file
+    file_format: TXT
+    source_language: en_US
+    source_file: fastlane/metadata/android/en-US/full_description.txt
+    translation_files_expression: fastlane/metadata/android/<lang>/short_description.txt
```
VishnuSanal, line 13: shouldn't this be `fastlane/metadata/android/<lang>/full_description.txt`?

VishnuSanal (changes requested):
@yurtpage thanks for the PR, but I think the paths were crossed over when copy-pasting. can you please take a look - or CMIIW!

transifex.yml
yurtpage: thank you, good catch, fixed.

yurtpage:
@VishnuSanal thank you, fixed

VishnuSanal (approved):

EmmanuelMess (approved):

PR #4417, opened 2025-05-27
transifex.yml: specify lang code mapping for Fastlane

yurtpage:
The #4410 added Fastlane descriptions to be translatable. The problem is that there is a lang codes mapping for Android `it_IT: it-rIT` (BTW the `zh-Hans: zh-rHN` is invalid).
The same mapping as applied now to the Fastlane which is not needed and this causes a problem in the #4414 where the `hu` was replaced with `hu-rHU`.

I can't find in documentation how to change the mapping per-file
https://help.transifex.com/en/articles/6265125-github-installation-and-configuration

But here in the `.tx/transifex.yaml` the mapping is set via `lang_map` field
https://github.com/equeim/tremotesf-android/blob/master/.tx/config#L14C1-L15C1

So here is a PR that tries this approach.


EmmanuelMess (approved):

EmmanuelMess:
This seems correct and has already been merged for other projects, so I am assuming it would work. Other than that, do you have Fastlane docs saying which codes to use?

VishnuSanal:
> This seems correct and has already been merged for other projects

can you please link them here for future reference?

VishnuSanal (approved):

EmmanuelMess:
@yurtpage 
imagen

PR #4482, opened 2025-11-17
Fix the whitespace around filename or extension causes false negative

### LINKED ISSUE #4478: [isCompressedFile] Whitespace around filename or extension causes false negative
**Affected Method (full signature)**
com.amaze.filemanager.filesystem.files.FileUtils.isCompressedFile(String path)

**Environment**
	•	App version: AmazeFileManager release/4.0 (unmodified source)
	•	Platform: Java 17 / JVM unit tests

⸻

**Steps to Reproduce:**

FileUtils.isCompressedFile(" /tmp/archive.rar ");   // → false (BUG)
FileUtils.isCompressedFile("/tmp/a .zip");          // → false (BUG)

	1.	Call the method with leading, trailing, or inline spaces around the filename or extension.
	2.	The internal extension extraction runs before trimming, so the comparison string includes spaces.

⸻

**Expected Behavior**

Whitespace around filenames or extensions should be ignored.
All of the above inputs should return true because the base extension is .rar or .zip.

**Actual Behavior**

Returns false whenever whitespace is present before or after the dot segment.

⸻

**Impact (Why it matters)**

Users with filenames that contain stray spaces (common when renaming or copying) see valid compressed files mis-categorized.
This affects file-type filters, previews, and UI sorting in AmazeFileManager.

⸻

**Suggested Fix (One paragraph, no code changes)**

Before extracting and validating the extension, normalize input by trimming and collapsing internal whitespace:
path = path.trim(); and ensure the substring following the last . is trimmed before comparison.
This guarantees .rar, .zip, .7z, etc. are recognized regardless of spacing or casing.

siddhesh-06:
## Description

1. Fix incorrect detection of compressed files when filenames contain whitespace
2. Add a MIME-type fallback for files with missing or malformed extensions

#### Issue tracker   

Fixes #4478

#### Automatic tests

- [ ] Added test cases
  
#### Manual tests
- [ ] Done  

#### Build tasks success  

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/filesystem/files/FileUtils.java
+++ b/app/src/main/java/com/amaze/filemanager/filesystem/files/FileUtils.java
@@ -968,8 +968,20 @@ public class FileUtils {
 
   /** Convenience method to return if a path points to a compressed file. */
   public static boolean isCompressedFile(String path) {
-    @Nullable String extension = MimeTypes.getExtension(path);
-    return ArraysKt.indexOf(COMPRESSED_FILE_EXTENSIONS, extension) > -1;
+    if (path == null) return false;
+
+    // Normalize input: trim leading/trailing whitespace first
+    String normalizedPath = path.trim();
+    if (normalizedPath.isEmpty()) return false;
+
+    // Extract extension using MimeTypes helper, then normalize it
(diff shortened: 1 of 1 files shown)


EmmanuelMess (approved):
Tested, fixes the bug.

PR #4486, opened 2025-11-30
Add Save as option for VIEW intents with content:// URIs

SolsticeSpectrum:
## Description

Adds "Save as" option when opening files via VIEW intent with content:// URIs.

Previously, when apps like Mattermost send a VIEW intent for a single file, users only saw "Amaze Document Viewer" which just prompts to download File Utilities. Now users will also see "Save as" option that triggers the existing save flow.

#### Automatic tests
- [ ] Added test cases

#### Manual tests
- [x] Done

- Device: Nothing Phone 2
- OS: Android 16

#### Build tasks success

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/ui/activities/MainActivity.java
+++ b/app/src/main/java/com/amaze/filemanager/ui/activities/MainActivity.java
@@ -639,6 +639,15 @@ public class MainActivity extends PermissionsActivity
          * http://teamamaze.xyz/open_file?path=path-to-file
          */
         path = Utils.sanitizeInput(uri.getQueryParameter("path"));
+      } else if (uri != null && ContentResolver.SCHEME_CONTENT.equals(uri.getScheme())) {
+        // save a single file to filesystem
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/amaze/filemanager/ui/activities/MainActivity.java
```diff
@@ -639,6 +639,15 @@ private void checkForExternalIntent(Intent intent) {
          * http://teamamaze.xyz/open_file?path=path-to-file
          */
         path = Utils.sanitizeInput(uri.getQueryParameter("path"));
+      } else if (uri != null && ContentResolver.SCHEME_CONTENT.equals(uri.getScheme())) {
```
VishnuSanal, line 642: @TranceLove shouldn't we probably sanitize the links here? WDYT?

VishnuSanal (approved):
@SolsticeSpectrum thanks for contributing to Amaze, this works well in testing from my testing. I have one comment to address, andwe are GTG, thanks! please create an issue for this feature, btw.-

PR #4509, opened 2025-12-18
Automatically check fastlane description for length

### LINKED ISSUE #4505: Fastlane changelogs much too long
I see you've added per-release changelogs, wonderful! Just one problem there: fastlane limits their size to 500 chars. Yours are more than twice as long. So before we enable them at our end, can you please ensure they stay within that limit? Thanks in advance!

EmmanuelMess:
## Description

Made this to autocheck the full_description for length, as it seems translations would exceed this in some languages even if the original is smaller.

#### Issue tracker   
Fixes #4505

#### Automatic tests

- [x] Added test cases
  
#### Manual tests
- [x] Done  
  

#### Build tasks success  


VishnuSanal (approved):
LGTM + tests pass, so should be GTG, thanks.

PR #4534, opened 2025-12-29
Fix duplicated translations

yurtpage:
## Description

For translations the Android now uses `values-lang-rREGION` scheme. In the sources there are two duplicated folders `values-eu` and `values-eu-rES`. I compared them and copied missing strings from the old to new and removed the duplicate.

Same for other few locales


yurtpage:
I also checked other translations and there was a lot of empty files that doesn't have any translations. I removed them. Also fixed minor issued that found.

EmmanuelMess:
Transifex will overwrite these changes, changes to translations must be done in Transifex.

EmmanuelMess:
If you want you can open an issue explaining which should be kept and which should be removed, and maybe we can remove them from Transifex directly.

Issue #3396, opened 2022-07-11
Crash: do not give permission to AmazeFileManager, then click recent files, app crashes


XYIheng:
**Describe the bug**
do not give permission to AmazeFileManager, then click recent files, app crashes on version 3.7.2

**To Reproduce**
1. install app and open it
2. do not give permission to app
3. click menu
4. click recent files
5. see crash

**Expected behavior**
no crash

**Screenshots**

https://user-images.githubusercontent.com/31406320/178181100-78c4a1fa-975b-4376-9ea4-2d038e353e04.mp4

**Smartphone (please complete the following information):**
 - Device: [e.g. Google Pixel 2]
 - OS: [Android 8]
 - Rooted: [e.g. No]
 - Version: [ 3.7.2]

**Additional context**

PR #4569, opened 2026-02-19
Fix crash when opening Recent Files without storage permission

geoffreysisco:
## Description
Fix crash when opening **Recent Files** while storage permission is denied on `release/4.0`.

### Root cause
`LoadFilesListTask.listRecentFiles()` performs a `MediaStore` query that can throw `SecurityException` when `READ_EXTERNAL_STORAGE` is denied. The exception escapes `AsyncTask.doInBackground()` and crashes the app.

### Fix
Catch `SecurityException` around the `MediaStore` query and return the current `recentFiles` list (empty) when permission is denied.

### Behavior
- **Permission denied** → no crash; Recent Files shows empty list
- **Permission granted** → unchanged behavior

### Issue tracker
Addresses #3396

### Verification (manual)
- Device: Android Emulator (Pixel 2)
- OS: Android 8.1 (x86)
- Install: fresh
- Package tested: `com.amaze.filemanager.debug`
- Version: 3.11.2 (`versionCode 124`), build `v3.11.2-76-gd99897689`

Steps:
1. Launch app
2. Deny storage permission
3. Open **Recent Files**

Results:
- Before: crash with `SecurityException`
- After: no crash with permission denied
- Regression: works normally with permission granted

### Build checks

Artifacts (logcat + screen recordings) captured locally; not included in the repo.

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/LoadFilesListTask.java
+++ b/app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/LoadFilesListTask.java
@@ -551,29 +551,34 @@ public class LoadFilesListTask
     c.set(Calendar.DAY_OF_YEAR, c.get(Calendar.DAY_OF_YEAR) - 2);
     Date d = c.getTime();
     Cursor cursor;
-    if (SDK_INT >= Q) {
-      Bundle queryArgs = new Bundle();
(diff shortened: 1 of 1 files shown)


EmmanuelMess:
Approved, but the actual issue here is the user being allowed to use the app without permissions given IMO.

EmmanuelMess (approved):

geoffreysisco:
Thanks for the review.

PR #4578, opened 2026-03-06
Fix browsing when storage permission is denied

### LINKED ISSUE #4570: User is allowed to use Amaze without granting permissions
**Describe the bug**
User is allowed to use Amaze without granting permissions

**To Reproduce**
1. Open device with API 34
2. Select grant
3. Go back immediately
4. Use app

**Expected behavior**
The grant permission dialog on app resuming.

geoffreysisco:
## Description

User is allowed to use Amaze without granting permissions.

Check storage permission on resume so backing out of the
grant flow does not allow filesystem browsing without permission.

Track the All Files Access dialog to avoid stale permission
dialog state when resuming.

#### Issue tracker
Fixes #4570

#### Automatic tests
- [ ] Added test cases

#### Manual tests
- [x] Done

Reproduced the original bug on upstream build and verified
correct behavior with patched build.

- Device: Pixel 9 Pro emulator
- OS: Android 16 / API 36

* Device: Samsung SM-S711U
* OS: Android 16 / API 36

Before/after reproduction videos:

Before (upstream behavior):
<video src="https://github.com/user-attachments/assets/e9f54c9e-84f4-4fde-a69a-102171cf7a41" height="500px"/>

After (patched behavior):
<video src="https://github.com/user-attachments/assets/e78bedfa-ed61-4659-8124-b3d75a3d1308" height="500px"/>

#### Build tasks success

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/ui/activities/MainActivity.java
+++ b/app/src/main/java/com/amaze/filemanager/ui/activities/MainActivity.java
@@ -413,7 +413,15 @@ public class MainActivity extends PermissionsActivity
       getPrefs().edit().putBoolean(PREFERENCE_BOOKMARKS_ADDED, true).commit();
     }
 
-    checkForExternalPermission();
+    String actionIntent = intent != null ? intent.getAction() : null;
+
+    if (Intent.ACTION_GET_CONTENT.equals(actionIntent)
+        || RingtoneManager.ACTION_RINGTONE_PICKER.equals(actionIntent)
+        || Intent.ACTION_VIEW.equals(actionIntent)
+        || Intent.ACTION_SEND.equals(actionIntent)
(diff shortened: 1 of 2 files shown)


EmmanuelMess (dismissed):
Thanks!

EmmanuelMess:
Looks like the rar tests failed, maybe it has a permissions issue?

geoffreysisco:
Thanks for flagging that. It does appear to be related to the permission flow — the failing rar tests seem to be triggered by a startup permission check path in MainActivity.

I’m currently tracking it down and narrowing the conditions so the permission prompt only runs for the intended external entry points. I’ll push an update to the PR once I’ve confirmed the fix locally.

EmmanuelMess (approved):

PR #4579, opened 2026-03-06
Fix Russian translation

marioaltecci:
Fix: Russian translation for battery saver settings.


EmmanuelMess:
Problems with the translations have to fixed in Transifex: https://app.transifex.com/amaze/amaze-file-manager/dashboard/

PR #4584, opened 2026-03-13
Fix file rename in drawer views

### LINKED ISSUE #4560: Rename not work in "Audios", "APKs", etc.
**Describe the bug**
Navigate to "Audios", "APKs", etc. rename a file and rename failed.

**To Reproduce**
1. Navigate to APKs
2. rename a file

**Expected behavior**
The file should be renamed.

**Screenshots**
https://github.com/user-attachments/assets/3ae26fe7-4e34-4ba1-815d-49d9a5af7c0c

 - Device: [Google pixel 9]
 - OS: [Android 16]
 - Rooted: No
 - Version: 3.11.2

**Additional context**
A similar problem to #4558

geoffreysisco:
Use file mode and parent directory from the selected item when invoking rename instead of fragment navigation state.

## Description

Rename operations launched from drawer category views could fail with
**Operation Unsuccessful** even when the same file could be renamed
successfully from a normal filesystem directory.

`MainFragment.rename()` was invoking rename using the fragment's current
navigation state:

* `mainFragmentViewModel.getOpenMode()`
* `mainFragmentViewModel.getCurrentPath()`

In drawer category views, those values represent the current category
view rather than the selected file's actual filesystem location.

This change uses the selected item's filesystem context instead:

* file mode via `f.getMode()`
* parent directory via `f.getParent(requireContext())`

Rename operations now resolve against the selected file's actual
location rather than the fragment's current navigation state.

#### Issue tracker
Fixes #4560

#### Automatic tests
- [ ] Added test cases

#### Manual tests
- [x] Done

- Device: Pixel 9 Pro emulator
- OS: Android 16
- Rooted: No
- Version: 3.11.2

#### Build tasks success

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
+++ b/app/src/main/java/com/amaze/filemanager/ui/fragments/MainFragment.java
@@ -1042,13 +1042,14 @@ public class MainFragment extends Fragment
               AppCompatEditText textfield =
                   dialog.getCustomView().findViewById(R.id.singleedittext_input);
               String name1 = textfield.getText().toString().trim();
+              String targetParent = f.getParent(requireContext());
 
               getMainActivity()
                   .mainActivityHelper
                   .rename(
-                      mainFragmentViewModel.getOpenMode(),
+                      f.getMode(),
                       f.getPath(),
-                      mainFragmentViewModel.getCurrentPath(),
+                      targetParent,
                       name1,
                       f.isDirectory(),
                       getActivity(),


VishnuSanal (approved):
LGTM 🚀 thanks!

EmmanuelMess (changes requested):
Don't use File to get the parent, use the hybrid file to prevent issues with remote modes and DocumentFile. I am not quite sure if you don't need to check if its remote or DocumentFile in this case.

geoffreysisco:
Updated — I removed the raw `File` parent lookup and now use the selected `HybridFileParcelable`’s inherited `HybridFile.getParent(Context)` handling instead.

That method already branches by `OpenMode`, including SMB/SFTP and `DOCUMENT_FILE`, so the rename call site does not need to add separate remote or DocumentFile checks here. The rename flow still passes the selected item’s mode/path, preserving the drawer-view fix while keeping parent resolution inside Amaze’s filesystem abstraction.

Verified with:

* `./gradlew spotlessApply`
* `./gradlew spotlessCheck`
* manual drawer/category rename test on Pixel 9 Pro emulator, Android 16
* rotation tested; no crash observed

VishnuSanal (approved):

EmmanuelMess (approved):

geoffreysisco:
Thanks for taking the time to review this and get it merged.

PR #4589, opened 2026-03-20
Fix drawer category views not refreshing after file operations

### LINKED ISSUE #4558: Refresh not work when deleting files in "Images", "Videos", etc.
**Describe the bug**
v3.11.2 Refresh not work when deleting files in "Images", "Videos", etc.

**To Reproduce**
1. Go to the drawer
2. click "Images", "Videos", "Audio" (any one of it)
3. delete a file

**Expected behavior**
The file should be deleted and disappear on screen

 - Device: Google Pixel 9
 - OS: Android 16
 - Rooted: No
 - Version: 3.11.2

**Additional context**

geoffreysisco:
## Description

Drawer category views did not consistently reflect filesystem changes
after file operations such as create and delete.

In some categories (Documents, APKs, Recent Files), queries could include
entries from Amaze's trash directory (.AmazeData), causing deleted files
to remain visible and allowing repeated deletes to create duplicate
Trash Bin entries.

This PR addresses the issue by:
- forcing reload for drawer category views to ensure fresh data after operations
- filtering Amaze trash paths from category queries

This ensures drawer views correctly reflect the current filesystem state
across all categories.

#### Issue tracker
Fixes #4558

#### Automatic tests
- [ ] Added test cases

#### Manual tests
- [x] Done

- Device: Pixel 9 Pro (emulator)
- OS: Android 16.0
- Rooted: No
- Version: 3.11.2

Verified in:
- Documents
- APKs
- Recent Files
- Images
- Videos
- Audio

Scenarios tested:
- Delete file → removed from drawer view
- Repeat delete → not possible (item is no longer visible after deletion)

#### Before

https://github.com/user-attachments/assets/4d083fba-5e38-4df2-9763-057fb53fecad

#### After

https://github.com/user-attachments/assets/9343bcff-60e7-4ba5-8a25-73a66d879b06

#### Build tasks success
Successfully running the following tasks locally:

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/LoadFilesListTask.java
+++ b/app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/LoadFilesListTask.java
@@ -376,22 +376,37 @@ public class LoadFilesListTask
 
   private List<LayoutElementParcelable> listImages() {
     final String[] projection = {MediaStore.Images.Media.DATA};
-    return listMediaCommon(MediaStore.Images.Media.EXTERNAL_CONTENT_URI, projection, null);
+    final String selection = MediaStore.Files.FileColumns.DATA + " NOT LIKE ?";
+    final String[] selectionArgs = new String[] {AppConfig.getTrashBinBasePath() + "%"};
+    return listMediaCommon(
+        MediaStore.Images.Media.EXTERNAL_CONTENT_URI, projection, selection, selectionArgs);
   }
 
   private List<LayoutElementParcelable> listVideos() {
     final String[] projection = {MediaStore.Video.Media.DATA};
-    return listMediaCommon(MediaStore.Video.Media.EXTERNAL_CONTENT_URI, projection, null);
+    final String selection = MediaStore.Files.FileColumns.DATA + " NOT LIKE ?";
+    final String[] selectionArgs = new String[] {AppConfig.getTrashBinBasePath() + "%"};
+    return listMediaCommon(
+        MediaStore.Video.Media.EXTERNAL_CONTENT_URI, projection, selection, selectionArgs);
   }
 
   private List<LayoutElementParcelable> listaudio() {
-    String selection = MediaStore.Audio.Media.IS_MUSIC + " != 0";
+    String selection =
+        MediaStore.Audio.Media.IS_MUSIC
+            + " != 0 AND "
+            + MediaStore.Files.FileColumns.DATA
+            + " NOT LIKE ?";
     String[] projection = {MediaStore.Audio.Media.DATA};
-    return listMediaCommon(MediaStore.Audio.Media.EXTERNAL_CONTENT_URI, projection, selection);
+    String[] selectionArgs = new String[] {AppConfig.getTrashBinBasePath() + "%"};
+    return listMediaCommon(
+        MediaStore.Audio.Media.EXTERNAL_CONTENT_URI, projection, selection, selectionArgs);
   }
 
   private @Nullable List<LayoutElementParcelable> listMediaCommon(
-      Uri contentUri, @NonNull String[] projection, @Nullable String selection) {
+      Uri contentUri,
+      @NonNull String[] projection,
+      @Nullable String selection,
+      @Nullable String[] selectionArgs) {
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/LoadFilesListTask.java
```diff
@@ -92,6 +92,7 @@ public class LoadFilesListTask
     extends AsyncTask<Void, Throwable, Pair<OpenMode, List<LayoutElementParcelable>>> {
 
   private static final Logger LOG = LoggerFactory.getLogger(LoadFilesListTask.class);
+  private static final String AMAZE_TRASH_PATH = "/storage/emulated/0/.AmazeData";
```
VishnuSanal: can we pls use `AppConfig.TRASH_BIN_BASE_PATH` here?

app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/LoadFilesListTask.java
```diff
@@ -437,7 +438,9 @@ else if (cursor.getCount() > 0 && cursor.moveToFirst()) {
     else if (cursor.getCount() > 0 && cursor.moveToFirst()) {
       do {
         String path = cursor.getString(cursor.getColumnIndex(MediaStore.Files.FileColumns.DATA));
-
+        if (path != null && path.startsWith(AMAZE_TRASH_PATH)) {
+          continue;
+        }
```
VishnuSanal: hmm, nice, but I think we can add this selection to the cursor itself like below. WDYT?

```kotlin
    final String[] projection = {MediaStore.Files.FileColumns.DATA};
    final String selection = MediaStore.Files.FileColumns.DATA + " NOT LIKE ?";
    final String[] selectionArgs = new String[] {AppConfig.TRASH_BIN_BASE_PATH + "%"};
    Cursor cursor =
        context
            .getContentResolver()
            .query(MediaStore.Files.getContentUri("external"), projection, selection, selectionArgs, null);
```

VishnuSanal (changes requested):
hi @geoffreysisco great work, thanks for contributing to amaze! :) I have some comments, please address them, and we are good to go, tahnsk!

- also, can you pls add the trash directory check to all quick access menu listings too? thanks!

app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/LoadFilesListTask.java
geoffreysisco: `TRASH_BIN_BASE_PATH` is currently private in AppConfig, so it’s not accessible here.

I can expose it via a getter (e.g. `AppConfig.getTrashBinBasePath()`) and use that instead — let me know if you’d prefer a different approach.

app/src/main/java/com/amaze/filemanager/asynchronous/asynctasks/LoadFilesListTask.java
geoffreysisco: I agree, that makes sense. I’ll move the trash exclusion into the query instead of filtering in the loop.

geoffreysisco:
Thanks for the review.

I’ve started addressing the comments.

Documents, APKs, and Recent files now exclude the trash directory at query time instead of filtering in the loop.

For Images/Videos/Audio, the stale delete behavior is already handled via the `forceReload` change in `MainFragment` for `OpenMode.CUSTOM`, but they still go through `listMediaCommon()` for querying.

When you mentioned “all quick access menu listings,” should the trash-path exclusion also be applied in `listMediaCommon()` for Images/Videos/Audio, or only to the MediaStore.Files-based listings touched here (Documents/APKs/Recent files)?

geoffreysisco:
I went ahead and applied the trash-path exclusion across all quick access menu listings.

For Images/Videos/Audio, I extended `listMediaCommon()`  to support `selectionArgs` and added the same exclusion there, so filtering is now handled consistently at query time instead of in the loop.

Let me know if you'd prefer this scoped differently.

VishnuSanal (approved):
LGTM, thanks! 🚀

Issue #4583, opened 2026-03-11
Can't dismiss overwrite box


Noitarud:
**Describe the bug**
options given are Skip Overwrite Rename. If I have 50 objects in queue I have a predicament.

**To Reproduce**
0. select objects
1. copy
2. paste (same directory sufficient)
3. try to exit dialogue without Skip Overwrite Rename

**Expected behavior**
Tapping the back button or tapping outside the box to get to main window.

**Smartphone (please complete the following information):**
 - Rooted: No
 - Version: 3.1.0

**Additional context**
i encountered issue because I copied instead of cut and seeing still present I did again.

EmmanuelMess:
(Testing the PR #4600 I seem to have realized that something doesn't make sense in this issue).

Why not press the "Do for all files" and then "Skip"? How should a dismiss work?

VishnuSanal:
> Why not press the "Do for all files" and then "Skip"? How should a dismiss work?

@EmmanuelMess agreed. this is the expected behaviour - and I don't think this is a "bug". maybe a UX glitch since it affects usability (I personally feel this is fine since it enforces an action from the user for the selected files than us deciding!). I think let's close this as not planned and reopen if needed.

@Noitarud pls reopen if needed, thanks!

PR #4600, opened 2026-04-06
Fix dismiss overwrite box

abdulroufsidhu:
#### Automatic tests

- [ ] Added test cases
  
#### Manual tests
- [x] Done  
  

#### Build tasks success  


EmmanuelMess:
@abdulroufsidhu I am sorry but I'm forcing this back to discussion at the issue #4583, because the fixes for the issue seem clunky. For example: dismissing as implemented by this PR only dismisses a single overwrite dialog, of a possible many, and it is not clear to the user what action was performed.

VishnuSanal:
thanks, but we don't think this is an issue anymore - [pls see](https://github.com/TeamAmaze/AmazeFileManager/issues/4583#issuecomment-4826239926), and apologies for the trouble, closing for now.

PR #4605, opened 2026-04-16
Chore: Fix Traditional Chinese translations

TranceLove:
## Description

Fix Traditional Chinese translations.

This is just for the characters; detailed wordings, meanings and complete translation still needs translators help :)

#### Build tasks success  


EmmanuelMess:
This should be done via Transifex otherwise it will be overwritten.

PR #4611, opened 2026-04-22
Keep SMB rename verification off the UI thread

### LINKED ISSUE #4610: Potential ANR / `NetworkOnMainThread` Risk: Synchronous SMB network call in `HybridFile.exists()`
**Environment**
* App: Amaze File Manager (`com.amaze.filemanager`)
* Branch: `master`
* File: [`HybridFile.java`](https://github.com/TeamAmaze/AmazeFileManager/blob/master/app/src/main/java/com/amaze/filemanager/filesystem/HybridFile.java)

### The Issue

While reviewing the `HybridFile` implementation, I noticed a potential main-thread blocking risk when dealing with SMB connections. 

Currently, `HybridFile.exists()` directly creates an `SmbFile` and calls its `exists()` method synchronously:

```java
public boolean exists() {
    boolean exists = false;
    // ...
    } else if (isSmb()) {
        try {
            SmbFile smbFile = getSmbFile(2000);
            exists = smbFile != null && smbFile.exists(); // <-- Synchronous network I/O
        } catch (SmbException e) {
            // ...
        }
    }
    // ...
    return exists;
}
```

Because `jcifs.smb.SmbFile.exists()` is a network-backed operation, it will block the calling thread while connecting, negotiating, or waiting on an unresponsive server.

### The Risk & Impact

If `HybridFile.exists()` is ever reached via a UI-thread call path (e.g., inside an `onClick`, a menu callback, or a `RecyclerView` adapter binding), it will freeze the UI. Depending on the device and server state, this can lead to severe UI jank, ANRs, or a `NetworkOnMainThreadException`.

We know this module family is susceptible to this pattern, as a nearly identical issue caused crashes in the past (see **Issue #3982**: `NetworkOnMainThreadException` on `HybridFile.getUsableSpace`). 

### Next Steps / Verification Required

I have not tracked down a confirmed production path that calls this specifically on the main thread, but because this is a core utility method, it is highly vulnerable to being misused in UI callbacks.

venkyqz:
Problem:
The rename completion path in `MainActivityHelper` may call `newFile.exists(context)` on the UI thread after a rename attempt. For SMB targets that falls into `HybridFile.exists()` and performs synchronous `SmbFile.exists()` network I/O on the main thread.

Root cause:
The post-rename existence probe was added as a SAF workaround because `DocumentFile.renameTo()` can report false even when the rename succeeds, but the fallback was applied uniformly to all backends.

Fix:
Limit the post-rename `exists(context)` verification to SAF-backed targets (`DOCUMENT_FILE` and `OTG`) and skip that UI-thread probe for SMB and other non-SAF modes.

Closes #4610

### DIFF
--- a/app/src/main/java/com/amaze/filemanager/utils/MainActivityHelper.java
+++ b/app/src/main/java/com/amaze/filemanager/utils/MainActivityHelper.java
@@ -388,7 +388,9 @@ public class MainActivityHelper {
                    * DocumentFile.renameTo() may return false even when rename is successful. Hence we need an extra check
                    * instead of merely looking at the return value
                    */
-                  if (b || newFile.exists(context)) {
+                  if (b
+                      || ((newFile.isDocumentFile() || newFile.isOtgFile())
+                          && newFile.exists(context))) {
                     Intent intent = new Intent(MainActivity.KEY_INTENT_LOAD_LIST);
 
                     intent.putExtra(


app/src/main/java/com/amaze/filemanager/utils/MainActivityHelper.java
```diff
@@ -388,7 +388,9 @@ public void done(final HybridFile hFile, final boolean b) {
                    * DocumentFile.renameTo() may return false even when rename is successful. Hence we need an extra check
                    * instead of merely looking at the return value
                    */
-                  if (b || newFile.exists(context)) {
+                  if (b
```
EmmanuelMess, line 391: This doesn't fix the issue (exists running on main thread) and seems incredibly easy to break if the implementation of `Operations.rename` changes (by changing how otg is handled or using DocumentFile for other file types). To fix the issue, the exists check should probably be moved to `Operations.rename`.

EmmanuelMess (changes requested):

PR #4692, opened 2026-07-30
hotfix/3.11.3 to release/4.0

VishnuSanal:

PR #4702, opened 2026-08-05
Fix Espresso tests failing on #4692

TranceLove:
#### Issue tracker   
Addresses #4692 

#### Automatic tests
Updated Espresso tests, no new tests added
  
#### Manual tests
- [ ] Done
  
- Device:
- OS:

#### Build tasks success  

#### Generative code

- [x] This PR used generative code tools (GenAI, LLMs, etc.)

- Model: Claude Sonnet
- Version: 4.6/5
- Provider: Github Copilot

#### Related PR  
Related to PR #4692

### DIFF
--- a/app/src/androidTest/java/com/amaze/filemanager/ui/fragments/BackupPrefsFragmentTest.kt
+++ b/app/src/androidTest/java/com/amaze/filemanager/ui/fragments/BackupPrefsFragmentTest.kt
@@ -27,13 +27,13 @@ import android.content.SharedPreferences
 import android.net.Uri
 import android.os.Build.VERSION.SDK_INT
 import android.os.Build.VERSION_CODES.TIRAMISU
+import android.os.Environment
 import androidx.lifecycle.Lifecycle
 import androidx.preference.PreferenceManager
 import androidx.test.core.app.ActivityScenario
 import androidx.test.core.app.ApplicationProvider
 import androidx.test.espresso.Espresso.onView
 import androidx.test.espresso.action.ViewActions
-import androidx.test.espresso.matcher.ViewMatchers.withId
 import androidx.test.espresso.matcher.ViewMatchers.withText
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.rule.GrantPermissionRule
@@ -43,6 +43,7 @@ import com.amaze.filemanager.ui.activities.PreferencesActivity
 import com.amaze.filemanager.ui.fragments.preferencefragments.BackupPrefsFragment
 import com.google.gson.GsonBuilder
 import com.google.gson.reflect.TypeToken
+import org.awaitility.Awaitility.await
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertFalse
 import org.junit.Assert.assertTrue
@@ -52,10 +53,11 @@ import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
 import java.io.File
+import java.util.concurrent.TimeUnit
 
 @RunWith(AndroidJUnit4::class)
 class BackupPrefsFragmentTest {
-    var storagePath = "/storage/emulated/0"
+    var storagePath = Environment.getExternalStorageDirectory().absolutePath
     var fileName = "amaze_backup.json"
 
     @Rule
@@ -94,6 +96,19 @@ class BackupPrefsFragmentTest {
     }
 
     /**
+     * Waits (with a timeout) for the given file to exist, since some writes to storage happen
+     * asynchronously on a background thread.
+     */
+    private fun waitForFile(
+        file: File,
+        timeoutSeconds: Long = 5L,
+    ) {
+        await().atMost(timeoutSeconds, TimeUnit.SECONDS).until {
+            file.exists()
+        }
+    }
+
+    /**
      * Test whether the exported file contains the expected preference values
      */
     private fun export(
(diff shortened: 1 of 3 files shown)


TranceLove:
Diagnosis on TabFragmentTest failures (By Copilot)

> Root cause: The tests used ActivityTestRule (deprecated) + Espresso.onView(withId(R.id.pager)). The problem: TabFragment is not displayed synchronously on startup — MainActivity.goToMain() runs asynchronously after a Completable completes — so by the time Espresso tried to find R.id.pager the ViewPager2 wasn't in the window hierarchy yet.
> 
> Changes made to TabFragmentTest.kt:
> 
> - Replaced ActivityTestRule with ActivityScenario — each test now launches a fresh ActivityScenario via withScenario { } which ensures proper lifecycle management and safe recreate() support.
> - Replaced Espresso swipe actions with direct ViewPager2 API calls — onView(withId(R.id.pager)).perform(swipeLeft()) was replaced with viewPager.setCurrentItem(index, false) called from scenario.onActivity { }.
> - Added Awaitility polling helpers — awaitPager(), awaitTabFragment(), and awaitCurrentItem() poll with a 10-second timeout, tolerating the asynchronous TabFragment loading. All scenario.onActivity {} calls inside the awaitility loops are wrapped in runCatching {} to survive the brief activity destroyed → recreated window during recreate().
> - Replaced rotation+sleep with scenario.recreate() — this is the proper way to simulate configuration changes in modern instrumented tests, and it avoids the async onActivity callback racing against the destroy/recreate lifecycle.
> - Guarded StoragePermissionHelper for API ≥ R — the permission helper launches its own MainActivity which left a stale activity instance on API < 30; it is now only called on Android 11+.

TranceLove:
Fixes to BackupPrefsFragmentTest (By Copilot)

> - Removed the incorrect runOnUiThread { onView()... } wrapper — Espresso's onView().perform() must run on the instrumentation/test thread, never inside onActivity {} or runOnUiThread {} callbacks (both run on the main/UI thread). Espresso handles UI-thread synchronization internally.
> - Moved all Espresso view actions and JSON-comparison assertions out of activityScenario.onActivity {} onto the test thread, capturing preferences via a lateinit var inside the block for later use.
> Added activityScenario.moveToState(Lifecycle.State.RESUMED) — Espresso requires the activity to be RESUMED to dispatch view actions.
> - Removed the erroneous onView(withId(R.id.home)) click — that ID is a MainActivity-only menu item unrelated to PreferencesActivity. Since exportPrefs() triggers an ACTION_SEND intent to MainActivity, which shows a snackbar with a "Save" action, only onView(withText(R.string.save)).perform(click()) is needed.
> - Added a waitForFile() polling helper since the actual file write to storagePath happens asynchronously after the click.

Copilot also reported a caveat that the test may still fail on API 21 devices, where legacy JAR verifier may be incompatible with modern APK signing on very old Android runtimes.

app/src/androidTest/java/com/amaze/filemanager/ui/fragments/BackupPrefsFragmentTest.kt
```diff
@@ -93,6 +93,23 @@ class BackupPrefsFragmentTest {
         import(exportFile)
     }
 
+    /**
+     * Waits (with a timeout) for the given file to exist, since some writes to storage happen
+     * asynchronously on a background thread.
+     */
+    private fun waitForFile(
```
EmmanuelMess, line 101: Don't we have watchers that can detect if a file was created?

app/src/androidTest/java/com/amaze/filemanager/ui/fragments/TabFragmentTest.kt
```diff
@@ -48,29 +42,31 @@ class TabFragmentTest {
...
-            val activity = activityRule.activity
-            val tabFragment =
-                activity.supportFragmentManager
-                    .findFragmentById(R.id.content_frame) as TabFragment
-
-            // Detach fragment through FragmentManager
-            activity.supportFragmentManager.beginTransaction().apply {
-                tabFragment.fragments.forEach { detach(it) }
-                commit()
+        withScenario { scenario ->
```
EmmanuelMess, line 79: This removed the request for rotation? Doesn't this break this test?

app/src/androidTest/java/com/amaze/filemanager/ui/fragments/TabFragmentTest.kt
```diff
@@ -48,29 +42,31 @@ class TabFragmentTest {
 
     @Before
     fun grantManageStoragePermission() {
-        StoragePermissionHelper.grantManageStoragePermission()
+        if (SDK_INT >= android.os.Build.VERSION_CODES.R) {
+            StoragePermissionHelper.grantManageStoragePermission()
```
EmmanuelMess: This Android version check should be in grantManageStoragePermission itself.

app/src/androidTest/java/com/amaze/filemanager/ui/fragments/TabFragmentTest.kt
```diff
@@ -80,55 +76,127 @@ class TabFragmentTest {
...
+    }
+
+    private fun setCurrentItem(
+        scenario: ActivityScenario<MainActivity>,
+        index: Int,
+    ) {
+        awaitPager(scenario)
+
+        scenario.onActivity { activity ->
+            activity.findViewById<ViewPager2>(R.id.pager).setCurrentItem(index, false)
```
EmmanuelMess: I am reluctant of doing it without a an action on the device. Tests run on emulator should be to test the reaction of the application to "real" actions, otherwise the test could/should be done headless.

app/src/androidTest/java/com/amaze/filemanager/ui/fragments/BackupPrefsFragmentTest.kt
```diff
@@ -93,6 +93,23 @@ class BackupPrefsFragmentTest {
...
+        file: File,
+        timeoutMillis: Long = 5000L,
+    ) {
+        val startTime = System.currentTimeMillis()
```
EmmanuelMess: this should probably use  `await().atMost(5, TimeUnit.SECONDS).until {` if possible.

EmmanuelMess (changes requested):

EmmanuelMess (approved):

PR #4713, opened 2026-08-15
Feature/search teleport to file

### LINKED ISSUE #4695: ALLOW the search to "teleport" to file
**Is your feature request related to a problem? Please describe.**
I want to search a file, and be teleported to its position, that's it

**Describe the solution you'd like**
Currently the app obly allow u to search the file, yes, but once found it only allows u to open it, and nothing else. I want to move/rename/teleport inside its folder etc etc.

**Describe alternatives you've considered**

**Additional context**

yasminhosam:
## Description
This PR adds a "teleport to file" icon in search results. Tapping the new icon navigates to the file's parent folder and highlights/selects it there, so the user can act on it directly (rename, move, delete, etc.) without opening it first.

Directory results in search are unaffected — tapping a directory result still opens it directly, as before.

This reuses the existing `scrollToFileName` + `getScrollPosition()` mechanism already present in `MainActivity`/`MainFragment`/`MainFragmentViewModel` , rather than duplicating scroll/highlight logic.

#### Issue tracker
Fixes #4695

#### Automatic tests
- [x] Added unit tests for specific individual functions
- [ ] Added headless tests for new app functionality
- [ ] Added emulator tests for new UI elements

#### Manual tests

- [x] Done

- Device: Samsung Galaxy A23
- OS: Android 14

#### Build tasks success

#### Generative code
- [x] This PR used generative code tools (GenAI, LLMs, etc.)

- Model: Claude Sonnet 5
- Version: 5
- Provider: Anthropic


app/src/main/java/com/amaze/filemanager/ui/activities/MainActivity.java
```diff
@@ -1069,6 +1069,44 @@ public void goToMain(String path, boolean hideFab) {
...
+    String parentPath = filePath.substring(0, lastSlash);
+    String fileName = filePath.substring(lastSlash + 1);
+    return new ParentAndFileName(parentPath, fileName);
+  }
+
+  /**
+   * Simple holder for a file's parent path and name. Not an android.util.Pair to keep it usable in
+   * plain JUnit tests (android.* classes are stubbed out in non-Robolectric unit tests).
+   */
+  static class ParentAndFileName {
```
EmmanuelMess: Any reason for not using HybridFile here? Does this work with external searches: SMB, FTP, cloud, etc?

yasminhosam: > Any reason for not using HybridFile here? Does this work with external searches: SMB, FTP, cloud, etc?

Thanks for pointing that out! You're right — I was doing naive string splitting on the raw path string, which doesn't correctly handle SMB/FTP/cloud paths.
I’ve updated teleportToFile to use HybridFile directly, so it now correctly handles local, SMB, FTP, and cloud paths.

I also added headless tests covering local, SMB, and FTP paths. The FTP test is skipped on Windows due to a Robolectric file-lock/cleanup issue, but the implementation itself works correctly.

yasminhosam:
Hi @EmmanuelMess, just following up on this PR since I addressed your feedback and updated the implementation/tests.

Whenever you have time, I’d appreciate another review. Thanks!

EmmanuelMess:
> Whenever you have time, I’d appreciate another review. Thanks!

Yes, I'm pretty preoccupied with research right now. If you want a faster review, please provide a comment on how you considered and validated the user experience (especially the back button after teleport), and how you tested on networked files (SMB, etc.).

Right now my biggest consideration is how to verify that the PR is thoughtfully implemented.

yasminhosam:
Thanks for the detailed feedback — here's a full response covering both points.

**Back button UX:**

I traced the back button handling (`MainActivity.onBackPressed()` → `MainFragment.goBack()`) and confirmed that when browsing folders, it doesn't use the FragmentManager back stack. Instead, it computes the parent of the current path and navigates there.

Therefore, after a teleport, pressing Back takes the user up one directory level from the location they were teleported to, rather than back to the search results or the previous screen. For example, if you're at Root, search for a file inside `FolderA/FolderB`, and teleport to it, pressing Back takes you to `FolderA`. This is consistent with the existing Back behavior when navigating folders normally.

While testing this manually, I also found a bug: `scrollToFileName` was not cleared after being consumed, so the previously teleported file could be incorrectly re-highlighted during later folder navigations. I fixed this by clearing it immediately after use in `MainFragment`, and moved `notifyItemChanged()` inside the `scrollPosition != -1` check to avoid calling it with an invalid index. I have just pushed these changes.

**Networked file testing (SMB/FTP/cloud):**

`teleportToFile` only uses `HybridFile.getParent()` and `getName()`, so it does not perform any network I/O. The relevant behavior here is how the different URI/path formats are parsed.

I added headless Robolectric tests covering local, SMB, and FTP path formats in `MainActivityTeleportTest`. Each test verifies that the parent path and file name are resolved correctly for the corresponding protocol. FTP is skipped on Windows due to an unrelated Robolectric temporary-directory cleanup issue.

I haven't tested against a live networked server yet, since the change itself is isolated to path handling and `HybridFile` already handles these path formats. If you think live-server validation would still be useful for this PR, I'm happy to add it.

EmmanuelMess:
@yasminhosam what have you, the person, done to verify the user interaction?

yasminhosam:
> @yasminhosam what have you, the person, done to verify the user interaction?

I tested the complete flow on my Galaxy A23: I searched for files from different locations, teleported to files nested in different folders, and then continued navigating from the destination.

For the Back behavior specifically, I checked whether Teleport should introduce a new navigation history or behave like normal folder navigation. The existing app treats Back as "go to the parent of the current folder," rather than "go back to the previous screen," so after teleporting I kept that same behavior instead of introducing a separate Back flow for Teleport.

I also tested what happens after continuing to browse normally after a teleport. I teleported to a file, then navigated somewhere else afterward, and the old file was still shown as highlighted even though I wasn't looking at it anymore. That made me treat the teleport state as a one-shot interaction and clear it after it's consumed.

https://github.com/user-attachments/assets/eb68304a-828d-4495-acd9-95b69cadc47a 



https://github.com/user-attachments/assets/a17a6c3b-50f7-446b-a297-e2805b84aac3

yasminhosam:
One more case I ran into while doing this repeatedly that teleporting to a file that's already in the folder I'm currently viewing didn't scroll or highlight anything at all. Turned out the list can reload more than once for a single navigation (once with a partial list while it's still loading, once with the full one), and I was clearing the target filename too early, before the full list had a chance to search it. Fixed that so it only clears once the file is actually found, and re-tested the same-folder case, the nested-folder case

app/src/androidTest/java/com/amaze/filemanager/ui/activities/MainActivityTeleportScrollInstrumentedTest.kt
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /**
+     * Verifies the target file's row becomes visible on screen after teleportToFile is called.
+     */
+    @Test
+    fun testTeleportScrollsTargetFileIntoView() {
+        activityRule.scenario.onActivity { activity ->
+            val file = HybridFile(OpenMode.FILE, targetFile.absolutePath)
+            activity.teleportToFile(file)
```
EmmanuelMess, line 101: In an emulator test, run as much as possible via the UI, in this case, run a search and press the teleport to file button. You might want to implement this as a Roboletric test, but that is your choice..

app/src/androidTest/java/com/amaze/filemanager/ui/activities/MainActivityTeleportScrollInstrumentedTest.kt
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        // Give the async directory load + scroll a moment to complete before asserting.
+        Thread.sleep(2000)
```
EmmanuelMess, line 105: No sleeps, use event based callbacks, unless it is a hack, which should be documented. Also, the emulator should not have animations of any kind when running instrumented tests, per docs.

app/src/test/java/com/amaze/filemanager/ui/activities/MainActivityTeleportTest.kt
```diff
@@ -0,0 +1,210 @@
...
+     * getAllTabs()).
+     *
+     * Since TabHandler is shared/upstream code, we refresh its internal `database`
+     * reference via reflection before every test instead of modifying it, pointing it
+     * at the current test's fresh ExplorerDatabase instance. Test-only workaround.
+     */
+    @Before
+    fun refreshTabHandlerDatabaseReference() {
+        runCatching {
+            val tabHandler = TabHandler.getInstance()
```
EmmanuelMess, line 73: Can't you do this via `@PublicForTesting`?
@TranceLove can you take a look at this?

app/src/main/java/com/amaze/filemanager/ui/activities/MainActivity.java
```diff
@@ -2456,11 +2467,11 @@ private void initLeftRightAndTopDragListeners(boolean destroy, boolean shouldInv
   /**
    * Invoke {@link FtpServerFragment#changeFTPServerPath(String)} to change FTP server share path.
    *
+   * @param dialog
```
EmmanuelMess, line 2470: Don't reorder lines to leave `@see` below other stuff.

app/src/main/java/com/amaze/filemanager/ui/activities/MainActivity.java
```diff
@@ -2552,8 +2567,8 @@ public void setListItemSelected(boolean value) {
   /**
    * Do nothing other than dismissing the folder selection dialog.
    *
-   * @see com.afollestad.materialdialogs.folderselector.FolderChooserDialog.FolderCallback
    * @param dialog
+   * @see com.afollestad.materialdialogs.folderselector.FolderChooserDialog.FolderCallback
```
EmmanuelMess, line 2571: Don't reorder lines to leave @see below other stuff.

app/src/main/java/com/amaze/filemanager/ui/fragments/data/MainFragmentViewModel.kt
```diff
@@ -228,20 +228,14 @@ class MainFragmentViewModel : ViewModel() {
     /**
      * Get the position of an item
      */
-    fun getScrollPosition(title: String): MutableLiveData<Int> {
```
EmmanuelMess, line 231: Why was this changed? Doesn't it have performance implications?

app/src/main/java/com/amaze/filemanager/ui/activities/MainActivity.java
```diff
@@ -2540,6 +2551,10 @@ public String getScrollToFileName() {
     return this.scrollToFileName;
   }
 
+  public void setScrollToFileName(String scrollToFileName) {
```
EmmanuelMess, line 2554: Add `@Nullable` to the type.

EmmanuelMess (changes requested):
This is the code review.
Also, tested this on an emulator.
There is a UX problem: when you teleport to a file, it is a sort of "broken selection", this can be seen when you select another file, then the amount of selected files is 2 and the UI fixes itself to allow the normal bulk operations for selected files.
As I [had mentioned](https://github.com/TeamAmaze/AmazeFileManager/issues/4695#issuecomment-5341385774) "highlighting is too similar to selected item from selection", it has to be differentiated (best solution) or it has to be correctly selected, with the UI correctly reflecting the selection.

yasminhosam:
> This is the code review. Also, tested this on an emulator. There is a UX problem: when you teleport to a file, it is a sort of "broken selection", this can be seen when you select another file, then the amount of selected files is 2 and the UI fixes itself to allow the normal bulk operations for selected files. As I [had mentioned](https://github.com/TeamAmaze/AmazeFileManager/issues/4695#issuecomment-5341385774) "highlighting is too similar to selected item from selection", it has to be differentiated (best solution) or it has to be correctly selected, with the UI correctly reflecting the selection.

You're right. I agree, I should have focused more on clearly differentiating the teleport highlight from the selection state.

I’ve changed it now so the teleport highlight is completely separate from selection and is not counted as a selected item. It uses a short two-blink highlight instead of the selection background, for both list and grid views.

I initially tried using `notifyItemChanged` to restore the item after the highlight, but it caused a visible icon flicker because of the rebind. So I went with restoring the background directly after the animation instead.

Before I push, I wanted to get your take on the approach in the video, and flag one open question: is a ~900ms two-blink flash the right feel, or would you prefer a single fade, a different duration, or a different visual treatment entirely?

I’ll continue working on the remaining review comments separately.


https://github.com/user-attachments/assets/7e321c9a-29c1-4192-9b04-a2d70b99fa28

app/src/main/java/com/amaze/filemanager/ui/fragments/data/MainFragmentViewModel.kt
yasminhosam: i was having a bug in teleporting to a file already in the currently open folder didn't scroll or highlight anything. Root cause is that `reloadListElements` can fire more than once per navigation — once with a partial list while the folder is still loading, once with the full list once it's done. `scrollToFileName` was being cleared right after the first call read it, regardless of whether the file was actually found. So by the time the second, complete pass ran, the target filename was already gone and nothing happened.

i fixed it by moving `setScrollToFileName(null)` to only run after `scrollPosition != -1`

yasminhosam: While fixing that I also noticed `getScrollPosition` itself had a separate issue: it returned a `MutableLiveData<Int>` initialized to -1, then launched a coroutine on `Dispatchers.IO` to search `adapterListItems` and `postValue` the real index once found. That's two values arriving at two different times with no guarantee about when the second one lands relative to anything else reading it is a race condition. And there's no real reason for it
 `adapterListItems` is just an in-memory list, there's no I/O to justify running the search on a background thread at all.
 i fixed it by make  `getScrollPosition` synchronous so it returns one correct value immediately,

EmmanuelMess:
> Before I push, I wanted to get your take on the approach in the video, and flag one open question: is a ~900ms two-blink flash the right feel, or would you prefer a single fade, a different duration, or a different visual treatment entirely?

As [I had mentioned](https://github.com/TeamAmaze/AmazeFileManager/issues/4695#issuecomment-5326667164), this time based approach has an accessibility issue: it is entirely time dependent, if you miss it, there is no indication that it happened.

